PR #7, opened 2020-06-04
[REVIEW] Adding Hash Bucket Op

alecgunny:
Adding hash bucket op and unit test for reducing categorical cardinality.

One possible feature worth considering for the future (but not strictly necessary) is to make num_buckets a kwarg and conditioning an encoder stat on its being left as None. In this case, the encoder would be used to count categories and a rule of thumb would then be used to decide on number of buckets.

This could even be subsumed into the Categorify op as a sliding scale with the frequency count, where everything beneath the frequency count is hashed to a given number of buckets (either provided explicitly by the user or decided via the statistic and the rule of thumb). The current Categorify behavior would reduce to the case that num_buckets=1, and the HashBucket op reduces to the case that the frequency threshold is set to infinity.

### DIFF
--- a/nvtabular/groupby.py
+++ b/nvtabular/groupby.py
@@ -288,7 +288,7 @@ class GroupByMomentsCal(object):
                 for cont_name in self.cont_col:
                     for su_op in self.supported_ops:
                         if su_op in self.stats_names:
-                            if su_op == 'count':
+                            if su_op == "count":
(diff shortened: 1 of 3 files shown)


alecgunny:
run tests

benfred:
rerun tests

benfred:
reruntests

benfred:
rerun tests

PR #77, opened 2020-06-09
[REVIEW] Add CSVDatasetEngine for DaskDataset

### LINKED ISSUE #61: [FEA] Implement CSV engine for `DaskDataset`
<a href="https://github.com/rjzamora"></img></a> **Issue by [rjzamora](https://github.com/rjzamora)**
_Thursday May 28, 2020 at 17:00 GMT_
_Originally opened as https://github.com/rapidsai/recsys/issues/181_

----

#136 introduces a new `DaskDataset` API for managing IO needs within a `DaskWorkflow`.  This API only supports `"parquet"` at the moment.  A similar engine for "csv" needs to be implemented.

We need a `CSVDatasetEngine` implementation in `nvtabular/dask/io.py`.

rjzamora:
Closes #61 

This PR adds basic CSV support for Dask-backed NVTabular workflows.  Note that `CSVDatasetEngine` is just a thin wrapper around `dask_cudf.read_csv`.  This PR also cleans up and improves some of the dask-based IO code. For example, the user may now pass a `storage_options` argument to `DaskDataset` (used by `fsspec` for non-local file systems).

### DIFF
--- a/nvtabular/dask/io.py
+++ b/nvtabular/dask/io.py
@@ -238,11 +238,11 @@ class ParquetDatasetEngine(DatasetEngine):
     """ ParquetDatasetEngine
 
         Dask-based version of cudf.read_parquet.
-        TODO: Improve dask_cudf.read_parquet performance so that
-        this class can be slimmed down.
     """
 
     def __init__(self, *args, row_groups_per_part=None):
+        # TODO: Improve dask_cudf.read_parquet performance so that
+        # this class can be slimmed down.
         super().__init__(*args)
         self._metadata, self._base = self.get_metadata()
         self._pieces = None
--- a/nvtabular/dask/io.py
+++ b/nvtabular/dask/io.py
@@ -169,7 +169,13 @@ class DaskDataset:
     """
 
     def __init__(
-        self, path, engine, part_size=None, part_mem_fraction=None, storage_options=None, **kwargs
+        self,
+        path,
+        engine=None,
+        part_size=None,
+        part_mem_fraction=None,
+        storage_options=None,
+        **kwargs,
     ):
 
         if part_size:
@@ -349,6 +355,7 @@ class CSVDatasetEngine(DatasetEngine):
     def __init__(self, *args, **kwargs):
         super().__init__(*args)
         self._meta = {}
+        self.names = kwargs.pop("names", None)
         self.csv_kwargs = kwargs
         # CSV reader needs a list of files
         # (Assume flat directory structure if this is a dir)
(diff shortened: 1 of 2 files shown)


rjzamora:
@jperez999 - It would be good to get your eyes on this since you have a lot of experience with the single-gpu csv reader

nvtabular/dask/io.py
```diff
@@ -159,17 +162,28 @@ class DaskDataset:
         to GPU memory capacity). Ignored if part_size is passed
         directly. Note that the underlying engine may allow other
         custom kwargs to override this argument.
+    storage_options: None or dict
```
rjzamora, line 165: @benfred - Note that I am explicitly adding `storage_options` here to hopefully help with future plans to support [remote data](https://docs.dask.org/en/latest/remote-data-services.html).

nvtabular/dask/io.py
```diff
@@ -159,17 +162,28 @@ class DaskDataset:
...
+        storage_options=None,
+        **kwargs,
+    ):
 
         if part_size:
             # If a specific partition size is given, use it directly
             part_size = parse_bytes(part_size)
         else:
             # If a fractional partition size is given, calculate part_size
+            part_mem_fraction = part_mem_fraction or 0.125
```
benfred, line 186: Should we just set the default to 0.125 in the method parameters (rather than go part_mem_fraction=None) ?

nvtabular/dask/io.py
```diff
@@ -178,42 +192,56 @@ def __init__(self, path, engine, part_size=None, part_mem_fraction=0.125, **kwar
...
     def to_ddf(self, columns=None):
         raise NotImplementedError(""" Return a dask_cudf.DataFrame """)
 
 
 class ParquetDatasetEngine(DatasetEngine):
+    """ ParquetDatasetEngine
+
+        Dask-based version of cudf.read_parquet.
+        TODO: Improve dask_cudf.read_parquet performance so that
+        this class can be slimmed down.
```
benfred, line 242: We should keep TODO's out of docstrings (which get pulled out for our user facing documentation).

I'd leave this as a comment instead, and link to an issue on dask_cudf about improving performance.

nvtabular/dask/io.py
benfred: nice! thanks for the link

nvtabular/dask/io.py
jperez999: If we set to a default value other than None, we are essentially trying to hide this keyword, right? If so, by setting it to none by default we force the user to understand that this knob exists and inform them they must set it. I feel like that is good behavior training for users because if not, when a user does a groupby operation or a join, if the 0.125 partition is to big they might just think its because the nvtabular doesn't work, and stop using. Or I am just being paranoid here...?

nvtabular/dask/io.py
rjzamora: Good catch - My mistake

nvtabular/dask/io.py
jperez999: So 0.125 is guaranteed to work for all operation sets?

rjzamora: There is certainly no guarantee, but this is a good balance between stability and performance in my experience.

PR #83, opened 2020-06-11
Add GPUDatasetIterator to daskdataset access via to_iter function

jperez999:


nvtabular/dask/io.py
```diff
@@ -249,7 +262,12 @@ def __init__(self, *args, row_groups_per_part=None):
         if row_groups_per_part is None:
             # TODO: Use `total_byte_size` metadata if/when we figure out how to
             #       correct for apparent dict encoding of cat/string columns.
-            path0 = self.fs.sep.join([self._base, self._metadata.row_group(0).column(0).file_path])
+            if os.path.isdir(self._base):
```
rjzamora, line 265: Lets try to avoid using `os` here since we already have an fsspec object to work with (`self.fs.isdir()`)

tests/unit/test_dask_nvt.py
```diff
@@ -163,6 +163,8 @@ def op_logic(self, *args, **kwargs):
     assert math.isclose(result.id.max(), processor.stats["maxs"]["id"], rel_tol=1e-3)
 
 
+# This test fails on known issue https://github.com/rapidsai/cudf/issues/5322
```
rjzamora, line 166: I merged a fix for this bug a week or two ago - Should work fine with nightly?

jperez999: We dont have that in CI, working on getting it in currently.

nvtabular/dask/io.py
```diff
@@ -177,10 +180,11 @@ def __init__(
         storage_options=None,
         **kwargs,
     ):
-
+        self.kwargs = kwargs
         if part_size:
             # If a specific partition size is given, use it directly
-            part_size = parse_bytes(part_size)
+            # part_size = parse_bytes(part_size)
+            part_mem_fraction = part_size
```
rjzamora, line 187: I know this might be a temporary change for debugging, but I want to make sure it is clear that `part_size` is different from `part_mem_fraction`.  It is the number of bytes the user wants for a partition size, and could be in passed in like: `part_size="1GB"` (hence the need for `parse_bytes`)

jperez999: oh!!! Nice, yeah, that parse bytes turns  0.1 0.01 into 0 which is not the desired action... So then to explicitly say I want to use a certain GPU fraction... I have to know how much actual memory that is? (like "24GB") or something like that.

rjzamora: > So then to explicitly say I want to use a certain GPU fraction... I have to know how much actual memory that is? (like "24GB") or something like that.

Yes - the default behavior is that `part_size` is not provided, and we will use a user-specified or default value for `part_mem_fraction` to figure out what `part_size` *should* be.  However, if the user already knows the specific size they want, they should be able to pass it directly with `part_size`.

rjzamora: Feel free to suggest a better description if the docstring is confusing for `part_size`

jperez999: Ahh ok nice, I was using the wrong arg then. I needed to focus my attention on part_mem_fraction. Thanks for clearing that up

jperez999:
rerun tests

rjzamora:
Thanks for attacking this Julio.  I have some thoughts here that I'd like to get your perspective on (you/others may disagree)... 

- We should probaby use this oportunity to move the `DaskDataset` code into `nvtabular/io.py` (and rename it to `Dataset` or `NVTDataset`). I think the two general goals here should be (1) to merge the dask/iterator APIs into one, and (2) to trim redundant code wherever possible.
- We *may* need to add (something like) a `to_iterator` method to `Dataset`, and avoid turning the class itself into an iterator.  I say this because I feel the current purpose of `DaskDataset` is to process a list of file paths and produce a `dask.DataFrame` (for a specific column selection) when needed by the undeerlying nvt workflow.  It seems natural to extend this to also produce an iterator (for a specific column selection) when needed by the undeerlying nvt workflow.  It is true that the column selection is more imortant for dask-based statistics gathering, but I still worry that it will become confusing to iterate on the `Dataset` object itself.

Any thoughts here @benfred ?

benfred:
@rjzamora I agree with both of your points - I think we should move the dask dataset code into nvtabular/io.py, and have an explicit 'iter' call.

I also we should consolidate the 'engine' code: we have a ParquetDatasetEngine and a CSVDatasetEngine in the dask/io code. Likewise we have a '_get_read_engine' in the nvtabular/io code that switches between the parquet/csv iterator implementations. 

Instead, I think we should have an 'iter' method on the ParquetDatasetEngine class which calls ParquetIterator for each file, and likewise a iter method  on CSVDatasetEngine which calls the CSVIterator (rather than have the base class call GPUDatasetIterator). The GPUDatasetIterator class can be deleted after this change.

nvtabular/dask/io.py
```diff
@@ -172,7 +172,7 @@ def __init__(
         storage_options=None,
         **kwargs,
     ):
-
+        self.kwargs = kwargs
```
rjzamora, line 175: is this used?

nvtabular/dask/io.py
```diff
@@ -211,6 +216,10 @@ def __init__(
     def to_ddf(self, columns=None):
         return self.engine.to_ddf(columns=columns)
 
+    def to_iter(self):
```
rjzamora, line 219: Im visualizing the `to_iter` method to actually create and return the iterator within this method (that is, do not greate an iterator attribute in `__init__`, but do it and return it right here.

rjzamora: In the long run, the `Workflow` class will need to know when to convert the `Dataset` object into an iterator (`to_itr`) or a dask_cudf `DataFrame` (`to_ddf`).

nvtabular/dask/io.py
jperez999: yes for passing to the GPU dataset iterator.

nvtabular/dask/io.py
rjzamora: I suggest we accept a `columns` argument here (just like `to_itr`) and just return `self.engine.to_itr(columns=columns)`.   That is, we should probably do the real logic on the engine level.
 The engine can use `names=columns` when creating the iterator (and the engine type will already be known at that level). Overall, I would be hesitant to add any new arguments to the `DaskDataset` class here, most information should already be availbale.

rjzamora: Summary of my previous comment:  Lets define a `to_itr` method on the `DatasetEngine` class and avoid passing itr specific arguments to `DaskDataset` - If anything, `GPUDatasetIterator` should be modified to accept the `part_size` (and/or `row_groups_per_part`) that we are already calculating here.

nvtabular/dask/io.py
```diff
@@ -228,6 +231,9 @@ def __init__(self, paths, part_size, fs, fs_token):
     def to_ddf(self, columns=None):
         raise NotImplementedError(""" Return a dask_cudf.DataFrame """)
 
+    def to_iter(self, columns=None):
```
benfred, line 234: This method is 'to_itr' in all the engine subclasses and 'to_iter' here and in the dataset class on line 214. Can we be consistent on how we name this? I'd prefer to_iter fwiw

nvtabular/dask/io.py
```diff
@@ -197,9 +196,10 @@ def __init__(
         # If engine is not provided, try to infer from end of paths[0]
         if engine is None:
             engine = paths[0].split(".")[-1]
-
         if isinstance(engine, str):
             if engine == "parquet":
+                if "names" in kwargs:
+                    kwargs.pop("names")
```
benfred, line 202: Do we need to pop off names here - is it even used anywhere? Like can we just remove setting the 'names' attribute in the unittest on test_io.py line 63 instead?

nvtabular/dask/io.py
```diff
@@ -370,3 +388,14 @@ def to_ddf(self, columns=None):
...
             self.paths, names=self.names, chunksize=self.part_size, **self.csv_kwargs
         )[columns]
+
+    def to_iter(self, columns=None):
+        part_mem_fraction = self.part_size / cuda.current_context().get_memory_info()[1]
+        itr = GPUDatasetIterator(
+            self.paths,
+            engine="csv",
+            gpu_memory_frac=part_mem_fraction,
+            names=self.names,
```
rjzamora, line 398: you don't need all the code to add `names` as a member variable.  This line could just be: `names=self.csv_kwargs.get("names", None)`

nvtabular/dask/io.py
```diff
@@ -359,14 +374,23 @@ class CSVDatasetEngine(DatasetEngine):
     def __init__(self, *args, **kwargs):
         super().__init__(*args)
         self._meta = {}
-        self.names = kwargs.pop("names", None)
         self.csv_kwargs = kwargs
+        self.names = self.csv_kwargs.get("names", None)
```
rjzamora, line 378: I don't feel too strongly about it, but there is really no reason to create a `names` attribute.  We are already storing `csv_kwargs`, which is expected to include `names`.  Note that the existing `test_dask_workflow_api_dlrm` test is already passing `names` to the dask_cudf.read_csv backend this way

jperez999:
rerun tests

nvtabular/dask/io.py
jperez999: @rjzamora I felt like since both to_iter and to_dd were using it, it would be easier to set in one place. But if we want that changed, I can make that happen no problem.

rjzamora:
@benfred - Any other requirements for this step, or can we merge?

benfred:
LGTM - thanks!

Issue #87, opened 2020-06-11
[FEA] Multi-GPU GroupBy Operator


benfred:
The existing GroupBy operator doesn't support multigpu operation - we should add support, similar to how we've done for the categorify op.

See also https://github.com/NVIDIA/NVTabular/issues/42 , we might be able to share a bunch of code between groupby and categorify since they are fundamentally doing a very similar operation.

PR #91, opened 2020-06-15
[REVIEW] Fix dask-based categorify logic for frequency-limited encoding

rjzamora:
While looking into #87 , I noticed that there is a bug in the Dask-based Categorify code.  This PR fixes the bug and improves the related test coverage.

The good news is that this fix does **not** seem to change the performance of the `dask-nvtabular-criteo-benchmark.py` benchmark.


rjzamora:
@benfred - Note that I will likely submit another PR to modify this "Categorify" code soonish, but I think it's still worth getting this fix in.

Issue #15, opened 2020-06-04
[FEA] Dedicated objects for columns_ctx, stats_ctx, phase, config


benfred:
<a href="https://github.com/alecgunny"><img src="https://avatars0.githubusercontent.com/u/14932242?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [alecgunny](https://github.com/alecgunny)**
_Tuesday Mar 24, 2020 at 19:28 GMT_
_Originally opened as https://github.com/rapidsai/recsys/issues/28_

----

**Is your feature request related to a problem? Please describe.**
Right now many of the central objects in a Workflow are constructed as lists or dicts with some expected structure. This can make parsing their meaning and use, as well as those of the constituent elements, difficult to understand when trying to debug or contribute. Examples include the columns_ctx, stats_ctx, phases, and the workflow config.

**Describe the solution you'd like**
Ideally, these objects would be replaced with dedicated objects with descriptive attributes and methods that make their functionality and components more clear. Methods on these objects could even simplify Workflow code by replacing methods which exist solely to update and retrieve information from these objects.

**Describe alternatives you've considered**
NamedTuples, or objects inheriting from NamedTuples, could be a simple way to ascribe fixed attributes, and could even maintain iterability to reduce short-term code updates.

benfred:
<a href="https://github.com/alecgunny"><img src="https://avatars0.githubusercontent.com/u/14932242?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [alecgunny](https://github.com/alecgunny)**
_Thursday Apr 23, 2020 at 19:04 GMT_

----

Expanding scope to include feedback on relationships between objects. Pseudo code demo can be found here https://github.com/alecgunny/pseudo_nvt (Even and Julio should have access already, let me know if anyone else wants to be added)

benfred:
<a href="https://github.com/alecgunny"><img src="https://avatars0.githubusercontent.com/u/14932242?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [alecgunny](https://github.com/alecgunny)**
_Friday May 01, 2020 at 15:22 GMT_

----

At its core, this is the gist of the sort of flow I think might make sense: putting less omnibus functionality into the workflow and breaking things up into separate function calls:
```
workflow = nvt.Workflow(...)

train_dataset = nvt.dataset('train.csv', batch_size=8192)
[...]

PR #94, opened 2020-06-16
[REVIEW] Workflow refactor

benfred:
In preparation for merging the dask/cudf workflow objects this is a first step
at simplifying some workflow implementation details:

* Removes the OperatorRegistry class, and the 'ops_args' member on the workflow
* Moves the master_task_list and task_sets from being member variables of the workflow class to being local variables of the load_config method
* Removes the operator registration code and stat_ops/df_ops/feat_ops  members
* Removes the 'to_cpu' parameter on the workflow class


benfred:
rerun tests

nvtabular/workflow.py
```diff
@@ -65,17 +65,7 @@ class BaseWorkflow:
...
-        New feature engineering operators can be added
-        later using the other functions
-    stat_ops : list of stats operator objects
-        Feature engineering ops objects such as
-        Categorify and Normalize. New feature engineering
-        operators can be added later using the other functions.
-    df_ops : list of data frame operator objects
-    to_cpu : bool, default True
-    config : bool
+    config : object
```
rjzamora, line 68: Maybe we can explain the `config` argument a bit more here?

nvtabular/workflow.py
```diff
@@ -85,21 +75,13 @@ def __init__(
         cat_names=None,
         cont_names=None,
         label_name=None,
-        feat_ops=None,
-        stat_ops=None,
-        df_ops=None,
-        to_cpu=True,
         config=None,
+        # TODO: move these out
```
rjzamora, line 79: Just to clarify - The plan is to turn "exporting" into a separate Op in the future?

nvtabular/workflow.py
benfred: I think we should actually make config a class in its own right (like Alec suggested here https://github.com/NVIDIA/NVTabular/issues/15).

nvtabular/workflow.py
benfred: Yes, we're planning on having separate 'export' ops in the future. The default 'export_path' does get used in the 'phases_export' function right now - though that could probably be removed too

nvtabular/workflow.py
```diff
@@ -641,23 +556,16 @@ def run_ops_for_phase(self, gdf, tasks, record_stats=True):
...
-                )
-            elif op._id in self.df_ops:
-                gdf = self.df_ops[op._id].apply_op(
-                    gdf,
-                    self.columns_ctx,
-                    cols_grp,
-                    target_cols=target_cols,
-                    stats_context=self.stats,
-                )
+
```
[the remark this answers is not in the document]
benfred: I think that this is ok now - the tests + examples seem to all work with these changes.

benfred:
> It's not related to any change you made, but one thing I noticed while looking this over is that the `config` argument seems to be a missed opportunity. It might be nice to consolidate the `load_stats` and `load_config` functionality so that the user can pass in a single file (as `config`) to specify _both_ the workflow configuration and known statistics. An obvious application would be "online-only" preprocessing.

Lets examine this in a future PR - I agree that we probably want to make online-only / inference support seamless, but have some concerns about serializing the ops that are part of the current config object.

PR #96, opened 2020-06-17
[WIP] Use pynvml when get_memory_info not supported

rjzamora:
Using the latest version of rmm will lead to test failures in NVTabular master.  I personally like to use the latest RAPIDS libraries, so It would be nice to add a "workaround" for the (hopefully temporary) lack of rmm functionality.  This PR adds a change in the dask-based IO code to "fall back" to using `pynvml` when numba (which uses rmm) returns a `NotImplementedError`.

Any chance this approach might be useful elsewhere in NVTabular? (cc @jperez999 @benfred )

### DIFF
--- a/nvtabular/dask/io.py
+++ b/nvtabular/dask/io.py
@@ -185,7 +185,19 @@ class DaskDataset:
                     "Using very large partitions sizes for Dask. "
                     "Memory-related errors are likely."
                 )
-            part_size = int(cuda.current_context().get_memory_info()[1] * part_mem_fraction)
+
+            try:
+                part_size = int(cuda.current_context().get_memory_info()[1] * part_mem_fraction)
+            except NotImplementedError:
+                # TODO: Remove this block when rmm approach is fully supported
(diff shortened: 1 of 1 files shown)


jperez999:
run tests

benfred:
We should definitely standardize on how we get GPU memory info - especially if we can't get the info from numba or RMM anymore .   There are a couple of other places we are currently using rmm, but some of those will go away once we finish the dask migration.

I think it would be better if we could get usage from RMM since that will show free space as well as total memory (opened an issue here https://github.com/rapidsai/rmm/issues/390), but if that isn't possible we should switch to something like this. 

For this pynvml change, can we just default to using pynvml instead of numba? Like instead of trying to use numba and then failing use pynvml, just use it all the time.  Also, we should probably list pynvml as a dependency in requirements.txt and the conda recipe.

rjzamora:
I will probably include this change in any PRs I submit (because my environments are typically using latest cudf), so closing here.

PR #98, opened 2020-06-17
[Review] Torch dl async update

jperez999:
This PR is put together to handle the async loading of chunks, for pytorch.


nvtabular/torch_dataloader.py
```diff
@@ -233,14 +219,36 @@ def __init__(
...
+    def load_chunk(self):
+        with self.lock:
+            try:
+                chunk = next(self.itr)
+            except StopIteration:
+                return
+            chunk = create_tensors_plain(
+                chunk, cat_cols=self.cat_cols, cont_cols=self.cont_cols, label_cols=self.label_cols
+            )
+            self.buffer.put(chunk)
```
benfred, line 251: I don't like creating a new thread to read in a single chunk here. I also think this code is safe to call from multiple threads because of you are storing mutable state on the class with the itr/lock/buffer members (and lines like ```self.itr = iter(self.itr)``` probably mean that you can only iterate over this class once?).

How about something like this instead :

```suggestion
        buffer = queue.Queue(1)
        threading.Thread(target=self._load_chunks, args=(buffer,)).start()

        for _ in range(self.num_chunks):
            chunk = buffer.get()
            yield from TensorItr(
                chunk,
                batch_size=self.batch_size,
                cat_cols=self.cat_cols,
                cont_cols=self.cont_cols,
                label_cols=self.label_cols,
            )

    def _load_chunks(self, output):
        for chunk in self.itr:
            chunk = create_tensors_plain(
                chunk, cat_cols=self.cat_cols, cont_cols=self.cont_cols, label_cols=self.label_cols
            )
            output.put(chunk)
```

This only creates 1 background thread,  and removes the lock/buffer member variables - making the buffer a local variable, and using thread locking from the queue. To limit to 1 background read of a chunk at a time we cap the buffer size to 1.

tests/unit/test_torch_dataloader.py
```diff
@@ -288,16 +287,30 @@ def test_gpu_dl(tmpdir, datasets, batch_size, gpu_memory_frac, engine):
...
     num_rows, num_row_groups, col_names = cudf.io.read_parquet_metadata(tar_paths[0])
     rows = 0
+    # works with iterator alone, needs to test inside torch dataloader
     for idx, chunk in enumerate(data_itr):
-        rows += len(chunk)
+        assert float(df_test.iloc[rows][0]) == float(chunk[0][0][0])
+        rows += len(chunk[0])
         del chunk
-
+    # assert_eq(df_test
```
benfred, line 299:
```suggestion
```

Issue #34, opened 2020-06-04
[OP] Add Min-Max scaling under Normalize() Class in ops.py


benfred:
<a href="https://github.com/rnyak"><img src="https://avatars1.githubusercontent.com/u/16246900?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [rnyak](https://github.com/rnyak)**
_Tuesday May 19, 2020 at 16:27 GMT_
_Originally opened as https://github.com/rapidsai/recsys/issues/145_

----

**Is your operator request related to a problem? Please describe.**
Min-Max scaling is one of the most common scaling techniques that is used for data pre-processing.

**Describe the solution you'd like**
A clear and concise description of the operation you'd like to perform on the column.  Please include:
 - Type: Preprocessing
 - input column type(s): Continuous numeric 
 - output column type(s): Float numeric
 - Expected transformation of the data after application: Data is transformed to be within [0-1] range.

**Optional: Describe operation stages in detail***
Statistics per chunk: Ex. compute the min and max of the column
Statistics combine: NA
Apply: (value-min)/(max-min)

**Additional context**

Example code for Sklearn `MinMaxScaler` class can be found here:
https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/preprocessing/_data.py#L200

gabrielspmoreira:
For scaling ops, I believe we should store the original statistics in a metadata file and output them after the pipeline, i.e. storing the original min and max values for Min-Max and the original mean and std for ZNorm.
This is required for inference (we should apply the same scaling for prediction) and also to recover features in the original scale (e.g. converting the output of a regression model to a prediction in the same scale as the label)

benfred:
@gabrielspmoreira We have the ability to load/save stats (including min/max ) already https://github.com/NVIDIA/NVTabular/blob/master/nvtabular/workflow.py#L862 . Its definitely required =).

@rynak for min-max scaling, is the correct calculation ```(value-mean)/(max-min)``` ? That won't transform values to a [0-1] range, I think what we should be looking for is ```(value-min)/(max-min)``` instead (substracting the min rather than mean)

rnyak:
@benfred yes correct, fixing the typo there.  it should be :

`(value-min)/(max-min)`

PR #99, opened 2020-06-18
[REVIEW] Min-Max scaling under Normalize() Class

albert17:
Implements issue #34 

Files modified: ops.py

I am adding a variable to choose between mean-std and min-max, this affects the Normalize interface, not sure if we want this or something else. If we want this, I will be updating the different examples/notebooks that call Normalize.

Also, I didn't find any test for Normalize. We are not doing anything there?

@benfred for the questions.

### DIFF
--- a/nvtabular/ops.py
+++ b/nvtabular/ops.py
@@ -751,7 +751,7 @@ class Normalize(DFOperator):
     default_in = CONT
     default_out = CONT
 
-    def __init__(self, ntype="std", columns=None, preprocessing=True, replace=True):
+    def __init__(self, ntype="mean_std", columns=None, preprocessing=True, replace=True):
         super().__init__(columns=columns, preprocessing=preprocessing, replace=replace)
         self.ntype = ntype
 
--- a/nvtabular/ops.py
+++ b/nvtabular/ops.py
@@ -751,6 +751,10 @@ class Normalize(DFOperator):
     default_in = CONT
     default_out = CONT
 
+    def __init__(self, ntype="std", columns=None, preprocessing=True, replace=True):
+        super().__init__(columns=columns, preprocessing=preprocessing, replace=replace)
+        self.ntype = ntype
+
     @property
     def req_stats(self):
         return [Moments()]
@@ -760,7 +764,10 @@ class Normalize(DFOperator):
         cont_names = target_columns
         if not cont_names or not stats_context["stds"]:
             return
-        gdf = self.apply_mean_std(gdf, stats_context, cont_names)
+        if self.ntype == "mean_std":
+            gdf = self.apply_mean_std(gdf, stats_context, cont_names)
+        elif self.ntype == "min_max":
+            gdf = self.apply_min_max(gdf, stats_context, cont_names)
         return gdf
 
     def apply_mean_std(self, gdf, stats_context, cont_names):
@@ -770,6 +777,23 @@ class Normalize(DFOperator):
                 new_col = f"{name}_{self._id}"
                 new_gdf[new_col] = (gdf[name] - stats_context["means"][name]) / (
                     stats_context["stds"][name]
+                )
+                new_gdf[new_col] = new_gdf[new_col].astype("float32")
+        return new_gdf
+
+    def apply_min_max(self, gdf, stats_context, cont_names):
+        # calculate min and max
+        min_max = {}
+        min_max["min"] = gdf[cont_names].min()
+        min_max["max"] = gdf[cont_names].max()
+        # create and fill new dataset
+        new_gdf = cudf.DataFrame()
+        for name in cont_names:
+            dif = min_max["max"][name] - min_max["min"][name]
+            if dif > 0:
+                new_col = f"{name}_{self._id}"
+                new_gdf[new_col] = (gdf[name] - stats_context["means"][name]) / (
+                    dif
                 )
                 new_gdf[new_col] = new_gdf[new_col].astype("float32")
         return new_gdf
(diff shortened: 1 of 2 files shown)


nvtabular/ops.py
```diff
@@ -787,6 +794,23 @@ def apply_mean_std(self, gdf, stats_context, cont_names):
                 new_gdf[new_col] = new_gdf[new_col].astype("float32")
         return new_gdf
 
+    def apply_min_max(self, gdf, stats_context, cont_names):
+        # calculate min and max
+        min_max = {}
+        min_max["min"] = gdf[cont_names].min()
+        min_max["max"] = gdf[cont_names].max()
```
benfred, line 801: The 'gdf' here isn't the full dataset - its just a piece thats loaded up into GPU memory. As such we can't calculate the min/max values here and need to do in a separate StatOperator

We do have a 'MinMax' op already that will calculate these statistics across the entire dataframe though: https://github.com/NVIDIA/NVTabular/blob/ddf58ecc8603ec09129cad420adf9916faa928c8/nvtabular/ops.py#L212-L253 . 

You should be able to use in the same way as we do with the Moments (getting values through the stats_context).

nvtabular/ops.py
```diff
@@ -773,7 +777,10 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None
         cont_names = target_columns
         if not cont_names or not stats_context["stds"]:
             return
-        gdf = self.apply_mean_std(gdf, stats_context, cont_names)
+        if self.ntype == "mean_std":
+            gdf = self.apply_mean_std(gdf, stats_context, cont_names)
+        elif self.ntype == "min_max":
+            gdf = self.apply_min_max(gdf, stats_context, cont_names)
```
benfred, line 783: I think we should create a new operator rather than have both minmax scaling and mean/std normalization in the same class. @jperez999 what do you think?

jperez999: Yes, each of those stats should be its own operator. The idea is to make stat operators that represent a single statistic. This allows users to clearly understand what stats are available in a workflow. It will also ensure minimal workflow size when all repeated operators are removed.

albert17:
I am using the Moments() to get the min max values.

I created a new class for the min max method.

What about unit tests for Normalize?

nvtabular/ops.py
```diff
@@ -788,6 +790,46 @@ def apply_mean_std(self, gdf, stats_context, cont_names):
...
+        gdf = self.apply_min_max(gdf, stats_context, cont_names)
+        return gdf
+
+    def apply_min_max(self, gdf, stats_context, cont_names):
+        new_gdf = cudf.DataFrame()
+        for name in cont_names:
+            dif = stats_context["maxs"][name] - stats_context["mins"][name]
+            if dif > 0:
+                new_col = f"{name}_{self._id}"
+                new_gdf[new_col] = (gdf[name] - stats_context["means"][name]) / (dif)
```
benfred, line 828: I don't think this is the right formula for min-max scaling - I believe it should be 
```suggestion
                new_gdf[new_col] = (gdf[name] - stats_context["mins"][name]) / (dif)
```
to actually scale to a [0, 1] range.  (I realize you implemented what Ronay asked for in the linked issue, asked a follow up question here to verify https://github.com/NVIDIA/NVTabular/issues/34#issuecomment-646850202).

If we do need the mean stat though, you will have to add Moments to the 'req_stats' return value (otherwise it won't get pulled in automatically).

albert17:
Fixed the min-max formula.

Added unit test for Normalize (mean-std) and NormalizeMinMax

tests/unit/test_ops.py
```diff
@@ -406,3 +406,119 @@ def test_dropna(tmpdir, datasets, engine="parquet"):
...
+
+    processor = nvt.Workflow(
+        cat_names=cat_names,
+        cont_names=cont_names,
+        label_name=label_name,
+        config=config,
+        to_cpu=False,
+    )
+
+    processor.update_stats(data_itr)
```
benfred, line 454: Can you update this to be against the latest master? I refactored the workflow/unittests last week . The workflow object no longer has a 'to_cpu' flag (#94) and the unittests have fixtures for creating the datasets / test dataframe (#106) to avoid some of the boilerplate in creating tests:

With these changes this should be something like:

```suggestion
def test_normalize(tmpdir, df, dataset, gpu_memory_frac, engine, op_columns):
    cat_names = ["name-cat", "name-string"] if engine == "parquet" else ["name-string"]
    cont_names = ["x", "y", "id"]
    label_name = ["label"]

    config = nvt.workflow.get_new_config()
    config["PP"]["continuous"] = [ops.Moments()]

    processor = nvt.Workflow(
        cat_names=cat_names,
        cont_names=cont_names,
        label_name=label_name,
        config=config,
    )

    processor.update_stats(dataset)
```
(and likewise for test_normalize_minmax)

nvtabular/ops.py
```diff
@@ -788,6 +790,46 @@ def apply_mean_std(self, gdf, stats_context, cont_names):
...
+        return gdf
+
+    def apply_min_max(self, gdf, stats_context, cont_names):
+        new_gdf = cudf.DataFrame()
+        for name in cont_names:
+            dif = stats_context["maxs"][name] - stats_context["mins"][name]
+            if dif > 0:
+                new_col = f"{name}_{self._id}"
+                new_gdf[new_col] = (gdf[name] - stats_context["mins"][name]) / (dif)
+                new_gdf[new_col] = new_gdf[new_col].astype("float32")
```
benfred, line 829: How are we going to handle a column that only contains a constant value? I think we still want to normalize it to be between 0 and 1, and include in the output.

How about if dif == 0 we just set to a constant value in [0,1]  - maybe 0.5?

albert17:
Updated tests, and updated op to also normalize constant values

PR #105, opened 2020-06-19
[REVIEW] Avoid future pyarrow deprecation problems

### LINKED ISSUE #103: [DEP] Handle future pyarrow-API deprecations
**Describe the bug**
Note that this is not really a "bug" **yet**, but will be in the future :)

Recent/future efforts in pyarrow will eventually cause problems for some of our dask-based I/O logic.  The critical changes include:

- Deprecation of [pyarrow.compat](https://github.com/apache/arrow/blob/cfe114478d0c90b0822aec1c8c24635c3d35330e/python/pyarrow/compat.py#L24), which both dask and NVTabuar use for "random" file naming.
- Replacement of the `ParquetDataset` API with a new [pyarrow.dataset](https://github.com/apache/arrow/blob/cfe114478d0c90b0822aec1c8c24635c3d35330e/python/pyarrow/dataset.py#L494) API.  For now, much of the original functionality is still supported.  However, it seems that the plan is to change the legacy API into a [thin dataset-API shim layer](https://github.com/apache/arrow/blob/cfe114478d0c90b0822aec1c8c24635c3d35330e/python/pyarrow/parquet.py#L1374) (and to ultimately stop supporting and deprecate `ParquetDataset`)

**TODO**:
- [ ] Stop using `pyarrow.compat` API (`guid` for file naming)
- [ ] Switch to `pyarrow.dataset` API for newer pyarrow versions (Allow `ParquetDataset` for `pyarrow<0.17`)

Note that these same steps need to be addressed in the upstream `dask`/`cudf` libraries, but those projects are waiting on a bit more functionality from the *new* API (`nvtabular` can make the changes sooner).

rjzamora:
Closes #103

Adds logic to use `pyarrow.datasets` API for newer versions of `pyarrow` -  Note that these changes will **not** be tested by CI, because we are stuck on an older version of cudf (which is tied to pyarrow 0.15).  The good news is that the performance of the new API seems to be the same for the `dask-nvtabular-criteo-benchmark.py` benchmark.  Also, it should make it easier for us to add filtering to the `DaskDataset` object (so that we can completely avoid reading in row groups that do not meet certain filter criteria).

### DIFF
--- a/nvtabular/dask/io.py
+++ b/nvtabular/dask/io.py
@@ -38,6 +38,11 @@ from fsspec.utils import stringify_path
 
 from nvtabular.io import _shuffle_gdf
 
+try:
+    import pyarrow.dataset as pa_ds
+except ImportError:
+    pa_ds = False
+
 
 class WriterCache:
     def __init__(self):
@@ -257,23 +262,40 @@ class ParquetDatasetEngine(DatasetEngine):
         # TODO: Improve dask_cudf.read_parquet performance so that
         # this class can be slimmed down.
         super().__init__(*args)
-        self._metadata, self._base = self.get_metadata()
-        self._pieces = None
+
+        if pa_ds:
+            # Use pyarrow.dataset API for "newer" pyarrow versions
+            self._legacy = False
+            self._pieces = None
+            self._metadata, self._base = defaultdict(int), ""
+            path0 = None
+            ds = pa_ds.dataset(self.paths, format="parquet")
+            # TODO: Allow filtering while accessing fragments.
+            #       This will require us to specify specific row-group indices
+            for file_frag in ds.get_fragments():
+                if path0 is None:
+                    path0 = file_frag.path
+                for rg_frag in file_frag.get_row_group_fragments():
+                    self._metadata[rg_frag.path] += len(list(rg_frag.row_groups))
+        else:
+            # Use pq.ParquetDataset API for <0.17.1
+            self._legacy = True
+            self._metadata, self._base = self.get_metadata()
+            self._pieces = None
+            if row_groups_per_part is None:
+                file_path = self._metadata.row_group(0).column(0).file_path
+                path0 = (
+                    self.fs.sep.join([self._base, file_path])
+                    if file_path != ""
+                    else self._base  # This is a single file
+                )
+
         if row_groups_per_part is None:
-            # TODO: Use `total_byte_size` metadata if/when we figure out how to
-            #       correct for apparent dict encoding of cat/string columns.
-            file_path = self._metadata.row_group(0).column(0).file_path
-            path0 = (
-                self.fs.sep.join([self._base, file_path])
-                if file_path != ""
-                else self._base  # This is a single file
-            )
             rg_byte_size_0 = (
                 cudf.io.read_parquet(path0, row_group=0).memory_usage(deep=True, index=True).sum()
             )
-            self.row_groups_per_part = int(self.part_size / rg_byte_size_0)
-        else:
-            self.row_groups_per_part = int(row_groups_per_part)
+            row_groups_per_part = self.part_size / rg_byte_size_0
+        self.row_groups_per_part = int(row_groups_per_part)
         assert self.row_groups_per_part > 0
 
     @property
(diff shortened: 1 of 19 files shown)


nvtabular/dask/io.py
```diff
@@ -184,7 +195,19 @@ def __init__(
...
+            except NotImplementedError:
+                # TODO: Remove this block when rmm approach is fully supported
+                import pynvml
+
+                pynvml.nvmlInit()
+                part_size = int(
+                    pynvml.nvmlDeviceGetMemoryInfo(pynvml.nvmlDeviceGetHandleByIndex(0)).total
+                    * part_mem_fraction
+                )
+                pynvml.nvmlShutdown()
```
benfred, line 210: Rather than have this block of code here and a similar block in the examples script - can we pull out to a common function?

Also - is there a reason to even try the cuda.current_context() code here, rather than just always getting memory info from pynvml?

nvtabular/dask/io.py
```diff
@@ -239,27 +262,48 @@ class ParquetDatasetEngine(DatasetEngine):
         Dask-based version of cudf.read_parquet.
     """
 
-    def __init__(self, *args, row_groups_per_part=None):
+    def __init__(self, *args, row_groups_per_part=None, legacy=False):
```
benfred, line 265: Whats the use case for this legacy parameter?

nvtabular/dask/io.py
rjzamora: Agreed - I haven't been tracking how likely the numba/rmm issue is to be fixed, so I was assuming this "workaround" may be removed soon.  However, we might as well make the pynvml approach a bit more standardized if it is here to stay :)

nvtabular/dask/io.py
rjzamora: This is just a switch that someone running with 0.17.1 can use to easily compare legacy pyarrow vs their new datasets API.  The argument is certainly not necessary, but I just left it in for now.

rjzamora:
> Eventually I think we should just drop support for pyarrow < 0.17 (but think that also means dropping support for cudf < 0.15 ?).

Right - We will need to keep the legacy behavior around until we drop rapids 0.14 support.

Another concern I have with the new pyarrow API is that I am not sure it supports all `fsspec` filesystem types - so we may want to keep legacy support in place until we are sure it meets our needs.

nvtabular/dask/io.py
benfred: The current proposal seems to indicate that you can enable a tracking resource adaptor to get a sense of how much memory is free in RMM: https://github.com/rapidsai/rmm/issues/415 .

I think right now we should standardize on the pynvml approach- but it doesn't need to be part of this PR. I think many of the other calls to rmm.get_info are going to go away in the next week or so as we finish the dask migration

benfred:
> Another concern I have with the new pyarrow API is that I am not sure it supports all `fsspec` filesystem types - so we may want to keep legacy support in place until we are sure it meets our needs.

I think for 0.2 we will need s3, gcs and hdfs support (we've had requests for all 3 from different customers).  Do you know if the new pyarrow api supports all those filesystem types?

rjzamora:
> I think for 0.2 we will need s3, gcs and hdfs support (we've had requests for all 3 from different customers). Do you know if the new pyarrow api supports all those filesystem types?

I don't believe gcs is supported, but the others should be fine...

nvtabular/dask/io.py
```diff
@@ -239,27 +262,48 @@ class ParquetDatasetEngine(DatasetEngine):
...
+            self._pieces = None
+            self._metadata, self._base = defaultdict(int), ""
+            path0 = None
+            ds = pa_ds.dataset(self.paths, format="parquet")
+            # TODO: Allow filtering while accessing fragments.
+            #       This will require us to specify specific row-group indices
+            for file_frag in ds.get_fragments():
+                if path0 is None:
+                    path0 = file_frag.path
+                for rg_frag in file_frag.get_row_group_fragments():
```
rjzamora, line 286: It seems pyarrow already replaced `get_row_group_fragments` with `split_by_row_group`... I guess using pyarrow.dataset is definitely living "on the bleeding edge" at the moment.

nvtabular/dask/io.py
benfred: Done in #107

nvtabular/dask/io.py
benfred: Should we update here before merging?

PR #106, opened 2020-06-20
[REVIEW] Fix reading CSV files with no header

benfred:
When reading in CSV files with no header we ended up ignoring the first line of
the file. This caused us to lose one row per CSV file we were reading in. This is a one line fix in nvtabular/io.py - but we also had the same problem in each unittest. Fix the unittest code by consolidating common code into test fixtures (which will also make the dask transition easier
since we'll only have to update in one place).


jperez999:
run tests

PR #107, opened 2020-06-22
Move queries for device memory size into new _device_mem_size utility

rjzamora:
Adds a new utility function to `nvtabular/io.py` to consolidate the logic needed to query the free/total device memory.  For newer versions of rmm, pynvml will be used to return the "total" device memory, even when the "free" memory is requested.  When this happens, `_device_mem_size` will print a warning message, but will not raise an error.  Note that pynvml **can** return the "free" memory, but this is probably not the number you want if you are using an rmm pool (it wouldn't consider any memory in the rmm pool to be free, even if that memory is available to your application).

### DIFF
--- a/examples/criteo-example.ipynb
+++ b/examples/criteo-example.ipynb
@@ -96,7 +96,7 @@
     }
    ],
    "source": [
-    "rmm.reinitialize(pool_allocator=True, initial_pool_size=0.8 * nvt.io._device_mem_size(kind='free'))"
+    "rmm.reinitialize(pool_allocator=True, initial_pool_size=0.8 * nvt.io.device_mem_size(kind='free'))"
    ]
   },
   {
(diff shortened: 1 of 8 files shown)


examples/criteo-example.ipynb
```diff
@@ -96,7 +96,7 @@
     }
    ],
    "source": [
-    "rmm.reinitialize(pool_allocator=True, initial_pool_size=0.8 * rmm.get_info().free)"
+    "rmm.reinitialize(pool_allocator=True, initial_pool_size=0.8 * nvt.io._device_mem_size(kind='free'))"
```
benfred, line 99: If we're exposing this as part of our public API (by using it in our notebooks etc) - I think the _device_mem_size function shouldn't have a leading underscore.

rjzamora: > If we're exposing this as part of our public API (by using it in our notebooks etc) - I think the _device_mem_size function shouldn't have a leading underscore.

Nice - I had the same thought, and was making this change :)

benfred:
Thanks!

PR #109, opened 2020-06-23
[REVIEW] Metadata format for Hugectr Parquet

albert17:
@oyilmaz-nvidia changes for metadata json format,


nvtabular/workflow.py
```diff
@@ -480,7 +480,7 @@ def get_final_cols_names(self, col_type):
         col_names = []
         for c_names in self.columns_ctx[col_type].values():
             col_names.extend(c_names)
```
oyilmaz-nvidia, line 482: Maybe you can directly add into a set here, then convert to a list.

nvtabular/io.py
```diff
@@ -405,6 +406,12 @@ def set_col_names(self, labels, cats, conts):
         self.labels = labels
         self.column_names = labels + conts
 
+    def set_col_idx(self, gdf):
+        counter = 0
+        for x in gdf.columns.values:
+            self.col_idx[str(x)] = counter
+            counter += 1
+
```
benfred, line 414: you should use pythons ```enumerate``` for this instead of maintaining a count yourself
```suggestion
        for i, x in enumerate(gdf.columns.values):
            self.col_idx[str(x)] = i
```

nvtabular/io.py
```diff
@@ -380,6 +380,7 @@ def __init__(
         self.column_names = None
         if labels and conts:
             self.column_names = labels + conts
+        self.col_idx = {}
```
benfred, line 383: If this is only used for HugeCTR it should be a member of the HugeCTRWriter class

nvtabular/io.py
```diff
@@ -405,6 +406,12 @@ def set_col_names(self, labels, cats, conts):
         self.labels = labels
         self.column_names = labels + conts
 
+    def set_col_idx(self, gdf):
```
benfred, line 409: Why do we need to use the gdf for this? Can we use the cats/conts/labels columns to define this mapping instead? I'm asking because we now need 3 function calls to initialize the hugectr writer properly (constructor, set_column_names, set_col_idx), and the API is pretty unwieldy

nvtabular/workflow.py
```diff
@@ -477,10 +477,10 @@ def get_final_cols_names(self, col_type):
         -----------
         col_type : str
         """
-        col_names = []
+        col_names = set()
         for c_names in self.columns_ctx[col_type].values():
-            col_names.extend(c_names)
-        return col_names
+            col_names.add(c_names)
+        return list(col_names)
```
benfred, line 483: I don't think this is correct - have you tested this? 

I think you will get an error saying 'unhashable type list' or similar when running.  ('add' on a set object isn't  the same operation as 'extend' on a list - you probably meant for ```col_names.update(c_names)``` instead).

Also can you add a unittest for this function (and potentially even the whole writing ctr workflow)?

nvtabular/io.py
```diff
@@ -473,14 +480,28 @@ def _write_thread(self):
...
-        data["cats_name"] = self.cats
-        data["conts_name"] = self.conts
-        data["conts_labels"] = self.labels
+            data["file_stats"].append(
+                {"file_name": f"{i}.parquet", "num_rows": self.num_samples[i]}
+            )
+        # cats
+        data["cats"] = []
+        for c in self.cats:
+            data["cats"].append((c, self.col_idx[c]))
```
benfred, line 495: Why do we need the col_idx data? Can't the reader just figure this out from either the parquet metadata or from the ordering of the cats/conts/labels columns?

nvtabular/io.py
albert17: It is only used for HugeCTR, but only when usign Parquet format. So I can add it as a memeber of ParquetWriter Class.

nvtabular/io.py
albert17: We want to know in which order are the columns in the gdf, so cats/conts/labels list cannot provide that information.

I agree with the API. I can gdf as a parameter of set_col_names and do all in that function. What do you think?

jperez999:
rerun tests

nvtabular/workflow.py
albert17: Sorry. My bad, I changed from list to set and I didn't expect that problem. I will be using update.

I am adding unittest.

nvtabular/io.py
albert17: They need it on the hugectr side when reading the data, it looks like they cannot do it and they need the index.

nvtabular/io.py
benfred: How about just lazily initializing this? Like initialize on the first call to add_data if its not initialized already?

albert17: Sounds good to me. @oyilmaz-nvidia for confirming.

nvtabular/workflow.py
albert17: This is that I have done so far for [unit test](https://github.com/NVIDIA/NVTabular/blob/c96f8dfdd5351203fdddf3a7afe89f273901e7d5/tests/unit/test_io.py#L91).

I don't know how to deal exactly with reading/writing files. I should be checking that files have been written correctly to ensure correct execution.

I am getting a deadlock when running this test that I assume it is related with that, because I have been running a program successfully.

I haven't seen any similar to follow the same format, but I may be missing something.

nvtabular/io.py
oyilmaz-nvidia: Yeah, we can do that as well.

jperez999:
rerun tests

albert17:
@benfred added discussed changes and adding unit test for hugectr.

albert17:
rerun tests

PR #110, opened 2020-06-23
[REVIEW] Refactor get_emb_sz

### LINKED ISSUE #27: [FEA] Move or remove get_emb_sz method on Categorify op
<a href="https://github.com/alecgunny"></img></a> **Issue by [alecgunny](https://github.com/alecgunny)**
_Wednesday May 13, 2020 at 16:53 GMT_
_Originally opened as https://github.com/rapidsai/recsys/issues/121_

----

It's not entirely clear to me why `Categorify.get_emb_sz` and its associated helper methods are included as methods on the `Categorify` object, especially when nothing in the code actually uses attributes from the object itself (with the exception of the `self.embed_sz`, which gets set rather than read from):
```python
    def get_emb_sz(self, encoders, cat_names):
        work_in = {}
        for key in encoders.keys():
            work_in[key] = encoders[key] + 1
        # sorted key required to ensure same sort occurs for all values
        ret_list = [
            (n, self.def_emb_sz(work_in, n))
            for n in sorted(cat_names, key=lambda entry: entry.split("_")[0])
        ]
        return ret_list

    def emb_sz_rule(self, n_cat: int) -> int:
        return min(16, round(1.6 * n_cat ** 0.56))

    def def_emb_sz(self, classes, n, sz_dict=None):
        """Pick an embedding size for `n` depending on `classes` if not given in `sz_dict`.
        """
        sz_dict = sz_dict if sz_dict else {}
        n_cat = classes[n]
        sz = sz_dict.get(n, int(self.emb_sz_rule(n_cat)))  # rule of thumb
        self.embed_sz[n] = sz
        return n_cat, sz
```
While I'm personally of the opinion that it's not our library's job to be providing rules of thumb for building deep learning models, just to provide the data for whatever rules the user wants to use, if we're intent on having this function somewhere it seems like it would be better served as a standalone function of a single encoder or as a property of the encoder itself:
```python
emb_szs = [get_embed_sz(proc.stats["categories"][column]) for column in proc.columns_ctx["categorical"]["base"]]

# or
emb_szs = [proc.stats["categories"][column].embed_sz[1] for column in proc.columns_ctx["categorical"]["base"]]
```

benfred:
I recently broke the example notebooks using get_emb_sz, since we no longer have easy access to the Categorify op.

Fix by moving get_emb_sz from a method on the categorify op to a plain function.
Also rename to get_embedding_size, and consolidate the couple of different
spots we were sorted category names to a single function (get_embedding_order).
This fixes  #27

### DIFF
--- a/nvtabular/ops.py
+++ b/nvtabular/ops.py
@@ -1248,7 +1248,7 @@ def get_embedding_order(cat_names):
     cat_names : list of str
         names of the categorical columns
     """
-    return sorted(cat_names, key=lambda entry: entry.split("_")[0])
+    return sorted(cat_names)
 
 
 def get_embedding_size(encoders, cat_names):
--- a/nvtabular/ops.py
+++ b/nvtabular/ops.py
@@ -1154,7 +1154,6 @@ class Categorify(DFOperator):
         Replaces the transformed column with the original input
         if set Yes
     cat_names :
-    embed_sz :
     """
 
     default_in = CAT
@@ -1171,7 +1170,6 @@ class Categorify(DFOperator):
         preprocessing=True,
         replace=True,
         cat_names=None,
-        embed_sz=None,
         out_path=None,
         split_out=None,
         na_sentinel=None,
@@ -1186,7 +1184,6 @@ class Categorify(DFOperator):
         self.gpu_mem_util_limit = gpu_mem_util_limit
         self.gpu_mem_trans_use = gpu_mem_trans_use
         self.cat_names = cat_names if cat_names else []
-        self.embed_sz = embed_sz if embed_sz else {}
         self.out_path = out_path or "./"
         self.split_out = split_out
         self.na_sentinel = na_sentinel or 0
@@ -1242,30 +1239,32 @@ class Categorify(DFOperator):
                 new_gdf[new_col] = new_gdf[new_col].astype(self.dtype, copy=False)
         return new_gdf
 
-    def get_emb_sz(self, encoders, cat_names):
-        # sorted key required to ensure same sort occurs for all values
-        ret_list = [
-            (n, self.def_emb_sz(encoders, n))
-            for n in sorted(cat_names, key=lambda entry: entry.split("_")[0])
-        ]
-        return ret_list
 
-    def emb_sz_rule(self, n_cat: int) -> int:
-        return min(16, round(1.6 * n_cat ** 0.56))
+def get_embedding_order(cat_names):
+    """ Returns a consistent sorder order for categorical variables
 
-    def def_emb_sz(self, classes, n, sz_dict=None):
-        """Pick an embedding size for `n` depending on `classes` if not given in `sz_dict`.
-        """
-        sz_dict = sz_dict if sz_dict else {}
-        n_cat = classes[n]
-        sz = sz_dict.get(n, int(self.emb_sz_rule(n_cat)))  # rule of thumb
-        self.embed_sz[n] = sz
-        return n_cat, sz
+    Parameters
+    -----------
+    cat_names : list of str
+        names of the categorical columns
+    """
+    return sorted(cat_names, key=lambda entry: entry.split("_")[0])
 
-    def get_embeddings(self, encoders, cat_names):
-        embeddings = {}
-        for col in sorted(cat_names):
-            path = encoders[col]
-            num_rows, _, _ = cudf.io.read_parquet_metadata(path)
-            embeddings[col] = (num_rows, self.emb_sz_rule(num_rows))
-        return embeddings
+
+def get_embedding_size(encoders, cat_names):
+    """ Returns a suggested size of embeddings based off cardinality of encoding categorical
+    variables
+    Parameters
+    -----------
+    encoders : dict
+        The encoding statistics of the categorical variables (ie. from workflow.stats["categories"])
+    cat_names : list of str
+        names of the categorical columns
+    """
+    # sorted key required to ensure same sort occurs for all values
+    ret_list = [(n, _emb_sz_rule(encoders[n])) for n in get_embedding_order(cat_names)]
+    return ret_list
+
+
+def _emb_sz_rule(n_cat: int) -> int:
+    return n_cat, int(min(16, round(1.6 * n_cat ** 0.56)))
--- a/examples/criteo-example.ipynb
+++ b/examples/criteo-example.ipynb
@@ -44,7 +44,7 @@
     "import torch\n",
     "import rmm\n",
     "import nvtabular as nvt\n",
-    "from nvtabular.ops import Normalize, FillMissing, Categorify, Moments, Median, Encoder, LogOp, ZeroFill\n",
+    "from nvtabular.ops import Normalize, FillMissing, Categorify, Moments, Median, Encoder, LogOp, ZeroFill, get_embedding_size\n",
     "from nvtabular.torch_dataloader import DLDataLoader, TorchTensorBatchDatasetItr, create_tensors_plain\n",
     "\n",
     "# tools for training\n",
@@ -304,7 +304,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "embeddings = [x[1] for x in proc.df_ops['Categorify'].get_emb_sz(proc.stats[\"categories\"], proc.columns_ctx['categorical']['base'])]"
+    "embeddings = [x[1] for x in get_embedding_size(proc.stats[\"categories\"], proc.columns_ctx['categorical']['base'])]"
    ]
   },
   {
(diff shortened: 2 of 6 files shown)


nvtabular/ops.py
```diff
@@ -1242,30 +1239,32 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context={}):
...
-            path = encoders[col]
-            num_rows, _, _ = cudf.io.read_parquet_metadata(path)
-            embeddings[col] = (num_rows, self.emb_sz_rule(num_rows))
-        return embeddings
+    Parameters
+    -----------
+    cat_names : list of str
+        names of the categorical columns
+    """
+    return sorted(cat_names, key=lambda entry: entry.split("_")[0])
```
rdipietro, line 1251: Where is this strange sort key coming from? If this requirement is coming from some under-the-hood library, then are we relying on stable sorting to break ties? E.g. imagine column names `'user_feature_1'` and `'user_feature_2'`.

nvtabular/ops.py
```diff
@@ -1242,30 +1239,32 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context={}):
...
+
+
+def get_embedding_size(encoders, cat_names):
```
rdipietro, line 1254: This comment might be out of the scope of this refactor, but we should try to make this function cleaner and clearer.

Right now, the function accepts a `dict` mapping names to cardinalities, along with a list of names. Then these are both used to simply pull out the cardinalities. Why not instead just let this function accept a list of integers (the cardinalities themselves)? This encourages unreadable code such as this (repeated throughout the notebooks): `get_embedding_size(proc.stats["categories"], proc.columns_ctx["categorical"]["base"])`

nvtabular/ops.py
rdipietro: @benfred not sure if you're notified without this mention

nvtabular/ops.py
rdipietro: @benfred not sure if you're notified without this mention

nvtabular/ops.py
jperez999: These are good points, the delimiter chosen is bad and the sort is not the best. This was created while specifically working on the criteo outbrains and rossman datasets. For those targets, the delimiter there was just in place to separate the operators from the columns when you do not replace are target column in an operator. This trickles down. I agree that this should be changed. But maybe not in this PR. Because currently that is how non-replaced columns are created. Essentially columns are named based on where in the chain of operations they are created,  appending operators chained to represent a given columns data. i.e. col_op1_op2_op3_... So if we change this we need to change to handled that as well. Otherwise we will break operator(replace=False) functionality.

rdipietro: If it breaks other functionality I definitely agree – let's leave it out of this PR.

However can you please elaborate on what would break? I'm pretty new to NVTabular so I don't see why a typical sort would break anything (using the entire string as the key).

E.g. imagine columns `['user_feature_1', 'user_feature_2', 'user_feature_3']` and applying an operator with `replace=True` to form `'user_feature_2'`. Then we'd end up with a sorted list `['user_feature_1', 'user_feature_2_op1', 'user_feature_3']`. What would break in this case?

nvtabular/ops.py
jperez999: This was built this way in case you changed the column ordering of the preprocessed dataset. We give a dictionary with the encodings and then a list dictating the order of the embeddings. If we agree this level of malleability is not warranted then I am good with only reflecting the embeddings in the order of preprocessing.

nvtabular/ops.py
jperez999: No thats not what would happen. the moment you set `replace=True` the output will be `['user_feature_1', 'user_feature_2', 'user_feature_3']` esesntially just the original columns... if you did `replace=False` after operator you would have `['user_feature_1', 'user_feature_2', 'user_feature_3', 'user_feature_2_op1']`. But to your question above, the reason for the sort is to guarantee an order. When I first started working with nvtabular, I was hitting an edge case (i dont remember specifically, I think it was with outbrains) where the loading of embeddings to the model and the tensors did not match up because of a difference in keyword ordering... So the sort, in every place you see it is to ensure that given a list of values, I can guarantee the order of that list. If we want to move to a different method for guaranteeing, in all cases, this functionality, I dont have a problem with that.

rdipietro: Ah yes sorry, I did mean `replace=False` there (not `replace=True`)

I see your point, and I agree that a sort can be helpful to guarantee a certain global order, but my question is instead about the key used to sort: `key=lambda entry: entry.split("_")[0]`

The question I'm asking is: what would break if we were to instead a *typical* key (`entry`, e.g. `'user_feature_2'`) rather than this modified key (`entry.split("_")[0]`, e.g. `'user'` instead of `'user_feature_2'`) ?

(Also, just to note, this may seem harmless, but this led to confusion + bugs for me while exploring the Rossmann notebook)

nvtabular/ops.py
rdipietro: Oh okay, so the motivation is to avoid mistakes from users who are unaware of the internal reordering. This makes sense 👍

My only comment then is that I think the interface should be clearer / cleaner. Would there be any drawback of adding this functionality within a `Workflow`? To me it makes sense since we are coupling *getting embedding sizes* with *the inherent ordering produced by a workflow*.

I see that right now the `Workflow` has the method `get_final_cols_names`. I could imagine two additional methods, `get_cat_cardinalities` and `get_cat_embedding_sizes`, which wouldn't even need to take any arguments and so would be extremely simple to use?

I'm imagining replacing code like this

```
sizes = get_embedding_size(proc.stats["categories"], proc.columns_ctx["categorical"]["base"])
```

with code like this

```
sizes = proc. get_cat_embedding_sizes()
```

alecgunny: At it's core, isn't this function just a mapping from an int cardinality to an int embedding dim? It seems like instead of forcing iteration inside the function, the function should just do that mapping on individual ints and let the user iterate on their calling of it. If we're concerned about the column reordering, it seems more sensible to expose the right ordering via some property and let the user use that to iterate, rather than continuing to bury confusing behavior beneath APIs (eventually someone will have a use case that requires extension, and not knowing about that reordering will break things).

This is probably best for a separate issue, but I do think we'll eventually want to drop the idea of ordering altogether and feed models via dicts like in the KerasSequenceDataset. Besides having the advantage of being cleaner and avoiding these sorts of problems, it will basically be mandatory once we start supporting multi-hot categorical features which can't be cleanly fit into rectangular matrices. This might cost us some performance on the embedding (since we're not doing a single slice on a 2D matrix), but the solution to this should be on the framework side to build an embedding op which can efficiently handle these dict inputs.

rdipietro: I thought @jperez999 's point was that this function should safeguard against using a user-defined ordering, because it is inevitable that users will a) not know of the underlying reordering and b) not know that an exposed 'correct' ordering exists, which in turn will lead to strange and hard-to-debug behavior.

That said, I can see either option being reasonable, and either is better than the current approach.

As an avid PyTorch user, I disagree with the *feed models with dicts* sentiment: it feels like a huge step backwards. Why is it mandatory for multi hot? I have not thought about that problem at length but I feel like other options should exist. For example why not use sparse binary tensors?

alecgunny: I guess my point is that we can't safeguard against this at every turn, and to a certain extent how and the order in which the user wants to retrieve their embedding sizes is sort of their business (and more exotic models will require more exotic retrieval methods). I would rather make this a known and mandatory part of the API that users need to navigate explicitly, and simplify function calls accordingly, rather than try to anticipate every use case where ordering might be important and head it off with a custom function.

Let's take the dict discussion somewhere else, it probably doesn't make sense here. But I think there are lots of good reasons to prefer it.

nvtabular/ops.py
benfred: I think @rdipietro has a good point here about subsorting - I think it should be fine to use ```entry``` as a sort key instead of ```entry.split("_")[0]``` in all cases (because we still be sorting by the first token up to '_', and just using the rest of the key to subsort after that). @jperez999 do you have any objections to this?

nvtabular/ops.py
benfred: There are a lot of good points here - but I also think we should keep the conversation going elsewhere. I'm going to merge this PR as is since I think we all agree its a small step in the right direction, even if maybe doesn't go quite far enough. Also the example notebooks are all broken in master, and this change fixes so I think we should get it =).

nvtabular/ops.py
benfred: I've made that change in the most recent commit

benfred:
rerun tests

PR #112, opened 2020-06-25
[REVIEW]  fixing torch dataloader name typo in the criteo_example.ipynb

rnyak:
The statement below in the criteo_example notebook, under the section `Data Loading` reads as follow:

```
We'll start by using the parquet files we just created to feed an NVTabular FileItrDataset, which will looop through the files in chunks. First, we'll reinitialize our memory pool from earlier to free up some memory so that we can share it with PyTorch.
```

However, we do no longer use `FileItrDataset` here, we use `TorchTensorBatchDatasetItr` instead. Thus, we should replace `FileItrDataset` with `TorchTensorBatchDatasetItr`  within the sentence above.

Similarly, if we are replacing `FileItrDataset` with `TorchTensorBatchDatasetItr` this should be changed in `criteo_example.py` and `gpu_benchmark` accordingly to be consistent.


benfred:
run tests

jperez999:
rerun tests

benfred:
run tests

rnyak:
@jperez999 @benfred In addition, we do shuffling for validation set also. Is that the intention? we set it to False in Rossmann notebook.

`proc.apply(valid_ds_iterator, apply_offline=True, record_stats=False, shuffle=True, output_path=output_valid_dir, num_out_files=35)`

jperez999:
rerun tests

PR #114, opened 2020-06-26
[REVIEW] Use dask workflow in unittests

benfred:
Use the dask Workflow for all unittests - after this change the single gpu BaseWorkflow is not created by the unittests

### DIFF
--- a/nvtabular/workflow.py
+++ b/nvtabular/workflow.py
@@ -31,7 +31,7 @@ import nvtabular.dask.io as dask_io
 from nvtabular.ds_writer import DatasetWriter
 from nvtabular.encoder import DLLabelEncoder
 from nvtabular.io import HugeCTRWriter, ParquetWriter, Shuffler
-from nvtabular.ops import DFOperator, Export, StatOperator, TransformOperator
+from nvtabular.ops import DFOperator, StatOperator, TransformOperator
 
 LOG = logging.getLogger("nvtabular")
 
@@ -61,7 +61,6 @@ class BaseWorkflow:
     label_name : list of str
         Names of the label column.
     config : object
-    export : bool, default False
     export_path : str, default "./ds_export"
     """
 
@@ -71,8 +70,6 @@ class BaseWorkflow:
         cont_names=None,
         label_name=None,
         config=None,
-        # TODO: move these out
-        export=False,
         export_path="./ds_export",
     ):
         self.phases = []
@@ -89,7 +86,6 @@ class BaseWorkflow:
 
         self.stats = {}
         self.ds_exports = export_path
-        self.export = export
         self.current_file_num = 0
         self.timings = {
             "shuffle_df": 0.0,
@@ -299,9 +295,6 @@ class BaseWorkflow:
         baseline, leftovers = self._sort_task_types(master_task_list)
         self.phases.append(baseline)
         self._phase_creator(leftovers)
-        # check if export wanted
-        if self.export:
-            self._phases_export()
         self._create_final_col_refs(task_sets)
 
     def _phase_creator(self, task_list):
@@ -332,22 +325,6 @@ class BaseWorkflow:
 
             if not added:
                 self.phases.append([task])
-
-    def _phases_export(self):
-        """
-        Export each phase from the dependency dictionary, that creates transformations.
-        """
-        for idx, phase in enumerate(self.phases[:-1]):
-            trans_op = False
-            # only export if the phase has a transform operator on the dataset
-            # otherwise all stats will be saved in the tabular object
-            for task in phase:
-                if isinstance(task[0], TransformOperator):
-                    trans_op = True
-                    break
-            if trans_op:
-                tar_path = os.path.join(self.ds_exports, str(idx))
-                phase.append([Export(path=f"{tar_path}"), None, [], []])
 
     def _find_parents(self, ops_list, phase_idx):
         """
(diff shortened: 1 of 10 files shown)


jperez999:
rerun tests

nvtabular/workflow.py
```diff
@@ -89,7 +86,6 @@ def __init__(
 
         self.stats = {}
         self.ds_exports = export_path
```
rjzamora, line 88: Doesn't seem like the workflow is using `self.ds_exports` either?

tests/unit/test_ops.py
```diff
@@ -106,27 +106,31 @@ def test_moments(client, tmpdir, df, dataset, gpu_memory_frac, engine, op_column
...
-    config["PP"]["categorical"] = [ops.Encoder(columns=op_columns)]
+    config["PP"]["categorical"] = [encoder]
 
     processor = nvt.Workflow(
-        cat_names=cat_names, cont_names=cont_names, label_name=label_name, config=config
+        cat_names=cat_names,
+        cont_names=cont_names,
+        label_name=label_name,
+        config=config,
+        client=client,
```
rjzamora, line 123: So, now that we are passing `client` here, is the single-GPU version of `Encoder` now untested?  This is certainly fine (since we are moving to the all-dask statistics backend), but perhaps we will need to remove much of the single-gpu code to pass the "coverage criteria" for CI to pass?

tests/conftest.py
```diff
@@ -179,3 +179,13 @@ def remove_sub_files_folders(path):
         for root, dirs, files in os.walk(path):
             for file in files:
                 os.remove(os.path.join(root, file))
+
+
+def get_cats(processor, col):
```
rjzamora, line 184: Nice cleanup.

Are any tests actually useing the `else` block here anymore?

tests/unit/test_ops.py
benfred: We should definitely remove soon, but it's still used by the example notebooks. 

For code coverage, I've changed the 'test_gpu_workflow_api' test to test both the dask and single gpu version: https://github.com/NVIDIA/NVTabular/pull/114/files#diff-78067b78469064255f1184dc1acadb7cR37 .

nvtabular/workflow.py
benfred: Good catch - it was only being used in the unittests. removed in the latest commit

tests/conftest.py
benfred: I've changed a couple of stat ops to test both the single/multi gpu versions temporarily (to get code coverage requirements higher) - but will remove this soon

benfred:
rerun tests

tests/unit/test_tf_dataloader.py
```diff
@@ -28,8 +26,7 @@
 @pytest.mark.parametrize("gpu_memory_frac", [0.01, 0.1])
 @pytest.mark.parametrize("engine", ["parquet"])
 @pytest.mark.parametrize("batch_size", [1, 10, 100])
-def test_tf_gpu_dl(tmpdir, datasets, batch_size, gpu_memory_frac, engine):
-    paths = glob.glob(str(datasets[engine]) + "/*." + engine.split("-")[0])
+def test_tf_gpu_dl(tmpdir, client, paths, dataset, batch_size, gpu_memory_frac, engine):
```
rjzamora, line 29: I just installed tensorflow and this test fails for me.  Your suspicion (mentioned offline) that CI is skipping this is likely true.  `KerasSequenceDataset` is expecting the user to pass in raw paths so it can create its own `GPUDatasetIterator`.   Therefore, it is fine to use the dask-based workflow to update the statistics, but we will need to update the dataloader to also accept the same `dataset` object as the workflow (and use `to_iter` to iterate through the dataset)

benfred: I agree that we're going to need to make those updates you mentioned - but the unittests *should* work with these changes. I've got TF installed on my system, and it works for me:

```
tests/unit/test_tf_dataloader.py ......                                                                          [100%]

=================================================== warnings summary ===================================================
tests/unit/test_tf_dataloader.py: 19 tests with warnings
  /home/ben/code/nvtabular/nvtabular/io.py:552: UserWarning: get_memory_info is not supported. Using total device memory from NVML.
    warnings.warn("get_memory_info is not supported. Using total device memory from NVML.")

-- Docs: https://docs.pytest.org/en/latest/warnings.html
=========================================== 6 passed, 19 warnings in 12.20s ============================================
```

Were you running the TF tests on this branch? What was the error you saw?

benfred:
rerun tests

tests/unit/test_tf_dataloader.py
rjzamora: Ah - my mistake.  You are still passing in `paths` here - makes sense.

rjzamora:
@benfred - Let me know how you want to align the overlapping/conflicting changes between this PR and #120 ...

Since CI is passing here, it seems fine to merge this first.  However, that the other PR makes it unnecessary to pass the `cleint` to all tests (testing is faster with `client=None`).

It is also fine with me if you want to merge 120 into this PR - However, that will make `BaseWorkflow` less accessible (and may even break un-tested example notebooks)

benfred:
@rjzamora Lets get this merged - we can always scale back the number of tests run with a client once #120 is in too. Are there any other changes you'd like with this PR ? (it sitll hasn't been approved yet =).

Issue #56, opened 2020-06-04
[FEA] Support "online" apply with dask


benfred:
<a href="https://github.com/benfred"><img src="https://avatars2.githubusercontent.com/u/69536?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [benfred](https://github.com/benfred)**
_Tuesday May 26, 2020 at 23:59 GMT_
_Originally opened as https://github.com/rapidsai/recsys/issues/175_

----

We should support the 'online' apply with the dask multigpu version, so that we can apply statistics computed by dask to the dataloaders. 

With the current version, this might have an issue with using the stats computed by the dask Encoder in the Categorify op.

benfred:
Done with https://github.com/NVIDIA/NVTabular/pull/120

PR #120, opened 2020-06-30
[REVIEW] Handle client=None in Dask-based Workflow

### LINKED ISSUE #119: [FEA] Dask Workflow with no client
We should be able to run the dask workflow with no client passed in.  This should help out with the code coverage issues we've been seeing, and mean that are existing examples will work as is with Dask.

rjzamora:
Closes #119 

This PR adds support for `client=None` in the Dask-based `Workflow`.  Some important notes:

- Much of this code is copied from #114 (So its fine if you want to merge these changes into that PR @benfred )
- **Need to discuss this**: Since we handle `client=None` in the Dask-based workflow, I renamed `DaskWorkflow` to `Workflow`.  This does not break any unit tests, but may cause problems in examples/notebooks.  The user will no longer get a `BaseWorkflow` object when they don't pass a client to `Workflow()` (they need to use `BaseWorkflow()` explicitly).


nvtabular/tf_dataloader.py
```diff
@@ -124,12 +125,13 @@ class KerasSequenceDataset(tf.keras.utils.Sequence):
 
   Parameters
   -------------
-  - file_pattern: str or list(str)
+  - paths_or_dataset: str or list(str)
```
rjzamora, line 128: Note that the changes in this file are not necessary for this PR, but I left them in (for now) - If @alecgunny will be changing everything here soon anyway, I will just reset this file.

rjzamora:
Note that it likely makes sense to add explicit support for `apply_offline=False` here (and to make sure shuffling works with `client=None`.  However, I am curious why there is no testing at all for `HugeCTRWriter` -  I was also thinking of adding `hugectr` output support within the dask-based workflow, but its hard to validate my changes without any "working" tests?

rjzamora:
Just started removing a bunch of the single-gpu code that we are no longer planning to support (relating to stats collection without Dask).  These changes should allow us to close #56, since the user can no longer try to collect global statistics when `apply_offline=False`.  If I understand correctly, it doesn't really make sense for the statistics to be collected "online" anyway.  The exception would be if we choose to add a new `StatOperator` method that defines the necessary logic to *update* an existing global statistic using a new sample. If we do decide this functionality is necessary, it can certainly be added later.

I think the next step here is to utilize the `ParquetWriter`, `Shuffler` and `HugeCTRWriter` within the offline Dask-based workflow (each worker can just use the existing functionality on a distinct set of files, with the client writing the metadata summary).  After the writer/shuffler API/code is consistent (and the examples are fixed/validated) we should be in good shape here.

rjzamora:
rerun tests

rjzamora:
@benfred @jperez999 - I'm sorry for letting this PR balloon so far beyond the initial goal.  As I enabled the dask-based workflow to work without a distributed client, it became clear that it may be difficult to move away from the single-gpu version in "bite-size" chunks.  That is - we really need everyone to stop using/learning with the single-gpu-only workflow

I certainly understand if there is interest in rolling back some of these changes.  However, I personally feel that we should pull off the single-gpu-only workflow like a band aid :) -- The current changes may break any existing code that assumes  "offline" preprocessing can be accomplished with only a `GPUDatasetIterator`,  or code that wants to generate HugeCRT output during "offline" preprocessing.  With that said, we should be able to divide and conquer remaining issues pretty easily.

nvtabular/workflow.py
```diff
@@ -853,37 +681,113 @@ def exec_phase(self, phase_index, record_stats=True):
...
+            if hugectr_gen_output:
+                raise ValueError(
+                    "TODO: Support HugeCTR output for offline processing with Dask."
+                    " This is part of the larger task of aligning online/offline API."
+                )
             self.update_stats(
                 dataset,
                 output_path=output_path,
-                record_stats=record_stats,
+                record_stats=record_stats or True,
```
rjzamora, line 757: This should probably be changed - not sure what I was trying to do here.

nvtabular/workflow.py
```diff
@@ -853,37 +681,113 @@ def exec_phase(self, phase_index, record_stats=True):
...
+                    self._update_stats(op)
+                    op.clear()
             del stats
 
     def apply(
         self,
         dataset,
         apply_offline=True,
-        record_stats=True,
+        record_stats=None,
```
benfred, line 702: The docstring says that record_stats is a boolean (and it seems to be used that way). Should the default value by False ?
```suggestion
        record_stats=False,
```

examples/criteo_benchmark.py
```diff
@@ -92,36 +92,37 @@ def parse_args():
...
-    engine=args.in_file_type,
-    gpu_memory_frac=float(args.gpu_mem_frac),
-    sep="\t",
+dataset_args = {"sep": "\t"} if args.in_file_type == "csv" else {}
+trains_ds = Dataset(
+    train_set, engine=args.in_file_type, part_mem_fraction=float(args.gpu_mem_frac), **dataset_args
 )
+valids_ds = Dataset(
+    valid_set, engine=args.in_file_type, part_mem_fraction=float(args.gpu_mem_frac), **dataset_args
+ )
```
benfred, line 117:
```suggestion
)
```

nvtabular/workflow.py
rjzamora: We probably want the default to be True, but we then we cannot raise a warning when `record_stats=True` is used with `apply_offline=True`. (we just won't calculate statistics online ever...) -  My thought with using `None` was that the default can depend on the value of `apply_offline`

PR #124, opened 2020-07-01
[REVIEW] BugFix FillMissing

bschifferer:
The FillMissing operator did not use the parameter fill_val to fill na values.
Instead it used a hardcoded value 0.

The dataset in the unittest did not have any na values. Therefore, it passed.


benfred:
thanks!

Issue #71, opened 2020-06-05
[FEA] TensorFlow data loader performance improvements


benfred:

alecgunny:
Progress being made here https://github.com/alecgunny/NVTabular/tree/tfexample

PR #125, opened 2020-07-01
[REVIEW] Adding TensorFlow example

alecgunny:
Building end-to-end TensorFlow Criteo example to demonstrate how to use NVTabular as a data loader and online preprocessor for TensorFlow. End version will probably end up restructuring existing Criteo example to make its preprocessing notebook callable


alecgunny:
Addresses #71

examples/tensorflow/accelerating-tensorflow.ipynb
```diff
@@ -0,0 +1,1647 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# import multiprocessing as mp\n",
```
jperez999, line 942: should we keep this code block in here if its completely commented out?

alecgunny:
@benfred I did since it was out of date and since the recipe I had to cook up to avoid using NVIDIA-internal bases was at least somewhat overly specific to the example, so I didn't feel like replacing it in the original repo made a lot of sense. I'm all for exposing a base "vanilla" NVTabular `Dockerfile` at the root level that corresponds to some public nightly build or maintains up-to-date recipes for people to extend to something like `examples/tensorflow/docker/Dockerfile`, but until we decide on what such a container should look like and how we want people to extend it for the "framework-specific" recipes, I felt like that recipe was getting counter productive (i.e. might be confusing for someone trying to do the example and noticing that the two Dockerfiles looked nothing like each other).

Agreed on the layers and the github.io. For the layers, I think a discussion about where to keep assets that are not strictly "nvtabular" is worth having first, and I didn't want to do any core repo rearranging.

PR #127, opened 2020-07-02
[REVIEW] Async torch Dataloaders

jperez999:


jperez999:
rerun tests

nvtabular/torch_dataloader.py
```diff
@@ -125,7 +125,7 @@ def create_tensors_plain(gdf, cat_cols=None, cont_cols=None, label_cols=None):
 
 
 def combine_tensors(cats, conts, label):
-    cats_list = [cats[x] for x in _get_embedding_order(cats.keys())] if cats else None
+    cats_list = [cats[x] for x in sorted(cats.keys())] if cats else None
```
rdipietro, line 128: There's one other instance of `_get_embedding_order` in this file. For clarify / consistency could consider making this change there, too

nvtabular/torch_dataloader.py
```diff
@@ -139,6 +139,10 @@ def combine_tensors(cats, conts, label):
 def _one_df(
     gdf, cats, conts, label, cat_names=None, cont_names=None, label_names=None,
 ):
+    if gdf.empty:
```
rdipietro, line 142: Remove left over debugging code?

nvtabular/torch_dataloader.py
```diff
@@ -218,20 +222,75 @@ def __len__(self):
...
-    def load_chunk(self, out):
         for chunk in self.itr:
-            chunk = create_tensors_plain(
-                chunk, cat_cols=self.cat_cols, cont_cols=self.cont_cols, label_cols=self.label_cols
-            )
-            out.put(chunk)
+            yield chunk
+
+
+class ChunkQueue:
```
rdipietro, line 229: Documentation here would help a lot, especially to describe what the difference between `num_chunks` and `batch_size` is.

nvtabular/torch_dataloader.py
```diff
@@ -218,20 +222,75 @@ def __len__(self):
...
+
+    def get(self):
+        return self.q_out.get()
+
+    def put(self, obj):
+        # check first for "end"
+        if isinstance(obj, str):
+            # clear the buffer
+            self.create_chunk(final=True)
+            # send bug out
```
rdipietro, line 256: What's meant by "send bug out" ?

nvtabular/torch_dataloader.py
```diff
@@ -218,20 +222,75 @@ def __len__(self):
...
+                return
+        if self.shuffle:
+            _shuffle_gdf(chunks)
+        chunks = create_tensors_plain(
+            chunks, cat_cols=self.cat_cols, cont_cols=self.cont_cols, label_cols=self.label_cols
+        )
+        # chunk tensorized
+        self.q_out.put(chunks)
+
+    def stitch_run(self, chunks):
```
rdipietro, line 284: I'm lost connecting `stitch_run` with what this function is doing (splitting a list of chunks up into one large chunk that's divisible by batch size and putting the "spill" back into the queue).

A better name might be something like `get_batch_size_divisible_chunk`?

nvtabular/torch_dataloader.py
```diff
@@ -249,7 +308,39 @@ def gdf_col(self, gdf):
         return (batch[0], batch[1]), batch[2].long()
 
 
-class TorchTensorBatchDatasetItr(torch.utils.data.ChainDataset):
+class AsyncTensorBatchDatasetItr(torch.utils.data.IterableDataset):
+    def __init__(self, paths, **kwargs):
```
rdipietro, line 312: Documentation would help a lot here, both to describe parameters + keyword args and describe usage

Also – should the keyword arg `sub_batch_size` be renamed to `batch_size`?

nvtabular/torch_dataloader.py
```diff
@@ -272,9 +363,11 @@ def __init__(self, paths, **kwargs):
         self.cur_path = None
         self.kwargs = kwargs
         self.rows = 0
+        batch_size = kwargs.get("sub_batch_size", 1)
```
rdipietro, line 366: Same comment here re: `sub_batch_size`. Rename to `batch_size`?

nvtabular/torch_dataloader.py
```diff
@@ -272,9 +363,11 @@ def __init__(self, paths, **kwargs):
...
         for file_path in self.paths:
             (num_rows, num_row_groups, columns,) = cudf.io.read_parquet_metadata(file_path)
-            self.rows += (num_rows // kwargs.get("sub_batch_size", 1)) + 1
+            chunks = int(num_rows / batch_size)
+            self.rows += chunks + 1 if num_rows % batch_size > 0 else chunks
```
rdipietro, line 370: The chance of this actually causing problems is small, but it's worth mentioning that the (int to float to int) could possibly cause problems. If we're okay with this, it'd probably be worth replacing with `chunks = math.ceil(num_rows / batch_size)`. But I'd instead suggest something like

```
chunks, spill = divmod(num_rows, batch_size)
if spill > 0: chunks += 1
self.rows += chunks
```

Also, I'm finding the names here confusing. Is it worth clarifying a bit? E.g. does a "row" correspond to a batch? And does the word "chunk" here means something completely different than the "chunk" in the `ChunkQueue`?

nvtabular/torch_dataloader.py
jperez999: Sorry bug out is a term I used in the military, that mean exit or leave.

nvtabular/torch_dataloader.py
jperez999: Its been changed to batch_size

nvtabular/torch_dataloader.py
jperez999: mean exit or leave... sorry this is my old military life spilling over.

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
This PR introduces the AsyncTensorBatchDataloader and replaces all other dataloader models (i.e. FileIterator, Chaindataset) the other dataloader classes have been removed to consolidate on this class. The updates have been made throughout the repo

tests/unit/test_notebooks.py
```diff
@@ -72,10 +72,10 @@ def _run_notebook(tmpdir, notebook_path, transform=None):
 def _get_random_criteo_data(rows):
     dtypes = {col: float for col in [f"I{x}" for x in range(1, 14)]}
     dtypes.update({col: int for col in [f"C{x}" for x in range(1, 27)]})
-    dtypes["label"] = int
+    dtypes["label"] = bool
```
benfred, line 75: Thanks for fixing this!

nvtabular/io.py
```diff
@@ -729,6 +729,9 @@ def __init__(self, paths, part_size, storage_options=None):
     def to_ddf(self, columns=None):
         raise NotImplementedError(""" Return a dask_cudf.DataFrame """)
 
+    def make_iter(self, columns=None, indices=None):
```
rjzamora, line 732: Unless I am missing something, you shouldn't need the `make_iter` definitions anymore.  This logic was recently moved up to the Dataset level.

nvtabular/io.py
```diff
@@ -918,21 +921,31 @@ def to_ddf(self, columns=None):
             return self._ddf[[columns]]
         return self._ddf
 
+    def make_iter(self, columns=None, indices=None):
```
rjzamora, line 924: Same here.

nvtabular/io.py
```diff
@@ -918,21 +921,31 @@ def to_ddf(self, columns=None):
...
     @property
     def num_rows(self):
         return len(self._ddf)
 
 
 class DataFrameIter:
-    def __init__(self, ddf, columns=None):
+    def __init__(self, ddf, columns=None, indices=None):
+        self.indices = (
+            indices if isinstance(indices, list) and len(indices) > 0 else range(ddf.npartitions)
```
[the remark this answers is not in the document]
jperez999: cant do this... bool not available for list type, it causes error saying you need to check any or all. So it make work with the range but if we give an actual list of values it will fail. Much cleaner this way... Originally wanted it like that.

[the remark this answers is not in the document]
jperez999: specifically something like this bubble:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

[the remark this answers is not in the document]
benfred: good point, I hadn't considered c- maybe should just use if not None instead ?

```suggestion
        self.indices = indices if indices is not None else range(ddf.partitions)
```

nvtabular/torch_dataloader.py
jperez999: changed

examples/criteo-example.ipynb
```diff
@@ -150,7 +150,12 @@
     "proc.add_cont_preprocess(Normalize())\n",
     "\n",
     "# categorification with frequency thresholding\n",
-    "proc.add_cat_preprocess(Categorify(freq_threshold=15))\n"
+    "proc.add_cat_preprocess(Categorify(\n",
+    "                            freq_threshold=15,\n",
+    "                            tree_width=4,\n",
+    "                            out_path=OUTPUT_DATA_DIR,\n",
+    "                            )\n",
```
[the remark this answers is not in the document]
jperez999: removed however we take a 10% time hit... (i.e. preprocessing training takes 20 minutes compared to 18 with tree_width=4)

nvtabular/torch_dataloader.py
```diff
@@ -240,7 +237,41 @@ def gdf_col(self, gdf):
...
+            label_cols=self.labels,
+        )
+        threading.Thread(target=self.load_chunk, args=(buff,)).start()
+        while True:
+            # self.load_chunk(buff)
+            chunk = buff.get()
+            if isinstance(chunk, str):
+                return
+            yield from TensorItr(chunk, batch_size=self.batch_size)
+            chunk = None
```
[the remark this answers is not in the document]
jperez999: introduced subclass iterator to handle this

nvtabular/io.py
jperez999: removed

nvtabular/torch_dataloader.py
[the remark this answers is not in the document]
jperez999: and created a cusom stop method in the chunk queue class to handle termination of ongoing loop and dump all memory usage.

examples/criteo-example.ipynb
```diff
@@ -166,8 +171,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "train_dataset = nvt.Dataset(train_paths, engine='parquet', part_mem_fraction=0.12)\n",
-    "valid_dataset = nvt.Dataset(valid_paths, engine='parquet', part_mem_fraction=0.12)"
+    "train_dataset = nvt.Dataset(train_paths, engine='parquet', part_mem_fraction=0.15)\n",
+    "valid_dataset = nvt.Dataset(valid_paths, engine='parquet', part_mem_fraction=0.15)"
```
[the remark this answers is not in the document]
jperez999: This changes performance quite a bit... we can make the change but it will result in like a 23 minute run... All the numbers are calibrated to give us the fastest possible run. Any deviations whether in part_mem_fraction or tree width will result in slower run time.

nvtabular/torch_dataloader.py
jperez999: removed

nvtabular/torch_dataloader.py
jperez999: fixed

nvtabular/io.py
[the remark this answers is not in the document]
jperez999: This method causes the following error:
`'IndexCallable' object cannot be interpreted as an integer`

examples/criteo-example.ipynb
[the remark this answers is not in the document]
benfred: Can we change the defaults inside the library to 0.15 part_mem_fraction and a 4 tree_width? Is there any reason not to do that?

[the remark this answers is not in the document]
jperez999: the part_mem_fraction is above the limit @rjzamora set out for partitions regularly. This is customized specifically for this data... Other data may perform better  with the current defaults...

PR #128, opened 2020-07-02
[WIP] Dask embedding api fix

jperez999:
making api a little cleaner.


examples/criteo_benchmark.py
```diff
@@ -146,7 +146,7 @@ def parse_args():
 # TODO: Implement the get_embedding_size for dask-based workflow
 embeddings = [
     x[1]
-    for x in get_embedding_size(proc.stats["categories"], proc.columns_ctx["categorical"]["base"])
+    for x in get_embeddings(proc)
```
benfred, line 149: I think this is called in a couple of other spots (gpu_benchmark.ipynb, criteo-example.ipynb, ossmann-store-sales-example.ipynb). Can you update those too?

nvtabular/ops.py
```diff
@@ -1226,5 +1226,19 @@ def get_embedding_size(encoders, cat_names):
...
+    return get_embeddings_dask(workflow.stats["categories"], cols) 
+
+
+def get_embeddings_dask(paths, cat_names):
+    embeddings = {}
+    for col in sorted(cat_names):
+        path = encoders[col]
+        num_rows, _, _ = cudf.io.read_parquet_metadata(path)
+        embeddings[col] = (num_rows, _emb_sz_rule(num_rows))
+    return embeddings
```
benfred, line 1240: I don't think you need to sort in get_embeddings_dask - we're already sorting in the get_embedding_order call from get_embeddings.

Also since we're not exposing get_embeddings_dask as part of our public API, lets either merge these functions (since they are short) or prefix with a _ . Also I don't think the 'encoders' value on line 1237 is defined ?

```suggestion
def get_embeddings(workflow):
    cat_names = get_embedding_order(workflow.columns_ctx["categorical"]["base"])
    encoders = workflow.stats["categories"]
    embeddings = {}
    for col in cat_names:
        path = encoders[col]
        num_rows, _, _ = cudf.io.read_parquet_metadata(path)
        embeddings[col] = (num_rows, _emb_sz_rule(num_rows))
    return embeddings
```

examples/criteo-example.ipynb
```diff
@@ -304,7 +304,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "embeddings = [x[1] for x in get_embedding_size(proc.stats[\"categories\"], proc.columns_ctx['categorical']['base'])]"
+    "embeddings = [v for k, v in get_embeddings(proc).items()]"
```
benfred, line 307: how about 
```suggestion
    "embeddings = list(get_embeddings(proc).values())"
```

examples/criteo_benchmark.py
```diff
@@ -75,12 +75,12 @@ def parse_args():
     data_path + df_train + x
     for x in os.listdir(data_path + df_train)
     if x.startswith("day") and x.split(".")[0].split("_")[-1] in train_days
-]
+][:1]
```
benfred, line 78:
```suggestion
]
```

examples/criteo_benchmark.py
```diff
@@ -75,12 +75,12 @@ def parse_args():
...
 valid_set = [
     data_path + df_valid + x
     for x in os.listdir(data_path + df_valid)
     if x.startswith("day") and x.split(".")[0].split("_")[-1] in valid_days
-]
+][:1]
```
benfred, line 83:
```suggestion
]
```

examples/criteo_benchmark.py
```diff
@@ -138,16 +138,17 @@ def parse_args():
...
+
 # TODO: Implement the get_embedding_size for dask-based workflow
-embeddings = [
-    x[1]
-    for x in get_embedding_size(proc.stats["categories"], proc.columns_ctx["categorical"]["base"])
-]
+embeddings = [v for k, v in get_embeddings(proc).items()]
+import pdb
+
+pdb.set_trace()
```
benfred, line 151:
```suggestion
embeddings = list(get_embeddings(proc).values())
```

nvtabular/ops.py
rdipietro: I like the idea of replacement here, but I'd like to suggest either leaving the old name as it is (`get_embedding_size`) or just making it plural (`get_embedding_sizes`).

The proposed name (`get_embeddings`) is confusing since we are using this function to get embedding sizes, not to get embeddings themselves.

PR #131, opened 2020-07-04
[REVIEW] S3 Support

benfred:
This adds basic capabilities to read datasets from S3 (and probably GCS/HDFS) with nvtabular.   The big change here is to add a basic unittest that tests out this capability, as well as fixing the couple of spots that broke.

As part of this I refactored out the GPUDatasetIterator class in favour of the unified Dataset class.


benfred:
rerun tests

PR #143, opened 2020-07-10
[REVIEW] Add unittest for our notebooks

benfred:
Add a unittest runner for our notebooks, checks that the rossmann example can
be run without error on a small dataset (10k rows) training for only a single
epoch. This should let us verify that everything runs, but only adds about 10s
to the unittest execution time.


tests/unit/test_notebooks.py
```diff
@@ -0,0 +1,61 @@
...
+    os.environ["INPUT_PATH"] = str(tmpdir)
+    os.environ["OUTPUT_PATH"] = str(tmpdir)
+    source = _read_notebook_source(notebook_path)
+    exec(source, {})  # pylint: disable=exec-used
+
+
+def test_rossman_notebook(tmpdir):
+    # datasets are stored zipped to avoid blowing up the repo, handle
+    dataset_path = os.path.join(TEST_PATH, "datasets", "rossmann")
+    _unzip_dataset(dataset_path, tmpdir)
```
jperez999, line 27: why store any data in the repo. Why cant we just store on machine and mount the volume on the CI container?

benfred: I've changed to generate random data rather than store in the repo

benfred:
There seems to be an issue with the criteo notebook segfaulting on RMM freeing buffers here so I've disabled that test (xskip) while I debug, but aside from that this should be good to go

PR #151, opened 2020-07-14
[REVIEW] Fix output-file count bug

### LINKED ISSUE #150: [BUG] Output-file count is incorrect for shuffle="partial"
**Describe the bug**
As originally discovered by @rnyak, there seems to be an extra output file written for `shuffle="partial"`.   That is, when using a single process with `out_files_per_proc=2`, there are **3** output files written.

```python
import nvtabular as nvt
import pandas as pd
import numpy as np

!rm -rf test.parquet processed
!mkdir processed
path = "test.parquet"
out_path = "processed/"

out_files_per_proc = 2

size = 25
df = pd.DataFrame({"a": np.arange(size)})
df.to_parquet(path, row_group_size=5, engine="pyarrow")

columns = ["a"]
dataset = nvt.Dataset(
    path, engine="parquet", row_groups_per_part=1
)
processor = nvt.Workflow(
    cat_names=[], cont_names=columns, label_name=[]
)

processor.add_preprocess(nvt.ops.Normalize())
processor.finalize()

processor.update_stats(
    dataset,
    output_path=out_path,
    shuffle="partial", 
    out_files_per_proc=out_files_per_proc,
)
!ls processed
```
**Output has 3 files (should be 2)**:
```
split.0.459d88bc3f52405bae6fd572570fd452.parquet
split.1.ff7fd6b1f7d44504b24351afd8b43d4c.parquet
split.2.f6cfed77c5a848e998c60b2d2613a4dd.parquet
```

**Expected behavior**
The `out_files_per_proc` parameter should result in the expected number of output files.

rjzamora:
Closes #150 

This PR simplifies the `_write_output_partition` code and fixes the bug described in #150

### DIFF
--- a/nvtabular/categorify.py
+++ b/nvtabular/categorify.py
@@ -191,7 +191,9 @@ def _write_uniques(dfs, base_path, col, on_host):
         # Make sure first category is Null
         df = df.sort_values(col, na_position="first")
         if not df[col]._column.has_nulls:
-            df = cudf.DataFrame({col: _concat([cudf.Series([None]), df[col]], ignore_index)})
+            df = cudf.DataFrame(
+                {col: _concat([cudf.Series([None], dtype=df[col].dtype), df[col]], ignore_index)}
+            )
         df.to_parquet(path, write_index=False, compression=None)
     else:
         df_null = cudf.DataFrame({col: [None]})
--- a/nvtabular/io.py
+++ b/nvtabular/io.py
@@ -38,7 +38,6 @@ from cudf.io.parquet import ParquetWriter as pwriter
 from dask.base import tokenize
 from dask.dataframe.core import new_dd_object
 from dask.dataframe.io.parquet.utils import _analyze_paths
-from dask.dataframe.utils import group_split_dispatch
 from dask.distributed import get_worker
 from dask.utils import natural_sort_key, parse_bytes
 from fsspec.core import get_fs_token_paths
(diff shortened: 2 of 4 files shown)


rjzamora:
Note - I will remove the "[WIP]" label after adding a new test for the bug fixed here.

nvtabular/categorify.py
```diff
@@ -191,7 +191,9 @@ def _write_uniques(dfs, base_path, col, on_host):
         # Make sure first category is Null
         df = df.sort_values(col, na_position="first")
         if not df[col]._column.has_nulls:
-            df = cudf.DataFrame({col: _concat([cudf.Series([None]), df[col]], ignore_index)})
+            df = cudf.DataFrame(
+                {col: _concat([cudf.Series([None], dtype=df[col].dtype), df[col]], ignore_index)}
```
rjzamora, line 195: It seems that the latest cudf branch requires us to be explicit about the type of `None` here.

nvtabular/io.py
```diff
@@ -617,30 +617,18 @@ def _write_metadata(meta_list):
...
-        del ind
-        del gdf
-        # Write each split to a separate file
-        for s, df in result.items():
-            prefix = fs.sep.join([processed_path, "split." + str(s)])
-            pw = get_cache().get_pq_writer(prefix, s, mem=True)
-            pw.write_table(df)
-    else:
+    out_files_per_proc = out_files_per_proc or 1
+    if shuffle and shuffle != "full":
```
benfred, line 621: I think if shuffle == 'full' is passed in we should throw a ```NotImplementedError``` rather than just not shuffle the data at all.

rjzamora: `shuffle == "full"` is currently a full shuffle per worker - So, it is implemented.  The reason we don't sort here, is becuase we are going to read back the in-memory file and sort again before persisting to disk.

rjzamora:
Thanks for the review Ben!

PR #152, opened 2020-07-15
Add a num_rows property to nvt.Dataset

benfred:
The keras dataloader was iterating over the entire dataset to figure out the
number of rows, which was unnecessarily slow. Use the parquet metadata instead
to get this information.


nvtabular/io.py
```diff
@@ -854,6 +862,11 @@ def get_metadata(self):
                     metadata = md
             return metadata, base
 
+    @property
+    def num_rows(self):
+        metadata, _ = self.get_metadata()
```
rjzamora, line 867: While it is certainly faster then loading every partition/chunk, it can still be expensive to re-collect the metadata (in a cloud fs, for example)...  Perhaps we should add a `metadata` property to the ParquetDatasetEngine class to avoid doing this more than once?

benfred: How about caching the results instead? I changed to a cached_property which should have the same result

rjzamora: This does prevent us from re-collecting metadata every time we need the row count.  However, I don't think it addresses the case that we already collected the metadata to produce a ddf, and then ask for the row count after that.. Is that not a likely use case?

rjzamora: I guess what I'm suggesting is that we drop the use of `get_metadata` (perhaps change it to `_get_metadata`) and make the conventional way to access the metadata a `metadata` cached property

jperez999:
rerun tests

nvtabular/io.py
benfred: I've made that change in the most recent commit.  

 I used ```lru_cache``` to memoize the result rather than ```cached_property``` since cached_property seems to be python 3.8+ only and we support earlier versions of python

rjzamora: Sounds good - Thanks for working this out!

PR #154, opened 2020-07-16
[REVIEW] Adding external-join op (LeftJoinExternal)

### LINKED ISSUE #17: [OP] Join onto external dataframe, file, dictionary
<a href="https://github.com/EvenOldridge"></img></a> **Issue by [EvenOldridge](https://github.com/EvenOldridge)**
_Thursday Mar 26, 2020 at 02:22 GMT_
_Originally opened as https://github.com/rapidsai/recsys/issues/32_

----

**Is your operator request related to a problem? Please describe.**
Outbrains dataset

Join to external data in multiple formats.  This mechanism will also be used / shared by https://github.com/rapidsai/recsys/issues/31 to apply the groupby operations.

 - Type: Feature Engineering
 - input column type(s): [Categorical]
 - input options:
   - Columns to join off of:  [Categorical] (just in case naming isn't the same?)
   - Categorical column names to be joined
   - Continuous column names to be joined
 - Expected transformation of the data after application: Data is joined 

**Optional: Describe operation stages in detail***
Statistics per chunk: N/A
Statistics combine: N/A
Apply: Join

Context:
Note that this is meant to be an iterator to gdf join and not an iterator to iterator join which we'll figure out in the future.

rjzamora:
Closes #17

**TODO**:

- [x] Gather feedback on assumptions [happy to expand if needed]
- [x] Add support for multiple "on" columns (MultiIndex join)
- [x] Add support for external tables as "cudf" (done), "pandas", "arrow", "parquet", "csv"
- [x] Use caching for "parquet" and "csv" external tables
- [x] Expand testing as various support is added. Also need to support/test column-name collisions.


benfred:
@rynak can you take a look at this when you have some time? I want to make sure this works for the outbrains workflow

rnyak:
> @rnyak can you take a look at this when you have some time? I want to make sure this works for the outbrains workflow

yes, I was looking at it on Friday. I had questions for Rick. we chatted a bit. will test it shortly.

rnyak:
@rjzamora @benfred we also use `inner` join in Outbrain W&D, which results in only intersection of joined data frames based on a `on` column. can't we have a `how` parameter that we can feed as `inner` or `left`? or we should create another op like `InnerJoin` ?

jperez999:
If we do inner join... what kind of limitations are we going to require? If no limitations will be imposed... are we going to detect the inner merge operation and do some kind of mini sampling to ensure we dont blow out gpu memory? Because the user will create the dataset (specifying part_mem_fraction) separate from the workflow object...

rjzamora:
> If we do inner join... what kind of limitations are we going to require? If no limitations will be imposed... are we going to detect the inner merge operation and do some kind of mini sampling to ensure we dont blow out gpu memory? Because the user will create the dataset (specifying part_mem_fraction) separate from the workflow object...

Unless I am misunderstanind, the inner join should not be any more of a problem than left join (inner is even more restrinctive than left).

I started by assuming (1) we are doing a left join and (2) we are removing duplicates in the external dataframe... The reason was so we will preserve the exact number of rows, to avoid updating all columns to perform the merge.

rjzamora:
> @rjzamora @benfred we also use inner join in Outbrain W&D, which results in only intersection of joined data frames based on a on column. can't we have a how parameter that we can feed as inner or left? or we should create another op like InnerJoin ?

Yes - we should be able to handle inner join and duplicates.  I may need help from @jperez999 to make sure I am updating all columns correctly since the join may add/remove rows.

rjzamora:
@rnyak - Thanks for the feedback (both here and offline).  I cleaned this up a little so you *should* only need to do:

```python
merge_op = ops.JoinExternal(df_ext, on, how="inner")
```

Thanks to @jperez999 for giving me some tips on how to handle the changing row count.

>are we going to detect the inner merge operation and do some kind of mini sampling to ensure we dont blow out gpu memory?

I think it is up to the user to set the initial partition size appropriately if they know they will be increasing the dataset size during preprocessing (not just a problem for join).   We can certainly add a warning message to make this clear, but I have reservations about complicating the op with sampling logic.

rnyak:
@rjzamora as we discussed, drop_duplicates should be applied both to left and right data frames for `inner` join, and only to the `left` data frame for `left` join. Otherwise, joined data frame might have duplicated rows.

rjzamora:
>@rjzamora as we discussed, drop_duplicates should be applied both to left and right data frames for inner join, and only to the left data frame for left join. Otherwise, joined data frame might have duplicated rows.

I added the `drop_duplicates_ext` option to drop duplicates.  This will drop the duplicates in the external table right before the join.  I think we will need to make it the users responsibility to ensure that there are no duplicates in the left table.  Since we will **not** be capturing the possibility of global duplicates in the left table, I think it would be confusing the "support" that option.  

(cc @EvenOldridge, in case you had a different opinion here)

rnyak:
> > @rjzamora as we discussed, drop_duplicates should be applied both to left and right data frames for inner join, and only to the left data frame for left join. Otherwise, joined data frame might have duplicated rows.
> 
> I added the `drop_duplicates_ext` option to drop duplicates. This will drop the duplicates in the external table right before the join. I think we will need to make it the users responsibility to ensure that there are no duplicates in the left table. Since we will **not** be capturing the possibility of global duplicates in the left table, I think it would be confusing the "support" that option.
> 
> (cc @EvenOldridge, in case you had a different opinion here)

@rjzamora thank you, will test it with added the `drop_duplicates_ext` option.

nvtabular/ops.py
```diff
@@ -801,6 +803,122 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None
         return new_gdf
 
 
+class JoinExternal(TransformOperator):
+    """
+    Join each dataset partition to an external table. For performance
+    reasons, only "left" and "inner" join transofrmations are supported.
```
benfred, line 809:
```suggestion
    reasons, only "left" and "inner" join transformations are supported.
```

rjzamora:
Thanks for your help with this @rnyak - I'll be happy to open a follow-up PR if you need further improvements.

Issue #51, opened 2020-06-04
[Task] Write global parquet metadata file 


benfred:
<a href="https://github.com/benfred"><img src="https://avatars2.githubusercontent.com/u/69536?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [benfred](https://github.com/benfred)**
_Friday May 22, 2020 at 19:52 GMT_
_Originally opened as https://github.com/rapidsai/recsys/issues/168_

----

We should write out a global metadata file when exporting a parquet dataset.

benfred:
<a href="https://github.com/benfred"><img src="https://avatars2.githubusercontent.com/u/69536?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [benfred](https://github.com/benfred)**
_Wednesday May 27, 2020 at 00:04 GMT_

----

See also https://github.com/rapidsai/cudf/issues/5284

PR #155, opened 2020-07-16
[REVIEW] Write global _metadata file for Dask-based parquet output

### LINKED ISSUE #59: [FEA] Write shared "_metadata" file for output parquet data in DaskWorkflow
<a href="https://github.com/rjzamora"></img></a> **Issue by [rjzamora](https://github.com/rjzamora)**
_Thursday May 28, 2020 at 16:43 GMT_
_Originally opened as https://github.com/rapidsai/recsys/issues/179_

----

The `DaskWorkflow` implemntation introduced in #136 does not currently write a shared parquet `"_metadata"` file when the `to_dataset` API is used.  This is because of the limitation discussed in [cudf#284](https://github.com/rapidsai/cudf/issues/5284).  Once that issue is resolved upstream, the `DaskWorkflow` code should be modified to write the shared file (having this file is a significant performance boost if/when dask is used to read the dataset back in).

Once the upstream changes are made, metadata file creation will look very similar to [this existing cudf method](https://github.com/rapidsai/cudf/blob/56769d4755091e3fe892fcdd020bd1e2a9a19cfe/python/dask_cudf/dask_cudf/io/parquet.py#L130).

The alternative is to skip `"_metadata"` creation, or to create the shared file with pyarrow after the fact (requiring a followup python process to parse all the footer metadata after it is written).

rjzamora:
Closes #59 
Address Dask component of #51

Now that [cudf#5411](https://github.com/rapidsai/cudf/pull/5411) is merged, we can/should write the global _metadata file after writing out a (processed/shuffled) parquet dataset.  This PR covers the Dask-related changes (note that we still need to align the single/multi-gpu IO mechanisms in the future).

### DIFF
--- a/nvtabular/io.py
+++ b/nvtabular/io.py
@@ -25,6 +25,7 @@ import warnings
 from collections import defaultdict
 from io import BytesIO
 from itertools import islice
+from threading import Lock
 from uuid import uuid4
 
 import cudf
@@ -48,6 +49,7 @@ from fsspec.utils import stringify_path
 # Use global variable as the default
 # cache when there are no distributed workers
 DEFAULT_CACHE = None
+lock = Lock()
 
 
 LOG = logging.getLogger("nvtabular")
@@ -562,6 +564,7 @@ class WriterCache:
             pw.close()
 
     def get_pq_writer(self, prefix, s, mem):
+        lock.acquire()
         pw, fil = self.pq_writer_cache.get(prefix, (None, None))
         if pw is None:
             if mem:
@@ -573,6 +576,7 @@ class WriterCache:
                 full_path = ".".join([prefix, outfile_id])
                 pw = pwriter(full_path, compression=None)
                 self.pq_writer_cache[prefix] = (pw, full_path)
+        lock.release()
         return pw
 
 
--- a/nvtabular/io.py
+++ b/nvtabular/io.py
@@ -616,9 +616,10 @@ def clean_pw_cache():
 
 def close_cached_pw(fs):
     md_dict = {}
-    for path, (pw, bio) in get_cache().pq_writer_cache.items():
-        fn = bio.split(fs.sep)[-1]
-        md_dict[fn] = pw.close(metadata_file_path=fn)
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -47,7 +47,7 @@
 def client():
     global _CLIENT
     if _CLIENT is None:
-        _CLIENT = Client(LocalCluster(n_workers=2))
+        _CLIENT = Client(LocalCluster(n_workers=2, threads_per_worker=1))
```
rjzamora, line 50: Turns out we get race conditions during worker caching when we allow them to use multiple threads.

benfred: Should we put a lock around the worker cache instead to handle this? (Or make the worker cache thread local?)

rjzamora: Sorry - I am probably wring to call this a "race condition."  For some reason, the worker can end up creating >1 ParquetWriter objects for the same output split.  Everything seems to "work" fine, but the number of output files ends up being wrong.

Not really sure how to add a lock here.  It seems that I can avoid the issue if I add a wait and retry block after the first call to `pq_writer_cache.get` in `get_pq_writer`.  I'll try to figure out the "correct" way to ensure synchronization

rjzamora: Okay - I am adding a lock for now, because that seems to work fine.

nvtabular/io.py
```diff
@@ -612,17 +614,21 @@ def clean_pw_cache():
         return
 
 
+def close_cached_pw(fs):
+    md_dict = {}
+    with get_cache() as cache:
+        for path, (pw, bio) in cache.pq_writer_cache.items():
+            fn = bio.split(fs.sep)[-1]
```
benfred, line 621: Can't bio be a bytes IO object if 'mem=True' when calling get_pq_writer? Should we use path instead?

rjzamora: I can add an assertion, but this function should never be called for the "full" shuffle. bio will always be a path here.

PR #158, opened 2020-07-17
Refactor Workflow-output API - Prepare for ThreadedWriter use in Dask

### LINKED ISSUE #51: [Task] Write global parquet metadata file

rjzamora:
This PR refactors some critical components of the `Workflow` class to simplify the single-GPU writers.  The goal of this PR is to (1) make it easier to leverage the single-GPU writer code within Dask, and (2) "simplify" the Workflow API a bit.

There are a lot of changes here, so I decided to stop short of actually chagning the Dask-based IO to use the single-gpu writer classes.

Closes #51

**General `Workflow` Changes**:

- Moving all iteration-based processing into `iterate_online` method
- Moving all dask-graph-based processing into `build_and_process_graph` method
- Simplifying `apply` method to call one of the above methods
- Narrowing scope of `update_stats` method (using this means the user **only** wants to update stats, not shuffle/write data or apply unnecessary ops)

**IO-Specific Changes**:

- Removing `ds_writer.py`
- Removing `Shuffler` (I feel that the shuffling strategy should be decoupled from the writer a bit more, for simplicity)
- Introducing a `output_format` argument to `Workflow` methods - `iterate_online` uses this arg to specify the `writer` to use.
- Refactoring `ThreadedWriter` to decouple metadata packaging/writing - This will make it easier to write global metadata with Dask
- Added global _metadata write for `ParquetWriter`  (Addressed #51)
- Moving the `write_to_dataset` definition from `BaseWorkflow` to `Workflow`, and using `iterate_online` or `build_and_process_graph` to (optionally) apply ops and write a dataset (with a desired format)
    - `write_to_dataset` does **not** allow statistics gathering - the statistics must already be available. When users want to do stats/transofrms/write altogether, they should use `apply`.


rjzamora:
@jperez999  - Let me know if the CI problem is on my end, otherwise I'm assuming it isn't (for now)

jperez999:
rerun tests

jperez999:
rerun tests

nvtabular/io.py
```diff
@@ -314,48 +398,47 @@ def _set_dtypes(self, chunk):
...
+        num_out_files=30,
+        num_threads=4,
+        cats=None,
+        conts=None,
+        labels=None,
+        shuffle=None,
+        fs=None,
     ):
         # set variables
+        self.hugectr_bin = False
```
benfred, line 436: Whats this hugectr_bin flag for? Can we remove it?

nvtabular/io.py
```diff
@@ -458,38 +598,28 @@ def _write_thread(self):
...
 
-        json.dump(data, metadata_writer)
-        metadata_writer.close()
+        # Aggregate metadata and write _metadata file
+        _write_pq_metadata_file(md_list, fs, out_dir)
+
+    def _close_writers(self):
+        md_dict = {}
+        for writer, fn in zip(self.data_writers, self.data_fns):
+            md_dict[fn] = writer.close(metadata_file_path=fn)
```
benfred, line 613: Is this right? Isn't data_fns just the filename of the parquet file without the path, and if so should we be using self.data_paths instead (and removing self.data_fns entirely)?

nvtabular/io.py
```diff
@@ -412,39 +504,87 @@ def add_data(self, gdf):
...
-    def __init__(
-        self, out_dir, num_out_files=30, num_threads=4, cats=None, conts=None, labels=None
-    ):
-        super().__init__(out_dir, num_out_files, num_threads, cats, conts, labels)
-        self.data_files = [os.path.join(out_dir, f"{i}.parquet") for i in range(num_out_files)]
-        self.data_writers = [pwriter(f, compression=None) for f in self.data_files]
+    def __init__(self, out_dir, use_guid=False, **kwargs):
+        super().__init__(out_dir, **kwargs)
+        self.data_fns = []
+        self.data_paths = []
```
benfred, line 577: Why do we need to store both data_fns and data_paths?

nvtabular/workflow.py
```diff
@@ -734,83 +718,92 @@ def apply(
...
+            elif output_format == "hugectr":
+                writer_cls = nvt_io.HugeCTRWriter
+            else:
+                raise ValueError("Output format not yet supported.")
+            output_path = output_path or "./"
+            output_path = str(output_path)
+            fs = get_fs_token_paths(output_path)[0]
+            writer = writer_cls(
+                output_path, num_out_files=out_files_per_proc, shuffle=shuffle, fs=fs
+            )
```
benfred, line 769: Can we move this block of code into a factory function in nvtabular/io.py ?  Something like a function that takes the output_format and output_path and returns a Writer object.

nvtabular/io.py
rjzamora: Yeah, sorry - Didn't mean to leave this in!

nvtabular/io.py
rjzamora: The path for each column chunk that is written to the global _metadata file must be a relative path.  Since we will write _metadata in the same directory as the data files, we only want to write the file names.

nvtabular/io.py
rjzamora: We don't need to, but it was convenient to access the file names this way (when updating the metadata for the global _metadata file) -- I'll see if I can just clip the file name off the end of the full path.

nvtabular/workflow.py
```diff
@@ -734,83 +718,92 @@ def apply(
...
-                        # This statistics depends on a transform
-                        # (Opt wont work)
-                        return
-                elif isinstance(task[0], TransformOperator):
-                    trans_tasks.append(task)
+        # Close writer and write general/specialized metadata
+        if writer:
+            general_md, special_md = writer.close()
+            writer_cls.write_special_metadata(special_md, fs, output_path)
+            writer_cls.write_general_metadata(general_md, fs, output_path)
```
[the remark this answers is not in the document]
rjzamora: I know we talked about this a bit offline, but you are right that this needs to be documented a bit better in the code (or just improved in another way)...  We are defining/using these class-level functions to allow the same `writer` class(es) to be used by Dask.   That is, if we want to enable Dask to use the same IO funcitonality, we need to avoid writing shared metadata on every worker, and instead aggregate the metadata on  the client process before writing out a single file.

nvtabular/io.py
rjzamora: I removed `self.data_fns`

nvtabular/workflow.py
rjzamora: Good suggestion - I added a `writer_factory` function in `io.py`.  Let me know what you think

tests/unit/test_notebooks.py
```diff
@@ -35,6 +35,7 @@ def test_optimize_criteo(tmpdir):
 
 
 def test_rossman_example(tmpdir):
+    pytest.importorskip("tensorflow")
```
benfred, line 38: thanks for this!

nvtabular/workflow.py
```diff
@@ -734,83 +718,87 @@ def apply(
...
-        self.phases[0] = stat_tasks
-        self.phases[1] = trans_tasks
+            # Note that we "could" have the special and general metadata
+            # written during `writer.close()` (just above) for the single-GPU case.
+            # Instead, the metadata logic is separated from the `Writer` object to
+            # simplify multi-GPU integration. When using Dask, we cannot assume
+            # that the "shared" metadata files can/will be written by the same
+            # process that writes the data.
+            type(writer).write_special_metadata(special_md, writer.fs, output_path)
+            type(writer).write_general_metadata(general_md, writer.fs, output_path)
```
benfred, line 775: Thanks for adding the comment!

fwiw, you can call classmethods on instances of that class so the ```type(writer)``` isn't strictly necessary (though obviously still works).

rjzamora: Ah right - Forgot about that.

PR #160, opened 2020-07-21
[REVIEW] Align Dask and Single-GPU Writer Logic

rjzamora:
This PR simplifies the Dask-based dataset output code to use the single-GPU `Writer` class(es).  We also change the HugeCTR "metadata" file names to be "_metadata.json" and "_file_list.txt" to avoid pyarrow/parquet dataset-discovery problems.

**TODO**:

- [x] Properly test "_metadata.json" and "_file_list.txt" content (json file may be wrong)
- [x] Use "full" and "partial" shuffle language for single-gpu case - Enable to full (per-worker) shuffle without Dask


rjzamora:
@oyilmaz-nvidia @albert17 - Note that the file name change I mentioned offline is included here.

rjzamora:
Lets give @benfred a chance to look this over before we merge :)

PR #181, opened 2020-07-28
[REVIEW] Add single-threaded code path for writing output dataset

rjzamora:
In order to reduce memory pressure for larger partition sizes, this PR adds a (default) code path to avoid using multiple threads for writing an output dataset.  The user can now specify the number of threads to use for writing by passing `num_io_threads` to Workflow methods like `apply`.

### DIFF
--- a/nvtabular/io.py
+++ b/nvtabular/io.py
@@ -882,10 +882,12 @@ class CSVDatasetEngine(DatasetEngine):
         Thin wrapper around dask_cudf.read_csv.
     """
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args)
+    def __init__(self, paths, part_size, storage_options=None, **kwargs):
+        super().__init__(paths, part_size, storage_options)
         self._meta = {}
         self.csv_kwargs = kwargs
+        self.csv_kwargs["storage_options"] = storage_options
+
         # CSV reader needs a list of files
         # (Assume flat directory structure if this is a dir)
         if len(self.paths) == 1 and self.fs.isdir(self.paths[0]):
--- a/nvtabular/io.py
+++ b/nvtabular/io.py
@@ -235,8 +235,9 @@ class ThreadedWriter(Writer):
         # Resolve file system
         self.fs = fs or get_fs_token_paths(str(out_dir))[0]
 
-        # Only use threading if num_threads > 0
-        if self.num_threads:
+        # Only use threading if num_threads > 1
+        self.queue = None
+        if self.num_threads > 1:
 
             # create thread queue and locks
             self.queue = queue.Queue(num_threads)
@@ -277,7 +278,7 @@ class ThreadedWriter(Writer):
         nrows = gdf.shape[0]
         typ = np.min_scalar_type(nrows * 2)
         if self.shuffle and self.shuffle != "full":
-            ind = cp.random.choice(np.arange(self.num_out_files, dtype=typ), nrows)
+            ind = cp.random.choice(cp.arange(self.num_out_files, dtype=typ), nrows)
         else:
             ind = cp.arange(nrows, dtype=typ)
             cp.floor_divide(ind, (nrows // self.num_out_files), out=ind)
@@ -287,7 +288,7 @@ class ThreadedWriter(Writer):
             gdf.scatter_by_map(ind, map_size=self.num_out_files, keep_index=False)
         ):
             self.num_samples[x] += len(group)
-            if self.num_threads:
+            if self.num_threads > 1:
                 self.queue.put((x, group))
             else:
                 self._write_table(x, group)
@@ -295,7 +296,7 @@ class ThreadedWriter(Writer):
 
         # wait for all writes to finish before exiting
         # (so that we aren't using memory)
-        if self.num_threads:
+        if self.num_threads > 1:
             self.queue.join()
 
     def package_general_metadata(self):
@@ -351,7 +352,7 @@ class ThreadedWriter(Writer):
         return None
 
     def close(self):
-        if self.num_threads:
+        if self.num_threads > 1:
             # wake up all the worker threads and signal for them to exit
             for _ in range(self.num_threads):
                 self.queue.put(self._eod)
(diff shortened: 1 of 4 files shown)


rjzamora:
Happy to change the default `num_io_threads` setting here, but I think it's necessary to have access to this option at the `Wrokflow` level

rjzamora:
@benfred - I am interested to get your thoughts here.

This PR was motivated by a comment from @jperez999 that the criteo-example notebook was taking much longer than the 16-17min he experienced in the past.  His comments led me to discover a few things:

1. The switch to the `ThreadedWriter`-based writers in the dask-based apply was increasing the memory overhead at write time, requiring the use of a smaller `part_mem_fraction` setting.  I discovered that the the "safest" approach is to avoid the use of threads (which is why I made that the default here).  However, it seems that the best performance comes with `num_io_threads=2`.
2. A small part of the performance regression is related to a small change in the way `ThreadedWriter` divides data between output files.  It seems that cudf's `scatter_by_map` method is a fairly optimal way to shuffle data into contiguous DataFrames.  Switching back to this approach gave me a 30s performance improvement, so I am including that change here.
3.  It turns out that using `Categorify` with `freq_threshold=0` (full encoding) is **much** faster than `freq_threshold>0` (partial encoding).  It seems that the `searchsorted`-based operation available for full encoding requires significantly less data copying than the merge-based operation needed for partial encoding.

nvtabular/io.py
```diff
@@ -228,55 +235,68 @@ def __init__(
         # Resolve file system
         self.fs = fs or get_fs_token_paths(str(out_dir))[0]
 
-        # create thread queue and locks
-        self.queue = queue.Queue(num_threads)
-        self.write_locks = [threading.Lock() for _ in range(num_out_files)]
+        # Only use threading if num_threads > 0
+        if self.num_threads:
```
benfred, line 239: Should we be running the multithreaded version with num_threads=1 (which it will right now)?

nvtabular/workflow.py
```diff
@@ -664,7 +664,9 @@ def apply(
         output_path="./ds_export",
         output_format="parquet",
         out_files_per_proc=None,
-        **kwargs,
+        num_io_threads=0,
+        nsplits=None,  # Deprecated
+        num_out_files=None,  # Deprecated
```
benfred, line 669: Is nsplits or num_out_files used anywhere? Can we just remove them rather than mark as deprecated?

nvtabular/io.py
```diff
@@ -228,55 +235,68 @@ def __init__(
...
-            self.num_samples[x] = self.num_samples[x] + to_write.shape[0]
-            self.queue.put((x, to_write))
+        # Generate `ind` array to map each row to an output file.
+        # This approach is certainly more optimized for shuffling
+        # than it is for non-shuffling, but using a single code
+        # path is probably worth the (possible) minor overhead.
+        nrows = gdf.shape[0]
+        typ = np.min_scalar_type(nrows * 2)
+        if self.shuffle and self.shuffle != "full":
+            ind = cp.random.choice(np.arange(self.num_out_files, dtype=typ), nrows)
```
benfred, line 280: Can we generate the range on the device (like cp.arange instead of np.arange)?  
```suggestion
            ind = cp.random.choice(cp.arange(self.num_out_files, dtype=typ), nrows)
```

nvtabular/io.py
```diff
@@ -228,55 +235,68 @@ def __init__(
...
-        # create thread queue and locks
-        self.queue = queue.Queue(num_threads)
-        self.write_locks = [threading.Lock() for _ in range(num_out_files)]
+        # Only use threading if num_threads > 0
+        if self.num_threads:
 
-        # signifies that end-of-data and that the thread should shut down
-        self._eod = object()
+            # create thread queue and locks
+            self.queue = queue.Queue(num_threads)
```
benfred, line 242: nitpick: we should set the queue to None when not in use (rather than just not have on the object at all)

nvtabular/workflow.py
```diff
@@ -695,15 +697,16 @@ def apply(
...
+        num_io_threads : integer
+            Number of IO threads to use for writing the output dataset.
+            For `0` (default), no dedicated IO threads will be used.
         """
 
-        # Deal with single-gpu compatibility
-        nsplits = kwargs.get("nsplits", None)
+        # Deal with 0.1.1 compatibility
         if nsplits:
             warnings.warn("nsplits is deprecated. Use out_files_per_proc")
```
benfred, line 707: If we're going to keep this warning in place (instead of just removing the old arguments), can we use a deprecation warning instead? 
```suggestion
            warnings.warn("nsplits is deprecated. Use out_files_per_proc", DeprecationWarning)
```

nvtabular/io.py
rjzamora: Yes.  I went "back and forth" a bit on this, but ultimately decided it makes sense to allow a single-threaded "multithreaded" case for debugging/testing purposes.

However, happy to avoid multithreding for anything < 2

nvtabular/workflow.py
rjzamora: Sure - I was still allowing users to pass these args but was complaining about it in a warning.  I am happy to remove them altogether now since so much of the api is changing anyway :)

nvtabular/io.py
rjzamora: oops typo

nvtabular/workflow.py
rjzamora: you've convinced me - lets just yank the old args :)

rjzamora:
> How much does perf increase with using 2 threads for IO?

I'll have to re-run to get apples-to-apples comparisons.  My rough estimate was something like 5%

> how much faster is full encoding than partial encoding? Can we make the partial encoding faster?

This is another case where apples-to-apples is tough, because the merge-based operation requires more memory than the searchsorted-based op - meaning we can handle larger partition sizes for full encoding.  Using the dask benchmark with one device, I was seeing a difference of roughtly 13mins vs 17mins

PR #187, opened 2020-07-30
[REVIEW] Multi-column groupby support

### LINKED ISSUE #121: [OP] Categorify multi-column support
Categorify is applied only to single columns - There are use cases, when a user want to apply Categorify to multiple columns.
For example in social media, where a user can write a post and another user can engage with a post.
The dataset has two columns:
- author: userid
- engaging user: userid

| Author  | Engaging User | Post |
| ------------- | ------------- | ------------- |
| User_A  | User_B  | 1 |
| User_A  | User_C  | 2 |
| User_B  | User_A  | 3 |
| User_C  | User_D  | 4 |

Both columns have userids, but I do not want to categorify them independently.

The result should be

| Author  | Engaging User | Post |
| ------------- | ------------- | ------------- |
| 0  | 1  | 1 |
| 0  | 2  | 2 |
| 1  | 0  | 3 |
| 2  | 3  | 4 |

I want to categorify both column, together.
I concat Author and Engaging User, factorize them and then apply the mapping to both columns

**Additional context**
Pseudocode:
```
uid = pd.factorize( np.concatenate( ( df['a_user_id'].values, df['b_user_id'].values ) ) )[0]
df['a_user_id'] = uid[:df.shape[0] ]
df['b_user_id'] = uid[df.shape[0]:]
```

rjzamora:
closes #121 

This PR aims to address **three** different cases requiring "mutli-column" groupby statistics:

1. **"Joint" Encoding**: As requested in #121, the user may want to perform a "joint" encoding of multiple categorical columns. this is relevant when the elements in distinct columns correspond to the same "kind" of data.  In this case, we want to concatenate the targeted columns to perform the groupby operation (in `GroupbyStatistics`).

2. **"Combo" Encoding**: The user may also want to perform a groupby on two different kinds of columns.  This would be required to convert 2+ categorical columns into a single encoded column, or to join groupby-aggregation statistics on multiple reference columns.

3. **Multi-column Groupby Join** (self explanatory - uses "combo" version of Groupby aggregation)

**TODO**:

- [x] Support both multi-column groupby scenarios in `CategoryStatistics` (`concat_groups` arg controls multi-column behavior)
- [x] Support `encode_type="joint"` for multi-column `Categorify`
- [x] Support `encode_type="combo"` for multi-column `Categorify`
- [x] Support multi-column `GroupBy` Op (now called `JoinGroupby`)

**RFC**:

- [ ] This PR renames `CategoryStatistics` to `GroupbyStatistics` -- Okay ?
- [ ] This PR renames `GroupBy` to `JoinGroupby` -- Okay ?
- [ ] This PR dissallows replacement for `Categorify` for `encode_type="combo"` -- Okay?


rjzamora:
CI should start passing when the latest cudf is picked up :)

rnyak:
@rjzamora Thanks so much for this PR.  I did a small test with a toy example. The resulted `new_gdf`  from `JoinGroupby` op does not seem correct. This is my test:

```
gdf = cudf.DataFrame({'A': [1, 2, 2, 6, 6, 8, 3], 'B': [2, 4, 4, 7, 7, 2, 5], 'C': [1.0 , 4.0, 4.0, 8.0, 8.0, 5.0, 1.0]})

proc = nvt.Workflow(
    cat_names=['A', 'B'],
    cont_names=['C'],
    label_name=[])

#drop_duplicates with groupby
proc.add_feature(JoinGroupby(columns=['A', 'B'], 
                             cont_names=['C'], 
                             stats=['count']))

train_dataset = nvt.Dataset(gdf, engine='parquet')

proc.apply(train_dataset, apply_offline=True, record_stats=True, output_path='./dataset/test/', shuffle=False, out_files_per_proc=1)
```
This is generated new_gdf:

```
new_gdf= cudf.read_parquet('./dataset/test/*.parquet')
new_gdf.head()
	C	A	B	A_count	B_count
0	1.0	1	2	1	2
1	4.0	2	4	2	2
2	4.0	2	4	2	2
3	8.0	6	7	2	2
4	8.0	6	7	2	2
```
The expected output should be like below (we group the gdf based on A and B columns, apply count on the C column):
```
	A	B	C	C_count
0	1	2	1.0	1
1	2	4	4.0	2
2	2	4	4.0	2
3	6	7	8.0	2
4	6	7	8.0	2
5	8	2	5.0	1
6	3	5	1.0	1
```

also from `nvt.ops.GroupbyStatistics()` we should be getting `cat_stats.C.parquet`, but I get `cat_stats.A.parquet` and `cat_stats.B.parquet` stat files. 

Please let me know if I am missing or confusing anything here.

rjzamora:
We only perform a “count” aggregation on the groups (cat colums), since the “count” of “C” is just the count of the group (the values in “C” make no difference). If you asked for a “sum“ or “std” aggregation, you would see a “C”-specific agg column, but the name will still specify a corresponding cat column.

Also, we only write parquet files for each Cat group, not each continuous column.  So, you should not see a cat_stats.C.parquet - the C-specific stats must be in the other files.  However, since you only asked for “count”, there are no C-specific stats

rnyak:
@rjzamora thanks.  I agree on this logic: `We only perform a “count” aggregation on the groups (cat colums)`. But still I am a bit confused on the generated results. Let's talk offline.

rjzamora:
>@rjzamora thanks. I agree on this logic: We only perform a “count” aggregation on the groups (cat colums). But still I am a bit confused on the generated results. Let's talk offline.

Sounds good.  In case it helps... I jut realized this discussion is under the "Multi-column" PR, so you probably intended for the results to correspond to a **multi-column** groupby.  If so, you need to do: `JoinGroupby(columns=[['A', 'B']], ...)` (note that the outer list allows multiple column groups, and the inner list is the actual multi-column group).  If you make that change, you should get a `"cat_stats.A_B.parquet"` file that looks like:

```python

   A  B  A_B_count
0  1  2          1
1  2  4          2
2  3  5          1
3  6  7          2
4  8  2          1
```

This probably gives you waht you expect after the merge.

PR #190, opened 2020-07-31
[REVIEW] Handle empty cat/cont column lists in pytorch loader

### LINKED ISSUE #149: [BUG] proc.ds_to_tensors does not support empy category/numeric columns
**Describe the bug**
proc.ds_to_tensors expect that categorical/numerical columns are defined.
I think there are datasets, 
* which does not have a categorical column - e.g. all columns are numeric.
* which does not have a numeric column - e.g. all columns are categorical

```
dataset = nvt.Dataset('./train.csv')
proc = nvt.Workflow(
        cat_names=['ROLE_ROLLUP_1'],
        cont_names=[],
        label_name=['ACTION']
)
proc.finalize()
proc.ds_to_tensors(dataset.to_iter())
```

Throws following error:
```
/opt/conda/envs/rapids/lib/python3.6/site-packages/nvtabular/torch_dataloader.py in _get_final_cols(preproc)
    154         preproc.create_final_cols()
    155     cat_names = _get_embedding_order(preproc.columns_ctx["final"]["cols"]["categorical"])
--> 156     cont_names = sorted(preproc.columns_ctx["final"]["cols"]["continuous"])
    157     label_name = sorted(preproc.columns_ctx["final"]["cols"]["label"])
    158     return cat_names, cont_names, label_name

TypeError: 'NoneType' object is not iterable
```

If I run, then it works.
```
dataset = nvt.Dataset('./train.csv')
proc = nvt.Workflow(
        cat_names=['ROLE_ROLLUP_1'],
        cont_names=['ROLE_ROLLUP_1'],
        label_name=['ACTION']
)
proc.finalize()
proc.ds_to_tensors(dataset.to_iter())
```

I know, that the error is not related to dataset.to_iter() - this works - it is related to proc.ds_to_tensors

**Expected behavior**
I think the expected behavior should be that it supports empty lists for categorical and numerical features

benfred:
Fixes #149


nvtabular/torch_dataloader.py
```diff
@@ -60,23 +59,23 @@ def __len__(self):
...
 
-def _to_tensor(gdf: cudf.DataFrame, dtype, to_cpu=False):
+def _to_tensor(gdf: cudf.DataFrame, dtype):
     if gdf.empty:
         return
     g = gdf.to_dlpack()
-    cols = gdf.columns
     t = from_dlpack(g).type(dtype)
     del g, gdf
-    return t, cols
```
jperez999, line 79: we going to want these columns later... in order to create the columns / tensors class (aka dictionary)... wont we?

benfred: We weren't using it: the only time we called _to_tensor we didn't use that return value (and went  ```return [cats[0], conts[0], label[0]]``` to only access the tensor and not the columns). Its simpler to just not include it here

benfred:
rerun tests

nvtabular/torch_dataloader.py
benfred: I think we will have to have a substantial rewrite before using a columns/tensor class - and we're not using this info now, and it's simpler just to leave off in the meantime

nvtabular/torch_dataloader.py
```diff
@@ -44,7 +43,7 @@ def __init__(self, tensors, batch_size=1, pin_memory=False, shuffle=False):
         self.tensors = tensors
         self.batch_size = batch_size
 
-        self.num_samples = self.tensors[0].size(0)
+        self.num_samples = self.tensors[2].size(0)
```
jperez999, line 46: why this change?

nvtabular/torch_dataloader.py
```diff
@@ -60,23 +59,23 @@ def __len__(self):
 
     def __iter__(self):
         for idx in range(0, self.num_samples, self.batch_size):
-            tens = [tensor[idx : idx + self.batch_size] for tensor in self.tensors]
-            yield tens[0], tens[1], tens[2]
-            del tens
+            yield [
```
jperez999, line 62: when you run the actual notebooks does this cause any OOM errors especially with bigger batch_sizes? I think thats why I put the `del tens`.

benfred: I got rid of the temporary variable - so don't need to explicitly clean it up. I *think* this should be good

nvtabular/torch_dataloader.py
benfred: Tensors are a tuple of (cats, conts, labels). Since this change is allowing the cats/conts to be optional (and setting to None if we don't have any) - we can't take the size of those to get the number of samples. The labels always have to be there, so I'm using that instead.

Issue #58, opened 2020-06-04
[OP] add isnull() and query() ops


benfred:
<a href="https://github.com/rnyak"><img src="https://avatars1.githubusercontent.com/u/16246900?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [rnyak](https://github.com/rnyak)**
_Wednesday May 27, 2020 at 22:54 GMT_
_Originally opened as https://github.com/rapidsai/recsys/issues/177_

----

**Is your operator request related to a problem? Please describe.**

The operators below are commonly used, and they will help us to do preprocs for Outbrain W&D model.

1. isnull()
2. query()

Examples:

**1.** Using `isnull()` we can return a dataframe with `source_id` rows do not have any corresponding `publisher_id`:

```
gdf = cudf.DataFrame({'publisher_id': [1, 2, np.nan, 3, np.nan],
                     'ad_id': [1, 2, 1, 3, 2],
                     'source_id': [603, 600, 601, 602, 603],
                     'ad_type': cudf.Series(["movie", "music", "movie", "fashion", "music"], dtype='category')})

gdf[gdf.publisher_id.isnull()]['source_id']

2    601
4    603
Name: source_id, dtype: int64
```

**2.** we should be able to select values from a DataFrame where a given condition is met, via the query API. For exp:

```
gdf.query("publisher_id > 2")

        publisher_id	ad_id	source_id	ad_type
3	3	             3	    602	    fashion
```

**Describe the solution you'd like**
A clear and concise description of the operation you'd like to perform on the column.  Please include:
 - Type: Feature Engineering 
 - input column type(s): Categorical & Continuous
 - output column type(s): Categorical & Continuous
 - Expected transformation of the data after application: 

1. Detect missing values, and return a dataframe accordingly.
2. Query the columns of a DataFrame with a boolean expression.

PR #191, opened 2020-08-04
[REVIEW] adding Filter op

rnyak:
Adding IsNull op wrt https://github.com/NVIDIA/NVTabular/issues/58.

@benfred can you check it out?  as we talked, this helps with outbrain case, but for filter, I am on working on creating a LambdaOp.


benfred:
The code in nvtabular/ops.py looks good - but can you add a unittest to check that this is working as we want ? 

Also I think we should only have a single outbrains example notebook, can you consolidate the notebooks to one and indicate that its for outbrains w&d in the filename?

rnyak:
> The code in nvtabular/ops.py looks good - but can you add a unittest to check that this is working as we want ?
> 
> Also I think we should only have a single outbrains example notebook, can you consolidate the notebooks to one and indicate that its for outbrains w&d in the filename?

Thanks. Yes will add the unit test.

For outbrain, we might need a folder for preprocessing and training sections :) yes we can combine all preprocs in one notebook, which will be pretty long actually.

rnyak:
@benfred @jperez999 I also added Filter op. The usage example would be:

```
gdf = cudf.DataFrame({'A': [1, 2, 2, 9, 6, np.nan, 3], 'B': [2, np.nan, 4, 7, 7, 2, 5], 'C': [1,4, 4, 8, 8, 5, 1], 'D':['a','b', 'c', np.nan, np.nan, 'g', 'k']})

proc = nvt.Workflow(
    cat_names=[],
    cont_names=['A', 'B', 'C', 'D'],
    label_name=[])

proc.add_preprocess(
    [
    Filter(
    #f=lambda gdf: gdf[gdf.A > 5]
    f=lambda gdf: gdf[~gdf.A.isin([2,5])]
    #f=lambda gdf: gdf[gdf.D.notnull()]
    )
    ]
)
train_dataset = nvt.Dataset(gdf, engine='parquet')

proc.apply(train_dataset, apply_offline=True, record_stats=True, output_path='./output_test/')

new_gdf= cudf.read_parquet('./output_test/*.parquet')
```

I will add unit tests once all is working. Let me know if something missing. Thanks.

nvtabular/ops.py
```diff
@@ -1102,3 +1102,43 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None
             new_gdf[col] = self.f(gdf[col], gdf)
         new_gdf.columns = [f"{col}_{self._id}" for col in new_gdf.columns]
         return new_gdf
+
+
+class Filter(TransformOperator):
+    """
+    Enables to call Methods to cudf.Series. Returns a subset of the original gdf.
```
benfred, line 1109: I think the docstring here could be made clearer: how about

```suggestion
    Filters rows from the dataset. This works by taking a callable that takes a dataframe, and returns 
    a dataframe with unwanted rows filtered out.
```
or something like that?

nvtabular/ops.py
```diff
@@ -1102,3 +1102,43 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None
...
+
+    Parameters
+    -----------
+    f : lambda function
+        defines the function executed on dataframe level, expectation is lambda gdf: gdf[gdf.col...]
+        col is the cudf.Series defined by the context
+        gdf is the full cudf.DataFrame
```
benfred, line 1116: I think we could improve this comment - we don't need to talk about cudf.Series (since this Filter op could work on multiple series/columns) and the example given doesn't show how to filter any rows. How about:

```suggestion
    f : callable
        Defines a function that filter rows from a dataframe. As an example  ```lambda gdf: gdf[gdf.a >= 0]``` would 
        filter out all rows that had a negative value in the ```a``` column.
```

nvtabular/ops.py
```diff
@@ -1102,3 +1102,43 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None
...
+    preprocessing : bool, default True
+        Sets if this is a pre-processing operation or not
+    replace : bool, default True
+        Replaces the transformed column with the original input
+        if set Yes
```
benfred, line 1121: I realize that ```replace``` is needed by the base class, but I don't think it makes sense in this context (like replace=False doesn't create new columns). Maybe just remove as a method parameter and always pass replace=True to the base class?

nvtabular/ops.py
```diff
@@ -1102,3 +1102,43 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None
...
+    """
+
+    default_in = ALL
+    default_out = ALL
+
+    def __init__(self, f, preprocessing=True, replace=True):
+        super().__init__(preprocessing=preprocessing, replace=replace)
+        if f is None:
+            raise ValueError("f cannot be None. LambdaOp op applies f to dataframe")
```
benfred, line 1130:
```suggestion
            raise ValueError("f cannot be None. Filter op applies f to dataframe")
```

benfred:
rerun tests

bschifferer:
I tested the FilterOp. The FilterOp stand alone works.
I tried out chaining - see comments below. I am not sure, if this is an issue or not.

The Filter op works, when there are previous ops:
`proc = nvt.Workflow(
    cat_names=['cat'],
    cont_names=['num'],
    label_name=[]
)

proc.add_preprocess(
    [
        ops.LambdaOp(
            op_name='plus100',
            f=lambda col, gdf: col+100, 
            columns=['num'], 
            replace=False),
        ops.Filter(f=lambda gdf: gdf[gdf['num_plus100']>100]),
    ]
)`

Ops, chained after the filter op, do not get a context passed:
`proc = nvt.Workflow(
    cat_names=['cat'],
    cont_names=['num'],
    label_name=[]
)

proc.add_preprocess(
    [
        ops.LambdaOp(
            op_name='plus100',
            f=lambda col, gdf: col+100, 
            columns=['num'], 
            replace=False),
        ops.Filter(f=lambda gdf: gdf[gdf['num_plus100']>100]),
        ops.LambdaOp(
            op_name='plus50',
            f=lambda col, gdf: col+50,  
            replace=False
        )
    ]
)`

PR #195, opened 2020-08-05
[REVIEW] Enable add_cat_preprocess for external join

rjzamora:
This PR changes the `default_in`/`default_out` of `JoinExternal` to be `CAT`/`ALL`.  It also fixes a minor bug in `op_default_check`.

### DIFF
--- a/nvtabular/ops.py
+++ b/nvtabular/ops.py
@@ -831,7 +831,7 @@ class JoinExternal(TransformOperator):
         if the data is originally stored on disk.
     """
 
-    default_in = CAT
+    default_in = ALL
     default_out = ALL
 
     def __init__(
--- a/nvtabular/ops.py
+++ b/nvtabular/ops.py
@@ -831,7 +831,7 @@ class JoinExternal(TransformOperator):
         if the data is originally stored on disk.
     """
 
-    default_in = ALL
+    default_in = CAT
     default_out = ALL
 
     def __init__(
--- a/nvtabular/workflow.py
+++ b/nvtabular/workflow.py
@@ -30,6 +30,10 @@ from nvtabular.worker import clean_worker_cache
 LOG = logging.getLogger("nvtabular")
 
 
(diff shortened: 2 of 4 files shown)


nvtabular/workflow.py
```diff
@@ -103,7 +103,7 @@ def op_default_check(self, operators, default_in):
         if not type(operators) is list:
             operators = [operators]
         for op in operators:
-            if op.default_in not in default_in:
+            if op.default_in != default_in:
```
benfred, line 106: So I have some questions about this ```op_default_check``` method:

1) If we don't pass a list of operators (like ```operators``` is a single value), it seems like this method is sort of broken. It will convert the operator to a list on line 104, perform the check and then remove the operator from that list - but since the list isn't returned this has no effect. I think maybe we should return the list here so that the callers can use that instead?

2) It doesn't seem like this line handles the 'ALL' case at all ? Should it be something like 

```suggestion
            if op.default_in != default_in and op.default_in != ALL:
```
instead ?

nvtabular/ops.py
```diff
@@ -831,7 +831,7 @@ class JoinExternal(TransformOperator):
         if the data is originally stored on disk.
     """
 
-    default_in = ALL
+    default_in = CAT
```
benfred, line 834: Do we only want to be able to join on CAT columns?

nvtabular/workflow.py
rjzamora: Your right on both points.  I intended to make both these changes originally, but somehow lost track of my "plan" here.

nvtabular/ops.py
rjzamora: We can join CAT columns either way, since this op cannot really do anything by default (we need to know `on` explicitly). Mking this change only affects out ability to use `add_*_preprocess` - However, I completely agree that this *should* be `ALL`, so I'll make that change.

PR #196, opened 2020-08-05
[REVIEW] Fix multi-file output bug

rjzamora:
This PR adds test coverage for the output of a Workflow withoug any Ops.  In this case, the input and output datasets should be identical (after sorting).  It seems that there has been a bug in the `ThreadedWriter` (not sure how long), causing duplicate data to be written to multiple output files in some cases.  The only fix I could find was to explicitly copy the data before passing to `_write_table` (or adding to the thread queue).

### DIFF
--- a/nvtabular/io.py
+++ b/nvtabular/io.py
@@ -18,6 +18,7 @@ import collections
 import functools
 import json
 import logging
+import math
 import os
 import queue
 import threading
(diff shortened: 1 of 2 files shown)


nvtabular/io.py
```diff
@@ -298,18 +299,17 @@ def add_data(self, gdf):
...
-            cp.floor_divide(ind, (nrows // self.num_out_files), out=ind)
-
-        # Use `scatter_by_map` to produce contiguous group for each output file
+            cp.floor_divide(ind, math.ceil(nrows / self.num_out_files), out=ind)
         for x, group in enumerate(
             gdf.scatter_by_map(ind, map_size=self.num_out_files, keep_index=False)
         ):
             self.num_samples[x] += len(group)
+            # It seems that the `copy()` operations here are necessary
+            # (test_io.py::test_mulifile_parquet fails otherwise)...
```
benfred, line 308: Do we know why we have to copy? Do we have to copy for the single threaded version?

rjzamora: Yeah - The problem is for both threaded and non-threaded code paths.  I am more than a bit perplexed by this.

At first I thought it may have something to do with the switch to `scatter_by_map`.  However, the test fail when I go back to the same `iloc`-based logic we used before the switch.  I know that the `ParquetWriter.write_table` method operates on a `data_view`, but I don't quite understand why the referenced data is changing after the `write_table` call (but before the data actually gets written to disk).  Note that I confirmed that the correct data is passed to each `_write_table` call, so the problem has to be related to the referenced data changing after the call (but before the write)

benfred: Do you have a minimal reproduction of this issue? It seems like this might be something we need to raise with the cudf team

rjzamora: > Do you have a minimal reproduction of this issue? It seems like this might be something we need to raise with the cudf team

I'm going to try to do this now

benfred: Thanks!

 I'm a little worried about the copy causing issues with running out of GPU memory, but I think we should probably merge in the meantime

rjzamora: Raised issue [here](https://github.com/rapidsai/cudf/issues/5858)

PR #200, opened 2020-08-07
[REVIEW] Add partition-shuffle option to Dataset.to_ddf 

rjzamora:
Adds a `shuffle` argument to `Dataset.to_ddf` and `Dataset.to_iter`.  Does not shuffle data within partitions - **does** shuffle the order of partitions.


rjzamora:
cc @EvenOldridge (just for awareness - since this feature was requested by you offline)

@benfred - After this is in, I'd like to rename the "shuffle" options available in the `Workflow` to something like `{"per-chunk", "per-worker", "full", None}` (where we can use the feature in this PR to start with a shuffled partition order for anything other than `None`)

PR #201, opened 2020-08-07
[REVIEW] Update shuffle options and reorder partitions

rjzamora:
This is a follow-up to #200 (adding partition-shuffling to `Dataset.to_ddf`).

- This PR changes `Workflow.apply` to start with shuffled `dask_cudf.DataFrame` partitions whenever the user specifies a `shuffle` option (other than `None`).
- This PR changes the `shuffle` options to :`{"per-chunk", "per-worker", "full", None}`. The `apply` docstring has been updated to explain the options in detail.  Note that `shuffle="full"` is not yet supported  (it can be implemented in Dask pretty easily, but will be slow/fragile for systems with limited device memory).

### DIFF
--- a/nvtabular/io.py
+++ b/nvtabular/io.py
@@ -15,6 +15,7 @@
 #
 
 import collections
+import enum
 import functools
 import json
 import logging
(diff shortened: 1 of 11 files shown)


nvtabular/io.py
```diff
@@ -58,6 +58,18 @@
...
 
+def _check_shuffle_arg(shuffle):
+    if shuffle not in ("per-chunk", "per-worker", "full", None):
+        if shuffle is True:
+            shuffle = "per-worker"
+            warnings.warn('`shuffle=True` is deprecated. Using `shuffle="per-worker".`')
+        elif shuffle is False:
+            shuffle = None
+        elif shuffle == "full":
+            raise ValueError('`shuffle="full"` is not yet supported.')
```
benfred, line 69: Should we also add a catch-all here for when the user passes something else? Right now if you pass say ```shuffle=1``` it will I guess work the same as 'per-chunk' - but its not at all clear.  Maybe something like 
```suggestion
            raise ValueError('`shuffle="full"` is not yet supported.')
        else:
            raise ValueError(f"shuffle='{shuffle}' isn't a valid shuffling type")
```

rjzamora: Ah - My bad.  Thanks for catching this :)

rjzamora:
> And then have people pass shuffle=Shuffle.PER_WORKER instead. What do you think?

The C/C++ side of me likes it alot -  The python side of me is slightly worried... but I'm mostly on board :)

PR #203, opened 2020-08-08
[REVIEW] Change ZeroFill to FillMissing / Clip

### LINKED ISSUE #3: Decoupling ZeroFill from Relu op
Currently the ZeroFill op also implicitly does something akin to a Relu op i.e. replacing negative values with 0 via `z_gdf[z_gdf < 0] = 0`.  I think these two behaviours should be decoupled and Relu should be made into an explicit op.

```
class ZeroFill(TransformOperator):
    default_in = CONT
    default_out = CONT

    @annotate("ZeroFill_op", color="darkgreen", domain="nvt_python")
    def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None):
        cont_names = target_columns
        if not cont_names:
            return gdf
        z_gdf = gdf[cont_names].fillna(0)
        z_gdf.columns = [f"{col}_{self._id}" for col in z_gdf.columns]
        z_gdf[z_gdf < 0] = 0
        return z_gdf
```

benfred:
The ZeroFill operator was not only replacing missing values with 0, it
was also replacing negative values with 0. Since this can cause confusion
this change names ZeroFill to ReLu.

The 'FillMissing' operator already handles the case of replacing missing
values to 0 so we don't need a replacement.

Closes #3


alecgunny:
Might it make more sense to replace this functionality entirely with a more general clipping op? That way users can achieve this functionality with the combination of a `FillNa` and a `Clip` with the min set to 0 and the max to `None` (or whatever value we want to indicate no clipping on a given side). We could even make a special subclass for this functionality (without the filling) that just initializes the min value to 0. I guess from a conceptual standpoint, I don't necessarily see why the filling and clipping would be done in the same op, or for that matter why `ReLu` would indicate the filling part.

This is also more pedantic, but I think standard practice is to capitalize the "u" in "ReLu".

benfred:
> Might it make more sense to replace this functionality entirely with a more general clipping op? 

That makes sense - both from a functionality and naming perspective. I'll make that change later this week

benfred:
@alecgunny I've changed this from ReLu to Clip in that last commit - I think it makes more sense this way and can be used in more situations.  Thanks for the suggestion!

PR #206, opened 2020-08-11
[REVIEW] introduce controller class for fw specific use of nvt dataloader

jperez999:


jperez999:
rerun tests

jperez999:
rerun tests

examples/rossmann-store-sales-example.ipynb
```diff
@@ -514,19 +514,19 @@
    "outputs": [],
    "source": [
     "import torch\n",
-    "from nvtabular.torch_dataloader import AsyncTensorBatchDatasetItr, DLDataLoader\n",
+    "from nvtabular.torch_dataloader import TorchAsyncItr, DLDataLoader\n",
```
benfred, line 517: Should we use this for the vanilla pytorch version here as well as the fastai version?

nvtabular/torch_dataloader.py
```diff
@@ -222,7 +222,45 @@ def __del__(self):
         self.buff.stop()
 
 
-class AsyncTensorBatchDatasetItr(torch.utils.data.IterableDataset):
+class TorchAsyncItr(torch.utils.data.IterableDataset):
```
benfred, line 225: Why do we need this class in addition to the AsyncTensorBatchDatasetItr class? This class is a thin wrapper over AsyncTensorBatchDatasetItr which in turn is a thin wrapper over AsyncIterator - can we remove some of this indirection and have this use AsyncIterator directly?

jperez999: The idea is that the AsyncTensorBatchDatasetItr  provides the generic iterator and does the factory stuff... the initial reason for this was to remove the requirement of the torch.utils.data.iterabledataset from the base components of the nvt dataloading logic. this way you just create a wrapper and the tensor iterator in the factory and you can easily use the rest of the functionality while having support for target framework.

examples/rossmann-store-sales-example.ipynb
jperez999: you mean use both the torch.utils.dataloader and the DLDataloader...

benfred: We have two versions of using the torch_dataloader in this notebook now with this PR https://github.com/NVIDIA/NVTabular/pull/147 - one that uses the fastai framework for training and one that only uses pytorch. I think this PR is only updating the fastai version

nvtabular/torch_dataloader.py
benfred: I guess what I was trying to ask is whats the difference between AsyncIterator and AsyncTensorBatchDatasetItr now that you've removed the torch compatibility from AsyncTensorBatchDatasetItr. (like what do we need AsyncTensorBatchDatasetItr for anymore - it doesn't seem to implement any functionality ?)

jperez999: you are absolutely right I can just factor that out. Good catch!

jperez999: Really good... yeah this PR is useless. I will close.

PR #207, opened 2020-08-11
Use main branch as default for NVTabular

benfred:
We're updating the default branch for NVTabular to be 'main' instead of 'master'. This PR changes the docs to 
build for the main branch instead, and updates documentation links to go to the main branch instead.


jperez999:
rerun tests

PR #213, opened 2020-08-13
[REVIEW] Add a ColumnSimilarity op

benfred:
Add a ColumnSimilarity op, that calculates the similarity between two columns
using tf-idf, cosine or inner product as the distance metric. For each row,
this calculates the distance between the two columns by looking up features
for those columns in a sparse matrix, and then computing the distance
between the rows of the feature matrices.


benfred:
@gabrielspmoreira can you take a look at this too?

PR #215, opened 2020-08-14
[REVIEW] MultiGPU and iterator name shortening

jperez999:


nvtabular/torch_dataloader.py
```diff
@@ -196,10 +196,14 @@ def __init__(
         batch_size=1,
         shuffle=False,
         library=None,
+        devices=None,
     ):
-        itr = TensorBatchDatasetItrFactory().create(dataset, library, shuffle=shuffle)
+        self.dataset = dataset
+        self.library = library
+        self.shuffle = shuffle
+        self.devices = devices if devices else [0]
```
benfred, line 204: Should we default to CUDA_VISIBLE_DEVICES rather than [0] ?

jperez999: No in this case we want to go to zero... zero is the first available device... When you set cuda devices whatever you set as visible is the count. starting with zero.

Issue #218, opened 2020-08-17
[FEA] Asynchronous TensorFlow DataLoader


benfred:
Similar to https://github.com/NVIDIA/NVTabular/issues/81, we should add the ability for the TF dataloader to load chunks concurrently with training.

benfred:
Closed with #224

PR #224, opened 2020-08-18
[REVIEW] Creating dedicated `loader` submodule to build TF async dataloader

alecgunny:
Addressing #218


nvtabular/loader/backend.py
```diff
@@ -212,23 +211,18 @@ def __init__(
...
-                self.dataset,
-                self.library,
-                indices=indices.tolist(),
-                device=dev,
-                total_devs=self.devices,
-            )
-            t1 = threading.Thread(target=self.buff.load_chunks, args=(dev, itr))
-            t1.daemon = True
-            t1.start()
+            cp.random.shuffle(self.itrs[0].indices)
```
jperez999, line 215: I dont think this will work for the shuffling... we need to have the same index after randomized in all iterators. Here you only shuffling one... so all other iterators will still grab a chunk, from the ordered list, right?

nvtabular/loader/backend.py
```diff
@@ -349,72 +312,80 @@ def create_tensors(self, gdf, cat_names=None, cont_names=None, label_names=None)
...
-                dataset, shuffle=shuffle, device=device, total_devs=total_devs, **kwargs
-            )
-        else:
-            raise ValueError(target)
-
+def _validate_workflows(workflows, cat_names, cont_names, label_names):
+    assert all([isinstance(w, BaseWorkflow) for w in workflows])
+    # TODO: commenting out until it's clearer what the
+    # columns in workflow.columns_cts["final"]["ctx"] mean
+    # for workflow in workflows:
```
jperez999, line 319: here what you should be getting is a list of operator names (IDs) that then need to be looked up within the columns context for the columns prescribed to that operator.

nvtabular/loader/backend.py
```diff
@@ -349,72 +312,80 @@ def create_tensors(self, gdf, cat_names=None, cont_names=None, label_names=None)
...
+            dataset,
+            cat_names,
+            cont_names,
+            label_names,
+            batch_size,
+            shuffle,
+            workflows=None,
+            devices=None
+    ):
+        indices = cp.arange(dataset.to_ddf().npartitions)
```
jperez999, line 342: I think the shuffle should be done here. That way all the iterators will have the same order of indexes randomized or not.

nvtabular/loader/backend.py
alecgunny: Yeah this is someone I was meaning to talk to you about. I think this implementation should work fine since, at the moment, they all get passed the same `indices` object and so the in-place shuffle done by `cp.random.shuffle` shuffles the underlying object they're all pointing to (hence why I only shuffle the first `itrs`'s `indices`). That said, if the intention is to have `AsyncIterator`s ever instantiated by `TensorBatchDatasetItr` objects that *weren't* created in the `__init__` of a `DataLoader` object (and thus could have disparate `indices` attributes), then this would be problematic. I guess it depends on how much of this we think should just be internal.

nvtabular/loader/backend.py
alecgunny: Yeah so this is part of what I was hoping to understand about the index shuffle: does it need to be done at the epoch level, or can we just do it up front? Let's talk later today about what the partition index shuffle is doing, I'll admit my Dask understanding is pretty limited

nvtabular/loader/backend.py
alecgunny: Hmmm ok yeah I see the operator names, so how would I use those to map to expected output column names?

jperez999: I think you are looking for this https://github.com/NVIDIA/NVTabular/blob/main/nvtabular/workflow.py#L383 that will take the columns context object and pulls the actual correct columns.

nvtabular/loader/backend.py
jperez999: So actually doing it at the per epoch level is better. Introduces more randomness. So to that point we probably should not do it here.

nvtabular/loader/backend.py
alecgunny: Ok I guess that makes sense, so we need to be calling `Workflow.finalize()` and `Workflow.create_final_cols()` once a workflow is done be added to?

This is probably a discussion for elsewhere, but I wonder if it might make sense to implement `columns_ctx` as a `@property` and just create this on the fly whenever users need it, making it easier to inspect at intermediate steps during creation.

alecgunny:
Some pretty major overhauls in the last few commits. Chief among them:
- Tensor -> framework conversion _and_ batching are now done up front in the async thread. This is done because the batching in TF is nontrivial and so it pays to do it with a single tf.split op rather than slicing at each iteration
- Conversion from `cats, conts, labels` representation to framework-specific representations are done via a `_handle_tensors` method, which defaults to doing nothing but can be overridden in child classes as it is in the `KerasSequenceLoader`
- TF 2.3.0 excludes `Sequence` objects from being used as `validation_data` in `.fit`. As a result, I've implemented a custom validation callback to use when training Keras models. In the long run, this is probably a better scenario anyway, since a lot of recommender metrics depend on the input data and this makes implementing them in a custom validation loop easier anyway.

alecgunny:
@jperez999 this code includes a `parts_per_chunk` kwarg on the `DataLoader` class, are we cool with that? I can add it in to the criteo example notebook

examples/criteo-example.ipynb
```diff
@@ -255,27 +327,27 @@
...
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 15,
    "metadata": {},
    "outputs": [],
    "source": [
-    "train_data_itrs = TorchAsyncItr(train_data, batch_size=BATCH_SIZE, cats=CATEGORICAL_COLUMNS, conts=CONTINUOUS_COLUMNS, labels=LABEL_COLUMNS, num_parts=NUM_PARTS)\n",
-    "valid_data_itrs = TorchAsyncItr(valid_data, batch_size=BATCH_SIZE, cats=CATEGORICAL_COLUMNS, conts=CONTINUOUS_COLUMNS, labels=LABEL_COLUMNS, num_parts=NUM_PARTS)"
+    "train_data_itrs = TorchAsyncItr(train_data, batch_size=BATCH_SIZE, cats=CATEGORICAL_COLUMNS, conts=CONTINUOUS_COLUMNS, labels=LABEL_COLUMNS)# , devices=[1,2,3,4])\n",
```
benfred, line 344: I think this might be a bad merge with Julio's recent changes?

alecgunny: yeah I had already added a `parts_per_chunk` kwarg, Julio added a similar one just in case my PR didn't make it in in time. Happy to switch syntax or to adjust this example using the new kwarg.

benfred: yeah, we should have the other version ```num_parts=NUM_PARTS```. Julio added to make the integration tests work automatically (https://github.com/NVIDIA/NVTabular/pull/247).  

There are also some other minor changes with docstrings from the PT classes (https://github.com/NVIDIA/NVTabular/pull/242) and for handling the case where the async thread threw an exception causing a hang (https://github.com/NVIDIA/NVTabular/pull/246).  We should also get these merged back in.

examples/criteo-example.ipynb
```diff
@@ -81,15 +115,14 @@
...
-   "execution_count": null,
+   "execution_count": 4,
    "metadata": {},
    "outputs": [],
    "source": [
     "# define some information about where to get our data\n",
     "INPUT_DATA_DIR = os.environ.get('INPUT_DATA_DIR', '/raid/criteo/tests/crit_int_pq')\n",
     "OUTPUT_DATA_DIR = os.environ.get('OUTPUT_DATA_DIR', '/raid/criteo/tests/test_dask') # where we'll save our procesed data to\n",
     "BATCH_SIZE = int(os.environ.get('BATCH_SIZE', 800000))\n",
-    "NUM_PARTS = int(os.environ.get('NUM_PARTS', 2))\n",
```
benfred, line 92: this line needs re-added

alecgunny: not sure why the diff isn't showing it but if you look [here](https://github.com/alecgunny/NVTabular/blob/tfasync/examples/criteo-example.ipynb) you'll see I just replaced it with a `PARTS_PER_CHUNK` variable, which is what I use in `tests/integration/test_notebooks.py` as well

examples/rossmann-store-sales-example.ipynb
```diff
@@ -68,12 +85,11 @@
...
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [],
    "source": [
-    "DATA_DIR = os.environ.get('INPUT_DATA_DIR', './data')\n",
-    "OUTPUT_DATA_DIR = os.environ.get('INPUT_DATA_DIR', './data')\n",
+    "DATA_DIR = os.environ.get(\"OUTPUT_DATA_DIR\", \"./data\")\n",
```
benfred, line 92: This probably should be INPUT_DATA_DIR, since it's also used for getting reading the dataset.

Right now the unittests are failling because of this in tests/unit/test_notebooks.py::test_rossman_example 

```
  File "/tmp/pytest-of-jenkins/pytest-20/test_rossman_example0/notebook.py", line 59, in <module>
    proc.apply(train_dataset, record_stats=True, output_path=PREPROCESS_DIR_TRAIN, shuffle=nvt.io.Shuffle.PER_WORKER, out_files_per_proc=2)
  File "/var/jenkins_home/nvtabular/nvtabular/workflow.py", line 729, in apply
    num_io_threads=num_io_threads,
  File "/var/jenkins_home/nvtabular/nvtabular/workflow.py", line 829, in build_and_process_graph
    self.exec_phase(idx, record_stats=record_stats)
  File "/var/jenkins_home/nvtabular/nvtabular/workflow.py", line 612, in exec_phase
    self._aggregated_dask_transform(transforms)
  File "/var/jenkins_home/nvtabular/nvtabular/workflow.py", line 587, in _aggregated_dask_transform
    ddf = self.get_ddf()
  File "/var/jenkins_home/nvtabular/nvtabular/workflow.py", line 575, in get_ddf
    return self.ddf.to_ddf(columns=columns, shuffle=self._shuffle_parts)
  File "/var/jenkins_home/nvtabular/nvtabular/io.py", line 809, in to_ddf
    ddf = self.engine.to_ddf(columns=columns)
  File "/var/jenkins_home/nvtabular/nvtabular/io.py", line 1060, in to_ddf
    return dask_cudf.read_csv(self.paths, chunksize=self.part_size, **self.csv_kwargs)[
  File "/opt/conda/lib/python3.7/site-packages/dask_cudf-0.15.0a0+4964.g4b6b7c0fd.dirty-py3.7.egg/dask_cudf/io/csv.py", line 19, in read_csv
    return _internal_read_csv(path=path, chunksize=chunksize, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/dask_cudf-0.15.0a0+4964.g4b6b7c0fd.dirty-py3.7.egg/dask_cudf/io/csv.py", line 59, in _internal_read_csv
    meta = dask_reader(filenames[0], **kwargs)._meta
  File "/opt/conda/lib/python3.7/site-packages/dask/dataframe/io/csv.py", line 649, in read
    **kwargs,
  File "/opt/conda/lib/python3.7/site-packages/dask/dataframe/io/csv.py", line 479, in read_pandas
    **(storage_options or {}),
  File "/opt/conda/lib/python3.7/site-packages/dask/bytes/core.py", line 125, in read_bytes
    size = fs.info(path)["size"]
  File "/opt/conda/lib/python3.7/site-packages/fsspec/implementations/local.py", line 60, in info
    out = os.stat(path, follow_symlinks=False)
FileNotFoundError: [Errno 2] No such file or directory: '/var/jenkins_home/nvtabular/data/train.csv'
```



```suggestion
    "DATA_DIR = os.environ.get(\"INPUT_DATA_DIR\", \"./data\")\n",
```

alecgunny: Ok I think I'm confused then, because the preprocessing notebook uses INPUT_DATA_DIR to read the original data, but then places the preprocessed data with the extra features in OUTPUT_DATA_DIR, which is where `train.csv` should live. Unless we're not using the preprocessing notebook first in this test? When I run the preproc notebook followed by the example notebook locally, this works.

PR #228, opened 2020-08-20
Fix cosine similarity mutating input matrix

benfred:
The ColumnSimilarity op was mutating the input matrix on the gpu for the cosine
and tfidf distances. Fix and add a unittest to detect this. Also add the
ColumnSimilarity op to the generated sphinx docs.


benfred:
rerun tests

PR #238, opened 2020-08-24
[REVIEW] JoC WnD Outbrain model preprocessing with NVT notebook

rnyak:
This is restructured preprocessing and feature engineering pipeline of JoC WnD Outbrain model with NVTabular.

@benfred for viz.


examples/preproc3-NVT_E2E.ipynb
```diff
@@ -0,0 +1,890 @@
...
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "os.environ[\"CUDA_DEVICE_ORDER\"] = \"PCI_BUS_ID\"\n",
+    "os.environ[\"CUDA_VISIBLE_DEVICES\"] = \"3\""
```
benfred, line 11: I'd like to leave these out of notebooks, and just rely on them being set at the OS level (like 'CUDA_VISIBLE_DEVICES=3' won't work for many users that don't have a 4 gpu machine).

examples/preproc3-NVT_E2E.ipynb
```diff
@@ -0,0 +1,890 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "documents_categories_cudf = cudf.read_csv(\n",
+    "    DATA_BUCKET_FOLDER + 'documents_categories.csv',\n",
+    "    names=['document_id', 'category_id', 'confidence_level'],\n",
+    "    dtype={'document_id': 'int64','category_id': 'int64','confidence_level': 'float32'},\n",
```
benfred, line 89: Do we need to specify the names? Seems like just using the CSV header would work pretty well (like remove both names and skiprows=1 parameters)

examples/preproc3-NVT_E2E.ipynb
```diff
@@ -0,0 +1,890 @@
...
+    "HASH_BUCKET_SIZES = {\n",
+    "    'document_id': 300000,\n",
+    "    'ad_id': 250000,\n",
+    "    'document_id_promo': 100000,\n",
+    "    'source_id_promo': 4000,\n",
+    "    'source_id': 4000,\n",
+    "    'advertiser_id': 2500,\n",
+    "    'publisher_id_promo': 1000,\n",
+    "    'publisher_id': 1000,\n",
+    "}"
```
benfred, line 137: Do we want to use hashing instead of Categorify here?

examples/preproc3-NVT_E2E.ipynb
```diff
@@ -0,0 +1,890 @@
...
+    "\n",
+    "workflow.add_feature([\n",
+    "    Dropna(['geo_location', 'platform']),\n",
+    "    JoinGroupby(columns=['ad_id', 'source_id', 'document_id_promo', 'publisher_id', 'advertiser_id', 'campaign_id'], \n",
+    "        cont_names=['clicked'],stats=['sum','count'])])\n",
+    "\n",
+    "workflow.finalize()\n",
+    "\n",
+    "GPU_MEMORY_FRAC = 0.15\n",
+    "train_dataset = nvt.Dataset(train_df, engine='parquet',  part_mem_fraction=GPU_MEMORY_FRAC)\n",
```
benfred, line 225: if you're passing a dataframe, you don't need the engine / part_mem_fraction parameters 
```suggestion
    "train_dataset = nvt.Dataset(train_df)\n",
```

examples/preproc3-NVT_E2E.ipynb
rnyak: Yes we use HashBucket with these features.

examples/preproc3-NVT_E2E.ipynb
rnyak: Done.

examples/preproc3-NVT_E2E.ipynb
rnyak: Done.

rnyak:
@benfred The updated nb has NVT + WnD model TF 2.3.

alecgunny:
Hey so there's a few problems with the training part of the E2E notebook, some of them perf-related and some of them related to wrong functionality:
- Right now, the "wide" branch of the network doesn't multiply the numeric features by any parameters, it just adds them as-is. It's also missing a bias add. I have an implementation that should fix these and be more optimal than the Keras embedding layers [here](https://yagr.nvidia.com/agunny/keras-recommender-workflow/-/blob/v1/recommender_tools/layers/embedding.py#L139).
- The Keras DenseFeatures layer is terrible for GPU perf and should be avoided at all cost. Consider using the implementation I've built [here](https://github.com/NVIDIA/NVTabular/blob/a2c87edae6869de973fdb1330f1402ffd787e928/examples/tensorflow/layers.py#L167)
- Overall, the code could be drastially simplified and avoid using the Keras pre-baked version of the model. Consider taking a look at the implementation I have [here](https://yagr.nvidia.com/agunny/keras-recommender-workflow/-/blob/v1/examples/workflow/architectures/wide_and_deep.py#L35) (where the values in `channels` are just lists of the appropriate `feature_column`s, and the values in `inputs` are just the corresponding dicts of Keras tensors you would traditionally feed to the `DenseFeatures` layer.

kkranen:
@alecgunny 

> Right now, the "wide" branch of the network doesn't multiply the numeric features by any parameters. It's also missing a bias add.

I believe `linear_output = categorical_output_contrib + tf.keras.layers.Dense(1)(numeric_dense_tensor)` should handle parameter multiplication of dense numeric features in the Dense layer. Similarly, the Dense layer should also provide a bias.

> The Keras DenseFeatures layer is terrible for GPU perf and should be avoided at all cost

I'm currently working on a version of this flow that incorporates your embedding implementation. I've been seeing around a 20-25% speedup versus the original. The changes should be incorporated into this PR soon.

> Overall, the code could be drastially simplified and avoid using the Keras pre-baked version of the model.

The reason that the Keras model is used here is to allow for dual optimizers for the Wide and Deep portions respectively (Wide uses FTRL and Deep uses Adam or Adagrad). I agree that the code can be simplified here, there are just a few design decisions that need to be made to decide how to do it.

alecgunny:
@kkranen 
> I believe linear_output = categorical_output_contrib + tf.keras.layers.Dense(1)(numeric_dense_tensor) should handle parameter multiplication of dense numeric features in the Dense layer. Similarly, the Dense layer should also provide a bias.
Got it, missed that line. It might help to do this all in one place to make that a little more clear, but my bad on not catching that.

> The reason that the Keras model is used here is to allow for dual optimizers for the Wide and Deep portions respectively
Good point, forgot about that part of the model

alecgunny:
Also I mentioned this to Ronay yesterday but it's probably worth noting here: using a `Sequence` subclass is explicitly not allowed to be passed to `validation_data` and will cause errors or unexpected behavior in 2.3.0. See the updated Rossmann example for how to get around this: basically I've just made a custom `KerasSequenceValidater` callback that does the validation instead.

rnyak:
> Also I mentioned this to Ronay yesterday but it's probably worth noting here: using a `Sequence` subclass is explicitly not allowed to be passed to `validation_data` and will cause errors or unexpected behavior in 2.3.0. See the updated Rossmann example for how to get around this: basically I've just made a custom `KerasSequenceValidater` callback that does the validation instead.

@alecgunny Thanks! Yes, I actually already applied `KerasSequenceValidater`, but somehow I see my older notebook version here. might be a commit issue or something like that.

rnyak:
@benfred the notebook is longer because `ScalarDenseFeatures` layer and `DotProductInteraction` layer were added. Better to import these from a `layers` file as in `accelerating-tensorflow.ipynb`.

benfred (commented):
* We should add instructions on how to download the dataset
* Frequency thresholds should be constant 0 rather than per column
* The 6 CTR lambdaop operators can be replaced by 1 operator on 6 columns instead. Something like
```python
LambdaOp(op_name='ctr',
     f=lambda col, gdf: (col + 0.1 * 10)/(gdf[col.name.replace("_clicked_sum", "_count")]+10), 
     columns=["ad_id_clicked_sum", "source_id_clicked_sum", "document_id_promo_clicked_sum", "publisher_id_clicked_sum", "advertiser_id_clicked_sum", "campaign_id_clicked_sum"], 
     replace=False)
```
 * We should interleave the text and nvtabular operators rather than having a huge block of text followed by a huge block of code
 * Can we avoid large lists of columns with LogOp/Normalize/FillMissing - and just use the default continuous columns?
 * Can we save the stats from the workflow, so we can run training without doing preproc?

rnyak:
> * We should add instructions on how to download the dataset
> * Frequency thresholds should be constant 0 rather than per column
> * The 6 CTR lambdaop operators can be replaced by 1 operator on 6 columns instead. Something like
> 
> ```python
> LambdaOp(op_name='ctr',
>      f=lambda col, gdf: (col + 0.1 * 10)/(gdf[col.name.replace("_clicked_sum", "_count")]+10), 
>      columns=["ad_id_clicked_sum", "source_id_clicked_sum", "document_id_promo_clicked_sum", "publisher_id_clicked_sum", "advertiser_id_clicked_sum", "campaign_id_clicked_sum"], 
>      replace=False)
> ```
> 
> * We should interleave the text and nvtabular operators rather than having a huge block of text followed by a huge block of code
> * Can we avoid large lists of columns with LogOp/Normalize/FillMissing - and just use the default continuous columns?
> * Can we save the stats from the workflow, so we can run training without doing preproc?

@benfred the changes were incorporated and the updated notebook was committed.

rnyak:
@benfred updated the WnD TF model section by importing `layers` from utils. fyi.

alecgunny:
@rnyak there is also a `ScalarLinearFeatures` layer that might be worth using for the wide part of the model.

For consistency with other examples in the markdown portions, I would recommend using code formatting when referring to python objects e.g. `tf.feature_column`, `numeric_column`, etc.

benfred (approved):

PR #240, opened 2020-08-24
Fix performance bug in to_ddf for shuffle=True

rjzamora:
After #201, when writing shuffled output data, the partitions of the original ddf will be reordered.  The approach introduced in #200 (using a `to_delayed` call) only required a few lines of code, but dramatically increases the size of the Dask graph for large datasets.  This PR removes the problematic approach.  Instead of re-ordering delayed partitions, we just construct a new high-level graph to produce a new Dask DataFrame with the new keys mapping to random indices of the original dataframe.

### DIFF
--- a/nvtabular/io.py
+++ b/nvtabular/io.py
@@ -810,10 +810,23 @@ class Dataset:
 
         # Shuffle the partitions of ddf (optional)
         if shuffle and ddf.npartitions > 1:
-            parts = ddf.to_delayed()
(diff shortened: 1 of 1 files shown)


benfred:
rerun tests

rjzamora:
rerun tests

benfred:
rerun tests

rjzamora:
Thanks @benfred !

PR #248, opened 2020-08-26
Add fold_groups option to GroupbyStatistics and add TargetEncoding

### LINKED ISSUE #115: [OP] TargetEncoding
TargetEncoding is a common feature engineering for categorical variables in tabular datasets. In RecSys2020, it was one of the most used techniques in our pipeline. It would be great if the target column is a parameter in the OP. 

We just consider the case of a numerical target or binary classification problem. For each categorical value, the mean of the target is calculated and a new feature (column) is created with the means of the categorical value. To prevent overfitting, to additional logics are applied:
1. Training CrossValidation: To prevent overfitting in the training dataset, a cross validation strategy is applied to the training set. The training set is splitted in k folds and the means for the i-th fold is calculated with all reamining folds. The fold assignmed is randomly. The validation/test set uses the full training dataset for calculating the mean.
2. Smoothing: To prevent overfitting for low cardinality categories, the means a smoothed with the overall mean of the target variable.

Function:
TE = ((mean_cat*count_cat)+(mean_global*p_smooth)) / (count_cat+p_smooth)

count_cat := count of the categorical value
mean_cat := mean of target value for the categorical value 
mean_global := mean of the target value in the dataset
p_smooth := smoothing factor

The OP should support multi-column Target Encoding. For Example, if the dataset contains Language and Gender, we can target-encode the combination pairs of ['Language', 'Gender']. The number of columns should be not limited to 2.
The OP has two parameters: 
* kfold := how many folds to apply on the training dataset 
* p_smooth := smoothing factor

My idea would be to create an op and define GroupByMoments as a statistic.
GroupByMoments has to be extended to support kfold operation.
The OP needs to differentiate between applying to the trainingset (using kfold) and validation/testset (using all folds)

**Additional context**
An example implementation can be found here:
https://github.com/rapidsai/recsysChallenge2020/blob/master/Baselines/DASK-XGB/Demo-e2e-multi-CPU.ipynb : MTE_one_shot
https://github.com/rapidsai/recsysChallenge2020/blob/master/final_models/XGB1334/XGB1334-submit-singleGPU.ipynb : target_encode_cudf_v3

rjzamora:
Closes #115 

This PR icludes a possible implementation of the `TargetEncoding` operator described in #115.  In order to leverage the existing functionality in `GroupbyStatistics`, this PR also adds a `fold_groups` option (and other supporting options) to that `StatOperator`.

The primary difference between this solution and [the one implemented for RecSys2020](https://github.com/rapidsai/recsysChallenge2020/blob/master/Baselines/DASK-XGB/Demo-e2e-multi-CPU.ipynb) is that the "training-set" vs "valid/test-set" behavior  is inferred at runtime (and not explicitly specified by the user).  That is, The `GroupbyStatistics` operation will now add a new "__fold__" column to the underlying `ddf` when the user asks for kfold sampling.  Therefore, the decision  to perform a fold-based target encoding comes down to the existence of the "__fold__" column... If this column does not exist, we use the "fold-free" statistics for encoding.

**TODO**:

- [x] Confirm general approach is acceptable
- [x] Validate correctness for a real example
- [ ] Improve performance with smarter "caching"


nvtabular/ops.py
```diff
@@ -880,6 +935,187 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None
...
+    in tabular datasets. We just consider the case of a numerical target or
+    binary classification problem. For each categorical value, the mean of
+    the target is calculated, and a new feature (column) is created with the
+    means of the categorical value. To prevent overfitting, to additional
+    logics are applied:
+
+        1. Training Cross Validation: To prevent overfitting in the training
+        dataset, a cross validation strategy is applied to the training set.
+        The training set is splitted in k folds and the means for the i-th
+        fold is calculated with all reamining folds. The fold assignmed is
```
benfred, line 950:
```suggestion
        fold is calculated with all remaining folds. The fold assigned is
```

nvtabular/ops.py
```diff
@@ -880,6 +941,196 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None
...
+        out_dtype=None,
+        preprocessing=True,
+        replace=False,
+        tree_width=None,
+        cat_cache="host",
+        out_path=None,
+        on_host=True,
+        name_sep="_",
+        stat_name=None,
+        drop_folds=True,
```
bschifferer, line 1021: The parameter are missing in the doc string.

rjzamora:
> I ran TargetEncoding with different configurations and the calculation is as expected.

Thanks @bschifferer !!

Should we leave performance tuning (caching, etc) for a future PR?

bschifferer:
> > I ran TargetEncoding with different configurations and the calculation is as expected.
> 
> Thanks @bschifferer !!
> 
> Should we leave performance tuning (caching, etc) for a future PR?

I haven't measured performance, yet. I think I am happy with the current PR. I would be fine using this version and create a new ticket for performance improvement.

benfred:
rerun tests

PR #259, opened 2020-09-01
[REVIEW] Rossmann preproc with cuDF

rnyak:
Still working on a workaround for couple of features that require ffill() and bfill() methods, that cudf does not currently support.


rnyak:
@benfred This is the final version. One thing to be aware of is that, the `cut` value to create `valid` set is different in cudf version than pandas version. This does not affect the valid acc, instead I got RMSE something around 0.17.

jperez999:
rerun tests

bschifferer:
Some comments (still review in progress):
* saving train, valid, test to *.csv in preproc and loading in example notebooks corrupt the column format. Trend and Trend_DE columns contain some strings afterwards. If I used .parquet, it does not corrupt the columns. I assume that some string contains "," as a character and then the columns get shifted
* column "SchoolHoliday" is from datatype object when I load the train dataframe in example.ipynb. We cannot operate mean operation on a column with object type. Therefore, the NVTabular workflow breaks. We need to convert it to a numeric datatype (e.g. np.int32 with df['SchoolHoliday'].astype(np.int32)), first (maybe only an issue related to use parquet)
* What is the RMSE in the notebook, I see PyTorch uses RMSPE or MSE?
* Using .parquet and np.int32, I get following numbers for PyTorch:
Epoch 24. Train loss: 0.0743. Train RMSPE: 0.3305. Valid loss: 0.0770. Valid RMSPE: 0.3391.
* Using .parquet and np.int32, I get following numbers for FastAI:
24 | 0.059864 | 0.031545 | 0.195930 | 00:02

bschifferer:
Final comments:
* Week calculation are a little bit different than pandas version. If you do googletrend['Week'].value_counts() after cell 7, then you see that week=52 has 56 values instead of 42. If you compare cudf/pandas in detail, the difference is may related to the day of a week (does the calendar week start with sunday or monday). **I do not think this needs to be changed**, as the model is trained on the data and will learn the differences

* Splitting training/test data: The training dataset is from (Timestamp('2013-01-01 00:00:00'), to Timestamp('2015-06-19 00:00:00') in pandas and (numpy.datetime64('2013-01-01T00:00:00.000000000'), numpy.datetime64('2015-06-18T00:00:00.000000000') in cuDF version. The number of rows differs by 917. This is related how the index of the dataframes are set. **I do not think this needs to be changed**

rnyak:
> Some comments (still review in progress):
> 
> * saving train, valid, test to *.csv in preproc and loading in example notebooks corrupt the column format. Trend and Trend_DE columns contain some strings afterwards. If I used .parquet, it does not corrupt the columns. I assume that some string contains "," as a character and then the columns get shifted
> * column "SchoolHoliday" is from datatype object when I load the train dataframe in example.ipynb. We cannot operate mean operation on a column with object type. Therefore, the NVTabular workflow breaks. We need to convert it to a numeric datatype (e.g. np.int32 with df['SchoolHoliday'].astype(np.int32)), first (maybe only an issue related to use parquet)
> * What is the RMSE in the notebook, I see PyTorch uses RMSPE or MSE?
> * Using .parquet and np.int32, I get following numbers for PyTorch:
>   Epoch 24. Train loss: 0.0743. Train RMSPE: 0.3305. Valid loss: 0.0770. Valid RMSPE: 0.3391.
> * Using .parquet and np.int32, I get following numbers for FastAI:
>   24 | 0.059864 | 0.031545 | 0.195930 | 00:02

thanks for these comments:

- I cannot reproduce` Trend and Trend_DE columns contain some strings afterwards` issue when I read csv files with cudf. Both column dtypes are `dtype: int64`.  

-  * column "SchoolHoliday"  dtype is  also `int64` (see below). I could run NVT workflow and TF and PyT models without issues.
```train.SchoolHoliday
0         1
1         1
2         1
3         1
4         1
         ..
802162    0
802163    0
802164    0
802165    0
802166    0
Name: SchoolHoliday, Length: 802167, dtype: int64
```

rnyak:
> Final comments:
> 
> * Week calculation are a little bit different than pandas version. If you do googletrend['Week'].value_counts() after cell 7, then you see that week=52 has 56 values instead of 42. If you compare cudf/pandas in detail, the difference is may related to the day of a week (does the calendar week start with sunday or monday). **I do not think this needs to be changed**, as the model is trained on the data and will learn the differences
> * Splitting training/test data: The training dataset is from (Timestamp('2013-01-01 00:00:00'), to Timestamp('2015-06-19 00:00:00') in pandas and (numpy.datetime64('2013-01-01T00:00:00.000000000'), numpy.datetime64('2015-06-18T00:00:00.000000000') in cuDF version. The number of rows differs by 917. This is related how the index of the dataframes are set. **I do not think this needs to be changed**

- I will look into `Week` column, and compare with pandas. Thanks for pointing that out.
- Yes correct, cudf prepoc pipelines does not have exactly same splits with pandas.

benfred (commented):
* notebook should check for existence of datafile and only download if necessary
* should add descriptions on what’s being done and why. Should reference fastai feature engineering and preproc in order to start in the same place as fastai etc.
* use gdf/pdf to distinguish and pandas and non-pandas dataframes (with bfill)
* There is a warning raised on cell [32]:
```
 = merge(train_df, df, ['Store', 'Date'], right_on=['Store', 'Date'])
test_df = merge(test_df, df, ['Store', 'Date'], right_on=['Store', 'Date'])
”

a warning is raised: “/opt/conda/envs/rapids/lib/python3.7/site-packages/cudf/core/join/join.py:368: UserWarning: can't safely cast column from right with type datetime64[ns] to datetime64[s], upcasting to datetime64[ns]
  "right", dtype_r, dtype_l, libcudf_join_type”
```
* this notebook should replace the existing rossmann preproc notebook, rather than be in addition to it

examples/rossmann-store-sales-preproc.ipynb
```diff
@@ -49,44 +55,35 @@
...
-      "googletrend.csv        state_names.csv\t test.csv\n",
-      "rossmann.tgz\t       store.csv\t train.csv\n",
-      "sample_submission.csv  store_states.csv  weather.csv\n"
-     ]
-    }
-   ],
+   "outputs": [],
+   "source": [
+    "isFile = os.path.isdir(INPUT_DATA_DIR)  \n",
+    "if isFile ==False:\n",
```
benfred, line 76: things like ```if isFile == False``` should be ```if not isFile```.

I think here though we should check the existence of the file, rather than the existence of the INPUT_DATA_DIR
```suggestion
    "if not os.path.isfile(os.path.join(INPUT_DATA_DIR, "rossmann.tgz")):\n",
```

examples/rossmann-store-sales-preproc.ipynb
```diff
@@ -49,44 +55,35 @@
...
+    "    ! mkdir -p $INPUT_DATA_DIR\n",
+    "    ! wget -O $INPUT_DATA_DIR/rossmann.tgz http://files.fast.ai/part2/lesson14/rossmann.tgz\n",
+    "    ! cd $INPUT_DATA_DIR && tar -xzf rossmann.tgz && ls"
```
benfred, line 79: I think mixing up bash/python like this is going to be problematic - there is a potential for the python INPUT_DATA_DIR variable to be different from the $INPUT_DATA_DIR environment variable (like if the env variable isn't set we provide a default path). Can we do all of this in python instead?

```suggestion
    "    os.makedirs(INPUT_DATA_DIR, exist_ok=True)\n",
    "    os.system(f'wget -O {INPUT_DATA_DIR}/rossmann.tgz http://files.fast.ai/part2/lesson14/rossmann.tgz')\n",
    "    os.system(f'tar -xzf {INPUT_DATA_DIR}/rossmann.tgz')\n"
```

benfred (approved):

PR #262, opened 2020-09-01
[WIP] Re-synchronizing data loaders

alecgunny:
Hacking around the `ChunkBuffer` class to return synchronous functionality to data loaders in light of performance regressions from asynchronous loading. Ideally, some of this functionality can be retained long-term to make asychronous data loading an `__init__` kwarg on the `DataLoader` class, in case users don't want it for some reason


nvtabular/loader/backend.py
```diff
@@ -142,9 +177,13 @@ def load_chunks(self, dev, dataloader):
                         spill = workflow.apply_ops(spill)
                     spill = dataloader._create_tensors(spill)
                     spill = dataloader._handle_tensors(*spill)
-                    self.put([spill])
+                    # ADDING FOR SYNC BEHAVIOR
+                    # self.put([spill])
+                    yield spill
```
benfred, line 182: Should this be ```yield [spill]``` instead to match previous behaviour?

nvtabular/loader/backend.py
```diff
@@ -55,16 +74,28 @@ def __init__(
         self.q_out = queue.Queue(qsize)
         self._stop_event = threading.Event()
 
+        # adding for sync behavior
+        self._itr = None
+        self._iterating = False
```
benfred, line 79: Can we just use ```self._itr is None``` as the check to see if we're iterating ?

nvtabular/loader/backend.py
alecgunny: ah good catch yes it should

nvtabular/loader/backend.py
alecgunny: Yeah that actually might make more sense I'll give that a test

jperez999:
rerun tests

alecgunny:
@jperez999 do these torch test failures help elucidate what might be going wrong at all?

jperez999 (approved):

PR #263, opened 2020-09-01
[REVIEW] Enable multiple column groups in TargetEncoding op

### LINKED ISSUE #261: [BUG]/[FE] TargetEncoding for multiple columns (separatly)
**Describe the bug**
TargetEncoding (TE) works well to encode one single column or multiple column at once. How to apply TE to multiple single columns? TE will encode them together 

```python
import nvtabular as nvt
import nvtabular.ops as ops

import numpy as np
import pandas as pd

import cudf

cat_1 = np.asarray(['baaaa']*12 + ['bvvvv']*10 + ['ddddc']*10)
cat_2 = np.asarray(['baaaa']*6 + ['bbaaa']*3 + ['bcaaa']*3 + ['bvvvv']*5 + ['bavvv']*5 + ['ddddc']*10)
num_1 = np.asarray([1,1,2,2,2,1,1,5,4,4,4,4, 1,2,3,4,5,6,7,8,9,10, 1,2,3,4,5,6,7,8,9,10])
pdf_1 = pd.DataFrame({'cat': cat_1, 'cat2': cat_2, 'num': num_1})

pdf_1.to_parquet('pdf_1.parquet')

proc = nvt.Workflow(
    cat_names=['cat', 'cat2'],
    cont_names=['num'],
    label_name=[]
)
proc.add_preprocess(
    ops.TargetEncoding(
        cat_group=['cat', 'cat2'],
        cont_target='num',
        kfold=1,
        p_smooth=5
    )
)

import glob

input_files = glob.glob('./pdf_1.parquet')

train_ds_iterator = nvt.Dataset(input_files, engine='parquet')

proc.apply(train_ds_iterator, 
           apply_offline=True, 
           record_stats=True, 
           shuffle=False, 
           output_path='./')

cudf.read_parquet('part.0.parquet')`

num	cat	cat2	TE_cat_cat2_num
0	1	baaaa	baaaa	2.821023
1	1	baaaa	baaaa	2.821023
2	2	baaaa	baaaa	2.821023
3	2	baaaa	baaaa	2.821023
4	2	baaaa	baaaa	2.821023
5	1	baaaa	baaaa	2.821023
6	1	baaaa	bbaaa	4.003906
7	5	baaaa	bbaaa	4.003906
8	4	baaaa	bbaaa	4.003906
9	4	baaaa	bcaaa	4.253906
10	4	baaaa	bcaaa	4.253906
11	4	baaaa	bcaaa	4.253906
12	1	bvvvv	bvvvv	3.703125
13	2	bvvvv	bvvvv	3.703125
14	3	bvvvv	bvvvv	3.703125
15	4	bvvvv	bvvvv	3.703125
16	5	bvvvv	bvvvv	3.703125
17	6	bvvvv	bavvv	6.203125
18	7	bvvvv	bavvv	6.203125
19	8	bvvvv	bavvv	6.203125
20	9	bvvvv	bavvv	6.203125
21	10	bvvvv	bavvv	6.203125
22	1	ddddc	ddddc	5.135417
23	2	ddddc	ddddc	5.135417
24	3	ddddc	ddddc	5.135417
25	4	ddddc	ddddc	5.135417
26	5	ddddc	ddddc	5.135417
27	6	ddddc	ddddc	5.135417
28	7	ddddc	ddddc	5.135417
29	8	ddddc	ddddc	5.135417
30	9	ddddc	ddddc	5.135417
31	10	ddddc	ddddc	5.135417
```

**Expected behavior**
I want to have the option to encode multiple columns via TE.

**Additional context**
I think, we should change the  cat_group parameter to:
 cat_group=['cat', 'cat2', ['cat', 'cat2]]

Then it will encode cat, cat2 and cat_cat2 together

rjzamora:
Closes #261 

This enables the TargetEncoding op to accomodate multiple column groups at once.  All coolumn groups must use the same continuous target column (we can introduce a follow-up PR to loosen this requirement if/when necessary).

Since there can now be multiple categorical-column groups, multi-column groups must be specified within two layers of square brackets.  As mentioned in the docstring:

```
    cat_groups : list of column-groups
        Columns, or column groups, to target encode. A single encoding
        will include multiple categorical columns if the column names are
        enclosed within two layers of square brackets.  For example,
        `["a", "b"]` means "a" and "b" will be separately encoded, while
        `[["a", "b"]]` means they will be encoded as a single group.
        Note that the same column can be used for multiple encodings.
        For example, `["a", ["a", "b"]]` is valid.
```


jperez999:
rerun tests

jperez999:
rerun tests

rjzamora:
@bschifferer - Let me know if these changes address your needs.  Sorry this was missing from the original PR!

rjzamora:
*Update*:  After running through some performance experiments, I am noticing that we may need to avoid leaving the `"__fold__"` column in `ddf` after returning from `GroupbyStatistics`...  In #248 , we changed the Workflow to update the underlying `ddf` after each statistics operation.  This can lead to more memory pressure than I originally anticipated.  The scheduler/workers will no longer "forget" about the ddf transformations performed during the statistics-gathering step, but now try to persist the changes in device memory.  For large datasets, this is only possible with device-to-host spilling (which can be slow)...  We may want to make the fold indices deterministic, and just add the `"__fold__"` column again during each TargetEncoding transform task. 
cc @benfred

benfred:
> _Update_: After running through some performance experiments, I am noticing that we may need to avoid leaving the `"__fold__"` column in `ddf` after returning from `GroupbyStatistics`... In #248 , we changed the Workflow to update the underlying `ddf` after each statistics operation. This can lead to more memory pressure than I originally anticipated. The scheduler/workers will no longer "forget" about the ddf transformations performed during the statistics-gathering step, but now try to persist the changes in device memory. For large datasets, this is only possible with device-to-host spilling (which can be slow)... We may want to make the fold indices deterministic, and just add the `"__fold__"` column again during each TargetEncoding transform task.
> cc @benfred

Thanks for the heads up!  I could see this causing issues with larger datasets. Are there any other transforms like this happening in other ops?

rjzamora:
> Thanks for the heads up! I could see this causing issues with larger datasets. Are there any other transforms like this happening in other ops?

TargetEncoding is the only op that requires its statistics-op dependency to modify the ddf.  Which is why I would like to roll back this behaviour if possible...

I am currently experimenting with a slightly different approach: Passing the partition index to every transform op, so that a deterministic fold mapping can be reconstructed at both statistics-gathering and transform time.  Unfortunately, this means we will need to plumb a `partition_index` argument for all the `apply_op` and `op_logic` definitions.  We can avoid this if we are okay with the same fold asignment being the same in every partition (which actually seems to be the case now anyway), but that seems like a rough limitation to me.  Also, since the `"__fold__"` column will no longer persist after GroupbyStatistics, we will need an additional argument for the user to specify if cross-validation should be used at transform time (e.g. `training_data=True`)

rjzamora:
**Update 2**:  I didn't find it to be much of an advantage to avoid updating ddf in `GroupbyStatistics`, and adding the partition index to all transofrmation options was getting a little "messy" (in my opinion).   For reference, I did try it out [in this commit](https://github.com/rjzamora/NVTabular/commit/a85c8ed14a7163a584452bfc091185fd46dc2dcd).  Given what I learned, I made a few different optimizations:

- Added `target_mean` argument to allow users to specify the global mean of the target continuous column (if it is already known).  It turns out that the dependency on the mean is the primary cause of the memory-pressure issues I was seeing before (the addition of a `"__fold__"` column is less important than I had thought).
- Modified `_write_gb_stats` to avoid any concatenation, and to avoid host-to-device transfer when the `on_host=True`.  More-specifically, we now loop through the incomming DataFrame's and append them to the same file with cudf or pyarrow, depending on the format of the DFs.  We no longer concatenate and convert to cudf, because we do not need to sort for the groupby statistics (just the uniques used in Categorify).  If, in the future, we enable the same `GroupbyStatistics` instance to cover both `Categorify` and other transformations, we will need to ensure that the data is sorted during/before the `Categorify` encoding step (if/when the `searchsorted` method is used).

rjzamora:
@benfred - I'm not sure I will have time to add any more optimizations here before 0.2 release.  I'll plan to merge this once the PR has your blessing.

nvtabular/ops/categorify.py
```diff
@@ -457,24 +460,43 @@ def _get_aggregation_type(col):
...
+                    pwriter = pq.ParquetWriter(path, pa_schema, compression=None)
+                pwriter.write_table(pa_table)
+            else:
+                # Use CuDF
+                df.reset_index(drop=True, inplace=True)
+                pwriter.write_table(df)
+            n_writes += 1
+
+    # Close writer and return path
+    pwriter.close()
```
benfred, line 494: Maybe we should move the ```pwriter.close()``` to after the ```if n_writes == 0``` check?  When there is no data written out, and on_host = True - I believe this line will throw a ```AttributeError: 'NoneType' object has no attribute 'close'``` exception

benfred (approved):
LGTM -

PR #265, opened 2020-09-02
Fix notebook location of device_mem_size

benfred:
Notebooks were failing because device_mem_size moved, fix.


albert17 (approved):
I just checked and it works.

Issue #23, opened 2020-06-04
[FEA] Add progress bars


benfred:
<a href="https://github.com/benfred"><img src="https://avatars2.githubusercontent.com/u/69536?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [benfred](https://github.com/benfred)**
_Monday May 04, 2020 at 19:08 GMT_
_Originally opened as https://github.com/rapidsai/recsys/issues/89_

----

We should default displaying progress information (using tqdm) so that users can view progress when running nvtabular

alecgunny:
Can we skyrocket this priority? Waiting for things to run for > 2 minutes with no feedback is a terrifying thing, let alone potentially 20 minutes +

PR #270, opened 2020-09-04
[REVIEW] Multi-GPU Notebook Improvements

rjzamora:
PR Includes a variety of changes intended to address feedback on `multi-gpu_dask.ipynb`

**Notes**:
- Added instructions for accessing the dask diagnostics dashboard.  The dashboard is likely the best way to visualize execution progress for NVTabular preprocessing.
- Added notes about device-host memory spilling
- Modified dataset-creation code to use dask (making it faster, and making it possible to visualize progress with the dask dashboard)
- Started 

**TODO**:
- [x] Multigpu - step 2 - remove … at end of description of “infiniband-based inter-process communication via UCX…”
- [x] Multigpu - step 4  remove … at end of description of “is that you are free to operate on it using a familiar cudf/pandas API…”
- [x] Multigpu - step 4  remove … at end of description of “we will start with a simpler Dataset definition for this example…”
- [x] Multigpu - step 5 remove … at end of description of “ we can use the NVTabular API as usual…”
- [x] Multigpu - step 6  remove … at end of description of “data to confirm that everything is looking good…”
- [x] Mutligpu - step 7 multiple removals of … in “aggregation, you could do somthing like this…” and “directly to read back our data…”
- [ ] Multigpu.html: Add hyperlink on RAPIDS to refer to https://rapids.ai/dask.html page.
- [x] Documentation:  “This step should be very inexpensive”, not sure `should` is written italic on purpose, but better to write it non-italic.
- [x] Functionality: Even though I set the `visible_devices = "2,3"`, the toy creation cell was running in the GPU:0. To avoid that I had to explicitly add `os.environ["CUDA_VISIBLE_DEVICES"] = "2, 3"` script at the beginning of the notebook in a conda env.
- [x] It is not clear to me why I get the same `npartition` when I set `part_size` param to "1GB"  and “2GB”, respectively, for a 3.9GB parquet file, I get 8 and 4 partitions. I was expecting `4` and `2` partitions instead. **The data may be 3.9GB in parquet format, but the corresponding cudf.DataFrame is usually much larger. Also, the `part_size` is an upper bound on the desired partition size. Therefore, ~6-8 partitions for a `part_size="1GB"` makes sense.  Notes on partitioning have been added to the notebook.**
- [x] Documentation- Notebook- Step7: `For example, if you want to convert the entire dataset into a groupby aggregation, you could do somthing like this…` . There is a typo in this sentence.
- [ ] Documentation - link format is wrong in Step 2 (which is provided by the RAPIDS `dask_cuda <https://github.com/rapidsai/dask-cuda>`__ library).
- [x] Documentation - step 2: what does memory spill mean? Does it need more explanation? “Spill host mem to disk near this limit”
- [x] Documentation - is it normal to write Dask upper case and dask_cudf lower case?
- [x] Documentation/Notebook - Can we time Step 3 (generating toy dataset)? I run it on NGC (with a virtual disk and it takes a long time, not sure if it is related to NGC)
- [x] Documentation - Step 6: “The parallel efficiency definitely drops at 8 GPUs.” sounds too negative - first, I thought that it will require more time as with 2 GPUs. I think the message is that it does not scale linearly (meaning 2 vs 8 GPUs are 4x more resources, but time drops only by factor 1.x)
- [x] Documentation - Notes on shuffling: “(with a an additional FULL optoin planned for a future release)” - spelling mistake of option
- [x] Documentation - Empty cell in the end
- [ ] We really, really, really, really need progress bars on preprocessing


rjzamora:
cc @rnyak @bschifferer @alecgunny (source of initial review comments)
cc @benfred (for viz)

Note that I did not add any progress bars to the notebook.  However, I did add documentation on the default diagnostics dashboard.  Hopefully this will provide the kind of visual feedback needed for long-running processes?  I originally tried using an actual progress bar, but had a lot of trouble getting it to work and/or display meaningful information.

benfred (approved):

benfred:
Lets track the progress bar issue separately from the rest (#23) -

PR #271, opened 2020-09-04
[REVIEW] Update ops documentation

benfred:
* remove preprocessing as a parameter for all ops
* remove 'although you can directly call methods ...' text
* add columns param description for difference lag
* add examples to docstrings for Clip/Filter ops
* Add parameters to Dropna/normalize/clip operators docstring
* Fix punctuation in operator docstrings


jperez999 (approved):
good bye preprocessing kwarg!!!

PR #288, opened 2020-09-09
pytorch model to repo

jperez999:
moves pytorch generic model code into repo


nvtabular/framework_utils/pytorch/models.py
```diff
@@ -0,0 +1,80 @@
...
+        if self.max_output:
+            x = self.max_output * torch.sigmoid(x)
+        x = x.view(-1)
+        return x
+
+def rmspe_func(y_pred, y):
+    "Return y_pred and y to non-log space and compute RMSPE"
+    y_pred, y = torch.exp(y_pred) - 1, torch.exp(y) - 1
+    pct_var = (y_pred - y) / y
+    return (pct_var**2).mean().pow(0.5)
```
alecgunny, line 60: Is this function necessarily something we need to include in nvtabular (and is a `models` module the best place for it)? If it has really common usage in recommendation problems, and if there's something about the implementation that is nontrivial, I'm all for it. But if this is just because it gets used in the Rossmann example, that only doesn't seem like a good reason to offer a pre-baked version of it

benfred (commented):
Should we store this in nvtabular/framework_utils/torch instead of nvtabular/framework_utils/pytorch?  Making this change would be consistent with how pytorch organizes their code (like we do `import torch` instead of `import pytorch`)

nvtabular/framework_utils/pytorch/models.py
jperez999: So maybe it belongs in an utils file in pytorch but I felt it belonged here because of the usage of torch library in line 58 
`y_pred, y = torch.exp(y_pred) - 1, torch.exp(y) - 1` But let me know if that should be removed.

jperez999:
yeah good call will change to torch

nvtabular/framework_utils/pytorch/models.py
alecgunny: I guess I'm not contesting whether it belongs in the `torch` section of NVTabular, but whether it's NVTabular's job to implement this function at all. I think it's worth having a broader discussion about what sorts of DL-side things we ought to be offering and what are best left as one-offs

jperez999: oh ok... yeah I will leave it in here for now because it cleans up the notebook a bit. Once we have decided one way or the other, I will remove if necessary.

benfred: I think we should probably just leave this function in the notebook. If we start to need this function in more than one spot we can move it out - but this seems very specific to Rossmann (converting back from log space, RMSPE at all etc).

examples/rossmann-store-sales-example.ipynb
```diff
@@ -55,21 +55,22 @@
...
       "\n",
-      "For more information about alternatives visit: ('https://numba.pydata.org/numba-doc/latest/cuda/overview.html', '#cudatoolkit-lookup')\n",
+      "For more information about alternatives visit: ('http://numba.pydata.org/numba-doc/latest/cuda/overview.html', '#cudatoolkit-lookup')\u001b[0m\n",
       "  warnings.warn(errors.NumbaWarning(msg))\n"
      ]
     }
    ],
    "source": [
     "import os\n",
+    "os.environ[\"CUDA_VISIBLE_DEVICES\"]=\"2\"\n",
```
benfred, line 73: We should leave CUDA_VISIBLE_DEVICES out of notebooks (especially since this breaks us on systems with less than 3 gpu's).

examples/rossmann-store-sales-preproc.ipynb
```diff
@@ -29,7 +29,7 @@
...
     "\n",
     "# Preprocessing the Rossmann Store Sales Dataset\n",
-    "Here we implement some feature engineering outlined by FastAI in [their example solution](https://github.com/fastai/course-v3/blob/master/nbs/dl1/lesson6-rossmann.ipynb) to the [Kaggle Rossmann Store Sales competition](https://www.kaggle.com/c/rossmann-store-sales) using Rapids cuDF library (GPU DataFrame library). We've simplified some sections and left out some of the documentation to keep things neat, so feel free to consult the original preprocessing [notebook](https://github.com/fastai/course-v3/blob/master/nbs/dl1/rossman_data_clean.ipynb) for explanations of the feature engineering going on."
+    "Here we implement some feature engineering outlined by FastAI in [their example solution](https://github.com/fastai/fastai/blob/master/courses/dl1/lesson3-rossman.ipynb) to the [Kaggle Rossmann Store Sales competition](https://www.kaggle.com/c/rossmann-store-sales). We've simplified some sections and left out most of the documentation to keep things neat, so feel free to consult the original notebook for explanations of the feature engineering going on."
```
benfred, line 32: It looks like you've reverted all the changes from #259 - is this intentional? I realize the changes there broke  the notebook, but I think we probably just want a minimal fix

nvtabular/framework_utils/pytorch/models.py
jperez999: done

benfred (approved):

PR #293, opened 2020-09-10
Docs data

jperez999:
This adds documentation mentioning that the NVT Dataset is optimized for parquet files with a rowgroup size of 128mb. And it links to the optimize_criteo notebook to show user an example of changing a CSV to optimized Parquet format.

### DIFF
--- a/nvtabular/io/dataset.py
+++ b/nvtabular/io/dataset.py
@@ -127,12 +127,12 @@ class Dataset:
         dataset = Dataset(ddf)
 
     Dataset Optimization Tips (DOTs)
-    When creating a NVT Dataset, to get maximum performance from the library,
-    the data must be crafted using parquet format files, with a optimized
-    row group size of 128mb. If your current dataset is represented in a format
-    other than parquet, NVTabular supports reading from CSV files. Take a look at
-    this notebook_, it is used to turn the original unformatted Criteo CSV dataset
-    and create a new parquet dataset optimized for use in an NVT Dataset.
+    The NVTabular dataset should be created from Parquet files in order
+    to get the best possible performance, preferably with a row group size
+    of around 128MB.  While NVTabular also supports reading from CSV files,
+    reading CSV can be over twice as slow as reading from Parquet. Take a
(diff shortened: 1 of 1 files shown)


jperez999:
rerun tests

jperez999:
rerun tests

nvtabular/io/dataset.py
```diff
@@ -125,6 +125,17 @@ class Dataset:
...
         ddf = ddf.sort_values("user_rank", ignore_index=True)
         dataset = Dataset(ddf)
+    
+    Dataset Optimization Tips (DOTs)
+    When creating a NVT Dataset, to get maximum performance from the library,
+    the data must be crafted using parquet format files, with a optimized 
+    row group size of 128mb. If your current dataset is represented in a format 
+    other than parquet, NVTabular supports reading from CSV files. Take a look at
+    this notebook_, it is used to turn the original unformatted Criteo CSV dataset
+    and create a new parquet dataset optimized for use in an NVT Dataset.
```
benfred: There are a couple parts here that I think could be improved:
 * The first sentence has too many clauses (When creating ..., to get maximum  perf..., the data must ..., with a ...). Can we simplify this a bit?
 * `If your current dataset is represented in a format  other than parquet, NVTabular supports reading from CSV file` -> doesn't really mention why you shouldn't use CSV. I think we should explicitly mention poor performance
*  In the last sentence, I think we should change  ```it is used to``` to ```for an example of```

So something like this maybe? 

```suggestion
    Dataset Optimization Tips (DOTs)
    
    The NVTabular dataset should be created from Parquet files in order to get the best 
    possible performance, preferably with a row group size of around 128MB.  While NVTabular
    also supports reading from CSV files, reading CSV can be over twice as slow as reading from Parquet. 
    Take a look at this notebook_ for an example of transforming the original Criteo CSV dataset
    into a new Parquet dataset optimized for use with NVTabular.
```

jperez999: thats way better.

jperez999:
rerun tests

benfred (approved):

PR #296, opened 2020-09-10
[REVIEW] Petastorm hook-in

jperez999:
Preliminary working setup with petastorm feeding into our general pytorch model. Good for benchmarking.


jperez999:
This PR will introduce the benchmarking so that all models are tested against the same model. Each model benchmark is created in a separate notebook so that, in future work, we can create write-ups around each version and explain why we use it in benchmarking and the results.

benfred (commented):
Can we move the benchmarking notebooks into a separate directory?

Something like ```benchmarks/torch/criteo``` instead of in the ```examples``` directory

benfred (approved):

PR #327, opened 2020-09-30
JoC WnD Outbrain model preprocessing with NVT notebook

rnyak:
Added a note for the train-validation split cell to warn the users about not running the stratified sampling cell again when they rerun the entire notebook.


rnyak:
closing this issue because this change is applied to new Outbrain nb.

PR #334, opened 2020-10-07
Workflow memory-pressure optimizations

rjzamora:
An NVTabular `Workflow` object will currently update the internal `ddf` after every transform stage.  This PR includes a few optimizations to avoid the memory-pressure issues that may result from this.

### DIFF
--- a/nvtabular/workflow.py
+++ b/nvtabular/workflow.py
@@ -567,6 +567,7 @@ class Workflow(BaseWorkflow):
         self.ddf = None
         self.client = client
         self._shuffle_parts = False
+        self._base_phase = 0
 
     def set_ddf(self, ddf, shuffle=None):
         if isinstance(ddf, (dask_cudf.DataFrame, Dataset)):
@@ -602,10 +603,16 @@ class Workflow(BaseWorkflow):
     def exec_phase(self, phase_index, record_stats=True, update_ddf=True):
         """
         Gather necessary column statistics in single pass.
-        Execute one phase only, given by phase index
+        Execute statistics for one phase only (given by phase index),
+        but (laxily) perform all transforms for current and previous phases.
         """
         transforms = []
-        phases = range(phase_index + 1)
+
+        # Need to perform all transforms up to, and including,
+        # the current phase (not only the current phase).  We do this
+        # so that we can avoid persisitng intermediate transforms
+        # needed for statistics.
+        phases = range(self._base_phase, phase_index + 1)
         for ind in phases:
             for task in self.phases[ind]:
                 op, cols_grp, target_cols, _ = task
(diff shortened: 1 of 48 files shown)


rjzamora:
Note that this PR modifies the parquet reader to begin using `dask_cudf.read_parquet` - After [dask#6715](https://github.com/dask/dask/pull/6715) lands, this will allow the IO tasks to be fused with subsequent transformation operations at the task level.  This will help avoid the unnecessary spilling of "untransformed" data.

rjzamora:
@jperez999 @benfred - Any idea what may be the cause of the current CI failure?  It doesn't seem related to this PR, but I may be wrong...

nvtabular/workflow.py
```diff
@@ -591,43 +591,46 @@ def _aggregated_op(gdf, ops):
...
-            if isinstance(op, TransformOperator):
-                stats_context = self.stats if isinstance(op, DFOperator) else None
-                logic = op.apply_op
-                transforms.append((self.columns_ctx, cols_grp, target_cols, logic, stats_context))
-            elif not isinstance(op, StatOperator):
-                raise ValueError("Unknown Operator Type")
-
-        # Preform transforms as single dask task (per ddf partition)
+        phases = range(phase_index + 1)
+        for ind in phases:
```
benfred, line 609: This is sorta confusing to me - maybe we should add some comments on whats happening? Before this function only executed a single phase (which is consistent with the docstring) - but now its executing everything after phase_index,  but not phase_index itself?

benfred (commented):
Looks good to me - but I did have one question about a part that I didn't quite understand.

Are the moments changes  anything that can be upstreamed into dask?

rjzamora:
> Are the moments changes anything that can be upstreamed into dask?

I'm sure the upstream implementation can be improved, but it can be tricky to preserve the *general* nature of the [`apply_concat_apply`](https://github.com/dask/dask/blob/9d19952c71daeb96ec6ecef884559827f04183e8/dask/dataframe/core.py#L4935) code path while improving performance for **both** cudf and pandas-backed collections.  In this case it will be especially hard to match the performance of a custom graph, since we can manually fuse tasks for count/mean/std in a way that dask cannot do.

nvtabular/workflow.py
benfred: err my bad - i see that its going from phase 0  to phase_index (instead of from phase_index+1 on) . My bad =). Still - I think we should mention why we are we repeating all the previous phases here.

rjzamora: Yeah, good point - I'll add some better comments :)

rjzamora: Thanks @benfred - Hopefully the new comments are somewhat helpful?

benfred (approved):

Issue #78, opened 2020-06-09
[FEA] Wrap HashBucket functionality into Categorify op


alecgunny:
Since hash bucketing is a useful functionality for out-of-vocabulary (oov) and below-frequency-threshold values, and since hash bucketing and categorical lookups are just two different sides of the same category -> index coin, I think it makes sense to combine them and allow the `freq_threshold` and `num_buckets` kwargs dictate the desired behavior (this would also include getting rid of the `use_freqs` kwarg, since that info can be inferred by the value of `freq_threshold`). One potential complication is that this makes the required stats dependent on the kwargs, since the pure-hash-bucket mode wouldn't need an encoder.

I've compiled [a table](https://docs.google.com/spreadsheets/d/1XKcYByU7whyj67h5zezCGM8Q1nhxnK3LuLTFGt_6mn0/edit?usp=sharing) about what I think reasonable behavior for different kwarg combinations might be. Feel free to take a look and comment feedback on what might make sense. I can start mocking up an implementation once we're locked in on something workable.

PR #346, opened 2020-10-12
[WIP] Wrap HashBucket functionality into Categorify op

rnyak:
This PR is created 
- to address FEA https://github.com/NVIDIA/NVTabular/issues/78 : currently it can perform hashing when calling categorify op providing freq_threshold = 0 and num_buckets as kwargs.  The hashing case where freq_threshold >1 is not applied yet.

### DIFF
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -784,7 +784,7 @@ def _encode(
                 gdf[selection_l], side="left", na_position="first"
             )
         labels[labels >= len(value[selection_r])] = na_sentinel
-    if search_sorted is True and freq_threshold > 0:
+    if search_sorted and freq_threshold > 0:
         raise ValueError("cannot use search_sorted=True with freq_threshold > 0.")
     if list_col:
         labels = _encode_list_column(gdf[selection_l[0]], labels)
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -102,8 +102,8 @@ class Categorify(DFOperator):
     name_sep : str, default "_"
         String separator to use between concatenated column names
         for multi-column groups.
-    search_sorted: default False. Set it True to apply searchsorted algorithm
-        in encoding.
+    search_sorted: bool, default False.
+        Set it True to apply searchsorted algorithm in encoding.
     """
 
     default_in = CAT
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -102,7 +102,7 @@ class Categorify(DFOperator):
     name_sep : str, default "_"
         String separator to use between concatenated column names
         for multi-column groups.
-    search_sorted: bool, default False.
+    search_sorted : bool, default False.
         Set it True to apply searchsorted algorithm in encoding.
     """
 
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -760,7 +760,7 @@ def _encode(
         value.index.name = "labels"
         value.reset_index(drop=False, inplace=True)
 
-    if search_sorted is False or freq_threshold > 0:
+    if not search_sorted or freq_threshold > 0:
         if list_col:
             codes = cudf.DataFrame({selection_l[0]: gdf[selection_l[0]].list.leaves})
             codes["order"] = cp.arange(len(codes))
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -102,6 +102,8 @@ class Categorify(DFOperator):
     name_sep : str, default "_"
         String separator to use between concatenated column names
         for multi-column groups.
+    search_sorted: default False. Set it True to apply searchsorted algorithm
+        in encoding.
     """
 
     default_in = CAT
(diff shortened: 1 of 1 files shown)


rnyak:
> These changes look good @rnyak - Thanks!
> 
> I think the only thing missing is a simple test to confirm that the encoding works for validation data with new categories (the motivation for this PR). It would also be best to test the `search_sorted=True` case now that it is not the default (otherwise we are losing test coverage for that algorithm).

@rjzamora do you mean adding a unit test?  if yes, sure I will add that. Also,  since I am wrapping hash_bucket functionality into categorify the code will change again.

rjzamora:
> @rjzamora do you mean adding a unit test? if yes, sure I will add that. Also, since I am wrapping hash_bucket functionality into categorify the code will change again.

Yep, exactly.  With every PR we want to make sure that (1) the unit tests are covering new functionality, and (2) the unit tests are still covering the functionality from before the PR.

nvtabular/ops/categorify.py
```diff
@@ -740,7 +747,7 @@ def _encode(name, storage_name, path, gdf, cat_cache, na_sentinel=-1, freq_thres
         value.index.name = "labels"
         value.reset_index(drop=False, inplace=True)
```
rjzamora, line 749: May need to raise an error if `freq_threshold>0` **and**  `search_sorted` is `True`.  However, it is probably best to catch this problem quickly in the `Categorify` initializer.

nvtabular/ops/categorify.py
```diff
@@ -753,8 +773,8 @@ def _encode(name, storage_name, path, gdf, cat_cache, na_sentinel=-1, freq_thres
         ).sort_values("order")["labels"]
         labels.fillna(na_sentinel, inplace=True)
         labels = labels.values
-    else:
-        # Use `searchsorted` if we are using a "full" encoding
+    if search_sorted is True:
```
benfred, line 776:
```suggestion
    if search_sorted:
```

nvtabular/ops/categorify.py
```diff
@@ -740,7 +760,7 @@ def _encode(name, storage_name, path, gdf, cat_cache, na_sentinel=-1, freq_thres
         value.index.name = "labels"
         value.reset_index(drop=False, inplace=True)
 
-    if freq_threshold > 0:
+    if search_sorted is False or freq_threshold > 0:
```
benfred, line 763:
```suggestion
    if not search_sorted or freq_threshold > 0:
```

nvtabular/ops/categorify.py
```diff
@@ -764,7 +784,8 @@ def _encode(name, storage_name, path, gdf, cat_cache, na_sentinel=-1, freq_thres
                 gdf[selection_l], side="left", na_position="first"
             )
         labels[labels >= len(value[selection_r])] = na_sentinel
-
+    if search_sorted is True and freq_threshold > 0:
```
benfred, line 787:
```suggestion
    if search_sorted and freq_threshold > 0:
```

nvtabular/ops/categorify.py
```diff
@@ -764,7 +784,8 @@ def _encode(name, storage_name, path, gdf, cat_cache, na_sentinel=-1, freq_thres
...
+        raise ValueError("cannot use search_sorted=True with freq_threshold > 0.")
```
benfred, line 788: We probably want this check in the constructor - I think at this stage we will have already calculated the categories and applied them to at least one partition before failing,

nvtabular/ops/categorify.py
```diff
@@ -277,7 +285,17 @@ def apply_op(
                 new_gdf[name] = new_gdf[new_col]
                 new_gdf.drop(columns=[new_col], inplace=True)
 
+        if self.num_buckets is not None and self.freq_threshold == 0:
+            hash_bucket_op = HashBucket(self.num_buckets)
+            new_gdf = hash_bucket_op.apply_op(new_gdf, columns_ctx, "categorical")
+
+        if self.num_buckets == 0:
+            raise ValueError(
+                "For hashing num_buckets should be an int > 1, otherwise set it to None."
```
benfred, line 294: We should move this check to the consturctor

nvtabular/ops/categorify.py
```diff
@@ -277,7 +285,17 @@ def apply_op(
                 new_gdf[name] = new_gdf[new_col]
                 new_gdf.drop(columns=[new_col], inplace=True)
 
+        if self.num_buckets is not None and self.freq_threshold == 0:
+            hash_bucket_op = HashBucket(self.num_buckets)
+            new_gdf = hash_bucket_op.apply_op(new_gdf, columns_ctx, "categorical")
```
benfred, line 290: Whats the use case for doing the full categorical encoding and then hashing the output?  (like  do we need to incur the expense of doign the categorical encoding first if we are just going to hash everything after?)

nvtabular/ops/categorify.py
rnyak: @benfred thanks, yes Rick also mentioned that, so I added this condition into the constructor. will push the changes.

nvtabular/ops/categorify.py
rnyak: @benfred good point. you are right. I am working on that part. The example for that would be the case where `freq_threshold >1 + num_buckets >1`.

rnyak:
@benfred closing this will create a new one.

Issue #312, opened 2020-09-22
[BUG] Categorical encoding of unknown values 


benfred:
**Describe the bug**
When doing categorical encoding , unknown values in the validation set should be mapped to an unknown category (id 0) . However, when run without frequency caps - we're using 'search_sorted' to lookup the categorical ids. When the categorical isn't in the train set, this maps to the nearest key by sort order instead of setting to an unknown value.

**Steps/Code to reproduce bug**
```
import cudf
train = cudf.DataFrame({'user_id': [1, 2, 3, 5, 6], 'item_id': ["a", "g", "c", 'd', 'e']})
val = cudf.DataFrame({'user_id': [1, 2, 4, 7, 8], 'item_id': ["a", "b", "c", "a", "b"]})
proc = nvt.Workflow(
    cat_names=cat_names,
    cont_names= [],
    label_name=[])
proc.add_cat_preprocess(nvt.ops.Categorify())
```

Running this code shows the valid set ids get mapped to ``` 1, 2, 4, 0, 0``` when it should be ``` 1, 2, 0, 0, 0```.

rjzamora:
I added the "good first issue" label, because I imagine an appropriate fix is to make the searchsorted algorithm an option.  Right now, we use this algorithm whenever [there is no frequency cap](https://github.com/NVIDIA/NVTabular/blob/b77fd7776631bb542ef99125a3c559f73f4143ed/nvtabular/ops/categorify.py#L737).  Instead, we could let the user specify the transformation algorithm they want - e.g. `algo="merge"` (default) or `algo="searchsorted"` (faster).

AnshuTrivedi:
Hi @rjzamora @benfred I want to fix issue and contribute.Please assign  this issue to me.

benfred:
Thanks @AnshuTrivedi  - we'd appreciate the contribution! Let me know how you make out =)

rnyak:
@benfred @AnshuTrivedi Please also consider these cases:
```
train = cudf.DataFrame({'user_id': [1, 2, 3, 5, 6], 'item_id': ["a", "g", "c", 'd', 'e']})
val = cudf.DataFrame({'user_id': [1, 2, 4, 7, 8, np.nan, np.NAN], 'item_id': ["a", "b", "c", "a", "b", np.NAN, '']})
```

[...]

PR #354, opened 2020-10-15
[REVIEW] add searchsorted as a param and fix to_pandas type casting error when nulls are present

rnyak:
This PR is created to fix the issue reported in https://github.com/NVIDIA/NVTabular/issues/312. 
The unit test `test_categorify_freq_limit` was also updated accordingly.

### DIFF
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -197,11 +197,10 @@ class Categorify(DFOperator):
         self.encode_type = encode_type
         self.search_sorted = search_sorted
 
-        self.isfreqthreshold = (
-            isinstance(self.freq_threshold, int) and self.freq_threshold > 0
-        ) or (isinstance(self.freq_threshold, dict))
-        if self.search_sorted and self.isfreqthreshold:
-            raise ValueError("cannot use search_sorted=True with freq_threshold > 0.")
+        if self.search_sorted and self.freq_threshold:
+            raise ValueError(
+                "cannot use search_sorted=True with anything else than the default freq_threshold"
+            )
 
     @property
     def req_stats(self):
@@ -277,7 +276,6 @@ class Categorify(DFOperator):
                 if isinstance(self.freq_threshold, dict)
                 else self.freq_threshold,
                 search_sorted=self.search_sorted,
-                isfreqthreshold=self.isfreqthreshold,
             )
             if self.dtype:
                 new_gdf[new_col] = new_gdf[new_col].astype(self.dtype, copy=False)
@@ -725,15 +723,7 @@ def _category_stats(
 
 
 def _encode(
-    name,
-    storage_name,
-    path,
-    gdf,
-    cat_cache,
-    na_sentinel=-1,
-    freq_threshold=0,
-    search_sorted=False,
-    isfreqthreshold=True,
+    name, storage_name, path, gdf, cat_cache, na_sentinel=-1, freq_threshold=0, search_sorted=False,
 ):
     value = None
     selection_l = name if isinstance(name, list) else [name]
@@ -762,7 +752,7 @@ def _encode(
         value.index.name = "labels"
         value.reset_index(drop=False, inplace=True)
 
-    if not search_sorted or isfreqthreshold:
+    if not search_sorted:
         if list_col:
             codes = cudf.DataFrame({selection_l[0]: gdf[selection_l[0]].list.leaves})
             codes["order"] = cp.arange(len(codes))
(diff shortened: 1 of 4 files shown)


nvtabular/ops/categorify.py
```diff
@@ -266,6 +276,8 @@ def apply_op(
                 freq_threshold=self.freq_threshold[name]
                 if isinstance(self.freq_threshold, dict)
                 else self.freq_threshold,
+                search_sorted=self.search_sorted,
+                isfreqthreshold=self.isfreqthreshold,
```
benfred, line 280: Why do we need this parameter? Can't we just keep the logic of freq_threshold > 0 in the encode function?

nvtabular/ops/categorify.py
```diff
@@ -753,13 +775,17 @@ def _encode(name, storage_name, path, gdf, cat_cache, na_sentinel=-1, freq_thres
...
+        # Use `searchsorted` if doing full encoding (optional).
         if list_col:
             labels = value[selection_r].searchsorted(
                 gdf[selection_l[0]].list.leaves, side="left", na_position="first"
             )
         else:
+            for c in selection_l:
+                typ = gdf[c].dtype
+                if typ != value[c].dtype:
+                    value[c] = value[c].astype(typ)
```
benfred, line 788: Is this for the problem of int columns with nulls getting converted to floats? If so - I'd rather solve in a way that doesn't require data being cast multiple times (like from int->float->int).

Rick started a thread on the cudf about this (https://github.com/rapidsai/cudf/issues/6536) - I think we probably want to spill to arrow rather than pandas (which would be out of scope of this PR

nvtabular/ops/categorify.py
```diff
@@ -753,13 +775,17 @@ def _encode(name, storage_name, path, gdf, cat_cache, na_sentinel=-1, freq_thres
         ).sort_values("order")["labels"]
         labels.fillna(na_sentinel, inplace=True)
         labels = labels.values
-    else:
-        # Use `searchsorted` if we are using a "full" encoding
+    if search_sorted:
```
benfred, line 778: Can this just be an else still - just to make it clear that we're doing one branch or the other?

nvtabular/ops/categorify.py
```diff
@@ -192,6 +195,13 @@ def __init__(
         self.cat_cache = cat_cache
         self.stat_name = "categories"
         self.encode_type = encode_type
+        self.search_sorted = search_sorted
+
+        self.isfreqthreshold = (
+            isinstance(self.freq_threshold, int) and self.freq_threshold > 0
+        ) or (isinstance(self.freq_threshold, dict))
+        if self.search_sorted and self.isfreqthreshold:
+            raise ValueError("cannot use search_sorted=True with freq_threshold > 0.")
```
benfred, line 204: Can we simplify this? Like isn't just testing if ```freq_threshold``` evaluates to true enough here, without needing a bunch of isinstance checks?

```suggestion
        if self.search_sorted and self.freq_threshold:
            raise ValueError("cannot use search_sorted=True with freq_threshold > 0.")
```

benfred (changes requested):

nvtabular/ops/categorify.py
rjzamora: Right - I told @rnyak I would take care of it separately, but may make sense to include the small change here.

The idea is to replace all instances (in `categorify.py`) of

```python
df = _concat(dfs, ignore_index)
if on_host:
    df.reset_index(drop=True, inplace=True)
    df = cudf.from_pandas(df)
```

with

```python
if on_host:
    df = pa.concat_tables(dfs)
    df = cudf.DataFrame.from_arrow(df)
else:
    df = _concat(dfs, ignore_index)
```
and all instances of

```python
.to_pandas()
```

to

```python
.to_arrow(preserve_index=False)
```

(where `pa` is `pyarrow`). 

Does this sound reasonable?

nvtabular/ops/categorify.py
rnyak: when  freq_threshold > 0 , I was getting error when `freq_threshold` is set as a dictionary, so I decided to create an extra param.

nvtabular/ops/categorify.py
rnyak: > Can we simplify this? Like isn't just testing if `freq_threshold` evaluates to true enough here, without needing a bunch of isinstance checks?

@benfred same as  above, that was due to an error I was getting when freq_threshold is a dictionary. simply putting `f self.search_sorted and self.freq_threshold:` did not work. I will test it again, just in case.

nvtabular/ops/categorify.py
rnyak: @benfred  what are we deciding here? shall we fix `nulls getting converted to floats` in another PR? should I leave this part of code as it is?

benfred: Lets do in a separate PR - this is unrelated to the searchsorted issue afaict. 

This code shouldn't be part of the new PR though (and should just be removed from this PR). The new PR should use Rick's suggestions

nvtabular/ops/categorify.py
benfred: I think we can remove the parameter. 

The _encode function doesn't take a dictionary as the freq_threshold param, and only takes integers (like we get the value from the dictionary on line 276 if its a dict)

benfred:
rerun tests

benfred (approved):

PR #357, opened 2020-10-16
[REVIEW] Write parquet list cols

benfred:
Currently cudf doesn't support writing list columns using the chunked writer.
This PR refactors to write out a new file for each partition instead of
using the chunked writer


benfred:
rerun tests

jperez999 (approved):

PR #369, opened 2020-10-20
[Review] Basic AvroDatasetEngine

### LINKED ISSUE #314: [FEA] Support reading datasets from Avro files
We should add support for reading from Avro datasets.

rjzamora:
Closes #314 (but `AvroDatasetEngine` should be revised when Avro supports gets into Dask-CuDF)

This is a basic/limited implementation of an Avro-based `DatasetEngine` (`AvroDatasetEngine`).  Note that this implementation can be further simplified once Avro support is standardized in `dask.dataframe`/`dask_cudf`.

Since ther is not avro->arrow suport in pyarrow, we require the `uavro` library for the "planning" stage (metadata parsing and dataset decomposition).  We also require the `fastavro` library for testing (to write out data in avro format), because uavro is **for reading only**.  Note that we can certainly drop the dependency on `uavro` altogether (in favor of `fastavro` only), but the current solution is clearly more performant.

cc @albert17

### DIFF
--- a/nvtabular/io/avro.py
+++ b/nvtabular/io/avro.py
@@ -13,7 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import warnings
+
 import cudf
+import numpy as np
 import uavro as ua
 from dask.base import tokenize
 from dask.dataframe.core import new_dd_object
@@ -44,6 +47,15 @@ class AvroDatasetEngine(DatasetEngine):
         # Get list of pieces for each output
         pieces, meta = self.process_metadata(columns=columns)
 
+        # TODO: Remove warning and avoid use of uavro in read_partition when
+        # cudf#6529 is fixed (https://github.com/rapidsai/cudf/issues/6529)
+        if len(pieces) > len(self.paths):
+            warnings.warn(
+                "Row-subset selection in cudf avro reader is currently broken. "
+                "Using uavro engine until cudf#6529 is addressed. "
+                "EXPECT POOR PERFORMANCE!! (compared to cuio-based reader)"
+            )
+
         # Construct collection
         token = tokenize(self.fs, self.paths, self.part_size, columns)
         read_avro_name = "read-avro-partition-" + token
@@ -73,14 +85,15 @@ class AvroDatasetEngine(DatasetEngine):
         for path in self.paths:
             file_size = self.fs.du(path)
             if file_size > self.blocksize:
+                part_count = 0
                 with open(path, "rb") as fo:
                     header = ua.core.read_header(fo)
                     ua.core.scan_blocks(fo, header, file_size)
                     blocks = header["blocks"]
 
-                    (file_row_offset, part_row_count) = (0, 0)
-                    (file_block_offset, part_block_count) = (0, 0)
-                    (file_byte_offset, part_byte_count) = (blocks[0]["offset"], 0)
+                    file_row_offset, part_row_count = 0, 0
+                    file_block_offset, part_block_count = 0, 0
+                    file_byte_offset, part_byte_count = blocks[0]["offset"], 0
 
                     for i, block in enumerate(blocks):
                         part_row_count += block["nrows"]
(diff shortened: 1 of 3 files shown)


albert17:
rerun tests

albert17:
Usage of `cudf.read_avro(filename, num_rows) `gives some problems:
  - cudf 0.15: Using parameter `num_rows` produces a segmentation fault (it is fixed in 0.16)
  - cudf 0.16: Using parameter `num_rows` does not have any effect (There is an issue created: https://github.com/rapidsai/cudf/issues/65290)

We are supposed to move to dask_cudf implementation when it is done in the cudf side, there is a FEA request: https://github.com/rapidsai/cudf/issues/6496

Tests work in local, I have added uavro package to CI environment. Existing tests are being passed successfully, which probably means that the examples are too small so there are not partitions, or that the implementation is not using `num_rows` as it is supposed. @rjzamora what do you think?

Following @benfred indications earlier today, we could get this PR in as it is and revisit for further testing when `num_rows` is fixed (cudf 0.17?).  Any other thing we would like to modify or change for this version?

Also we are interested in tracking when these types in the schema are supported by cudf: https://github.com/rapidsai/cudf/issues/6323, https://github.com/rapidsai/cudf/issues/6322, https://github.com/rapidsai/cudf/issues/6319

rjzamora:
> Tests work in local, I have added uavro package to CI environment. Existing tests are being passed successfully, which probably means that the examples are too small so there are not partitions, or that the implementation is not using num_rows as it is supposed. @rjzamora what do you think?

Right. The test does not seem to be testing proper block splitting.  My mistake - One of the test parameters *used* to be `10_000` (rather than `1_000`), because fastavro seems to use a block size of ~2k rows for this schema.

tests/unit/test_io.py
```diff
@@ -296,3 +298,62 @@ def test_parquet_lists(tmpdir, freq_threshold):
...
+@pytest.mark.parametrize("part_size", [None, "1KB"])
+@pytest.mark.parametrize("size", [100, 1000])
+@pytest.mark.parametrize("nfiles", [1, 2])
+def test_avro_basic(tmpdir, part_size, size, nfiles):
+
+    # Require uavro and fastavro library.
+    # Note that fastavro is only required to write
+    # avro files for testing, while uavro is actually
+    # used by AvroDatasetEngine.
+    fa = pytest.importorskip("fastavro")
```
benfred, line 312: We should probably get fastavro/uavro installed on the CI container to test this out (more a note for myself than anything =)

nvtabular/io/avro.py
```diff
@@ -0,0 +1,138 @@
...
+                                file_row_offset + part_row_count,
+                                0,
+                            )
+                            (file_block_offset, part_block_count) = (
+                                file_block_offset + part_block_count,
+                                0,
+                            )
+                            (file_byte_offset, part_byte_count) = (
+                                file_byte_offset + part_byte_count,
+                                0,
```
benfred, line 108: I think this can be more cleanly expressed without tuple unpacking - something like 
```suggestion
                            file_row_offset += part_row_count
                            file_block_offset += part_block_count
                            file_byte_offset += part_byte_count
                            part_row_count = part_row_count = part_block_count = 0
```

nvtabular/io/avro.py
```diff
@@ -0,0 +1,138 @@
...
+            file_size = self.fs.du(path)
+            if file_size > self.blocksize:
+                with open(path, "rb") as fo:
+                    header = ua.core.read_header(fo)
+                    ua.core.scan_blocks(fo, header, file_size)
+                    blocks = header["blocks"]
+
+                    (file_row_offset, part_row_count) = (0, 0)
+                    (file_block_offset, part_block_count) = (0, 0)
+                    (file_byte_offset, part_byte_count) = (blocks[0]["offset"], 0)
```
benfred, line 83: nitpick: the parenthesis around each tuple isn't necessary:

```suggestion
                    file_row_offset, part_row_count = 0, 0
                    file_block_offset, part_block_count = 0, 0
                    file_byte_offset, part_byte_count = blocks[0]["offset"], 0
```

nvtabular/io/avro.py
```diff
@@ -0,0 +1,138 @@
...
+
+        return pieces, meta
+
+    @classmethod
+    def read_partition(cls, fs, piece, columns):
+
+        path = piece["path"]
+        if "rows" in piece:
+            skiprows, num_rows = piece["rows"]
+            df = cudf.io.read_avro(path, skiprows=skiprows, num_rows=num_rows)
```
benfred, line 131: Should we log a warning that num_rows isn't working properly in cudf here?

benfred (commented):
Thanks for this! Looks mostly good to me - left some minor comments below

tests/unit/test_io.py
benfred: nvm - I see alberto has done this. (thanks!)

benfred (approved):

rjzamora:
Ah - I changed/simplified the squashed commit message and still forgot to remove `"[Review]"`... Sorry about that.

benfred:
> Ah - I changed/simplified the squashed commit message and still forgot to remove "[Review]"... Sorry about that.

I do this *all* the time - which is why I was suggesting that we get rid of the ```[REVIEW]``` tags in the title and just go with GitHub 'Draft' pr status' for WIP =)

Issue #371, opened 2020-10-21
[FEA] Hashed cross op


alecgunny:
Hashed feature crossing is a common element of a lot of TensorFlow workflows due to the [`crossed_column`](https://www.tensorflow.org/api_docs/python/tf/feature_column/crossed_column) feature column. Support for this might look something like

```
class HashedCrossing(Op):
    def __init__(self, columns, bucket_sizes, **kwargs):
        kwargs["replace"] = False # omnibus behavior acrosses all crosses probably doesn't make sense
        super().__init__(**kwargs)

        assert len(columns) == len(bucket_sizes)
        self.columns = columns # list of (column_1, column_2, ..., column_k) tuples
        self.bucket_sizes = bucket_sizes # list of ints

    def op_logic(self, gdf, target_columns):
        # this will only cover single-hot use case
        # list like support will be complicated by cartesian product across inputs
        for columns, bucket_size in zip(self.columns, self.bucket_sizes):
              val = 0
              for column in columns:
                  val |= gdf[column].hash_values() # or however we want to do this aggregation
              val = cudf.Series(val).hash_values() % bucket_size
              gdf["_x_".join(columns)] = val
        return gdf
```

alecgunny:
Solved by #379

PR #379, opened 2020-10-23
Adding ops for feature column functionality and feature column to workflow mapping function

alecgunny:
Increasing NVTabular compatibility with TensorFlow feature column API by adding remaining necessary ops (cross op and bucketize) and a function which can map from a set of feature columns to an NVTabular workflow that performs all analogous preprocessing. Addresses #371 

`HashedCross` doesn't support multi-hot yet, and I'm not sure that extending to it will be necessarily easy. For reference, the TF cross op handles multi-hots by doing a cartesian product of all indices for each feature. See the documentation [here](https://www.tensorflow.org/api_docs/python/tf/feature_column/crossed_column).

Still need to add bucketized support and test everything.


alecgunny:
@benfred what's our stance on adding ops without list support? If we're ok with it, should we add a support matrix in the documentation?

alecgunny:
With the addition of bucketize, we should have full TF feature column coverage (minus the sequence columns which I won't worry about for now). The shared embeddings and weighted shared embeddings are more of Keras layers than they would be preprocessing steps, so will still need to add layers that cover those. But overall this should put us in pretty good shape. Just need to build tests, docs, and an example and we should be good to go

alecgunny:
rerun tests

alecgunny:
Adding an example for using the new utilities that will hopefully act as a simple introduction to the more advanced tf notebook

nvtabular/ops/hashed_cross.py
```diff
@@ -0,0 +1,57 @@
...
+        self.crosses = crosses
+        self.num_buckets = num_buckets
+
+    @annotate("HashedCross_op", color="darkgreen", domain="nvt_python")
+    def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None):
+        new_gdf = cudf.DataFrame()
+        for columns, bucket_size in zip(self.crosses, self.num_buckets):
+            val = 0
+            for column in columns:
+                val ^= gdf[column].hash_values()  # or however we want to do this aggregation
```
benfred, line 54: Our categorify op lets you pass in column groups, and takes an 'encode_type' parameter - which if set to 'combo' does the categorical encoding on the cross like this op:

https://github.com/NVIDIA/NVTabular/blob/f39e65e95d0af1d44ae9c2073a06c5a442d4de93/nvtabular/ops/categorify.py#L76-L81  

What do you think about rolling the functionality for this op into the HashBucket op to be consistent? (like if passed a multi column group, then doing the cross). 

Ronay is working on merging the HashBucket functionality with the categorify op - and this is one of the things that I think we could do to minimize the delta in functionality between the two

nvtabular/ops/hashed_cross.py
```diff
@@ -0,0 +1,57 @@
...
+            val = cudf.Series(val).hash_values() % bucket_size
```
benfred, line 55: Do we need to hash the value again after aggregating with xor? The most common hash value function for integers is the identity function (since your unlikely to do better ) - so I feel like we could simply down to something like : 
```suggestion
            val = val % bucket_size
```

nvtabular/loader/tensorflow.py
```diff
@@ -305,8 +305,8 @@ def _handle_tensors(self, cats, conts, labels):
         # would require output layers to match naming
         if len(self.label_names) > 1:
             labels = tf.split(labels, len(self.label_names), axis=1)
-        else:
-            labels = [labels]
+        # else:
+        #     labels = [labels]
```
benfred, line 309: Lets just delete this:
```suggestion
```

benfred (commented):
This is awesome! Thanks for this

nvtabular/ops/hashed_cross.py
alecgunny: Hmmm that's an interesting idea... I think that could make sense. I'll need to think more about it from a conceptual standpoint but they definitely seem equivalent

nvtabular/loader/tensorflow.py
alecgunny: Yeah I have this just straight up removed in my mh branch, fine with that

nvtabular/ops/hashed_cross.py
alecgunny: Yeah that probably makes sense. I'll admit my hash math background isn't super strong but I know the xor maintains the uniform distribution so doing the hash on each probably makes this superfluous

nvtabular/ops/hashed_cross.py
alecgunny: Based on the meeting today, do we want to shelve this until after the 0.3 release? While I agree that this functionality should all be wrapped together, the API standpoint feels non-trivial and at this point it might make more sense just to get this in for TF users (especially if it get primarily used by the `make_feature_column_workflow` function, since any API changes will be handled on the backend and wouldn't require users to update their code)

benfred: sounds good - we can always come back to this

benfred (approved):

benfred:
rerun tests

nvtabular/framework_utils/tensorflow/feature_column_utils.py
```diff
@@ -0,0 +1,251 @@
...
+
+        if isinstance(cat_column, fc.BucketizedColumn):
+            key = cat_column.source_column.key
+
+            # check if the source numeric column is being fed
+            # directly to the model. Keep track of both the
+            # boundaries and embedding dim so that we can wrap
+            # with either indicator or embedding later
+            if key in [col.key for col in numeric_columns]:
+                buckets[key] = (column.boundaries, embedding_dim)
```
andrey-klochkov-liftoff, line 135: I believe here and two lines below it should be `cat_column.boundaries`. It fails to find the `boundaries` column in my attempts to use this utility, and it passes if I replace `column.boundaries` with `cat_column.boundaries`.

PR #380, opened 2020-10-26
Wrap HashBucket functionality into Categorify op

rnyak:
This PR is created to address FEA #78. The goal is to perform hashing in Categorify op. After this PR is completed, a user can perform either

- full embedding (no frequency capping no hashing)
- frequency capping by setting frequency threshold param
- only hashing by setting `num_buckets` param
- frequency hashing (frequency capping + hashing the categories under the freq threshold  to num_buckets)


rnyak:
@benfred @rjzamora the current PR works for non-list columns, but `list-column` hashing is not completed due to https://github.com/NVIDIA/NVTabular/issues/381.

nvtabular/ops/categorify.py
```diff
@@ -829,3 +925,42 @@ def _encode_list_column(original, encoded):
         size=original.size,
         children=(original._column.offsets, encoded),
     )
+
+
+def _hash_bucket(gdf, num_buckets, col):
+    nb = num_buckets[col]
+    if is_list_dtype(gdf[col].dtype):
+        encoded = _encode_list_column(gdf[col], gdf[col].list.leaves.hash_values() % nb)
```
benfred, line 933: I don't think we need the encode_list_column call just yet here -  (and I don't think it will work?). Can you add a unittest w/ list processing to check

nvtabular/ops/categorify.py
```diff
@@ -766,11 +842,31 @@ def _encode(
...
-        labels = codes.merge(
-            value, left_on=selection_l, right_on=selection_r, how="left"
-        ).sort_values("order")["labels"]
-        labels.fillna(na_sentinel, inplace=True)
-        labels = labels.values
+                if buckets:
+                    if c in buckets:
+                        codes[c + "_hashed"] = _hash_bucket(gdf, buckets, c)
+        # apply frequency hashing
+        if freq_threshold and buckets:
```
benfred, line 849: Rather than 3 independent if statements - can you refactor to a 'if / elif / else'  chain?

nvtabular/ops/categorify.py
```diff
@@ -196,28 +199,72 @@ def __init__(
...
         self.search_sorted = search_sorted
+        self.hash_type = hash_type
 
         if self.search_sorted and self.freq_threshold:
             raise ValueError(
                 "cannot use search_sorted=True with anything else than the default freq_threshold"
             )
+        if num_buckets:
+            if isinstance(num_buckets, dict):
+                columns = [i for i in num_buckets.keys()]
```
benfred, line 210:
```suggestion
                columns = list(num_buckets)
```

benfred: If the user also passes in a list of columns here - this will get a little weird. Should we warn if columns is not null?

nvtabular/ops/categorify.py
```diff
@@ -196,28 +199,72 @@ def __init__(
...
+                self.num_buckets = num_buckets
+            elif isinstance(num_buckets, (tuple, list)):
+                assert columns is not None
+                assert len(columns) == len(num_buckets)
+                self.num_buckets = {col: nb for col, nb in zip(columns, num_buckets)}
+            elif isinstance(num_buckets, int):
+                self.num_buckets = num_buckets
+            else:
+                raise TypeError(
```
benfred, line 219: This should be a ValueError instead of a TypeError

nvtabular/ops/categorify.py
```diff
@@ -196,28 +199,72 @@ def __init__(
...
         self.encode_type = encode_type
         self.search_sorted = search_sorted
+        self.hash_type = hash_type
 
         if self.search_sorted and self.freq_threshold:
             raise ValueError(
                 "cannot use search_sorted=True with anything else than the default freq_threshold"
             )
+        if num_buckets:
+            if isinstance(num_buckets, dict):
```
benfred, line 209: Lets not have two levels of if statements here - can we get rid of the top level `if num_buckets` statement in favour of the checks below ?

nvtabular/ops/categorify.py
```diff
@@ -123,6 +124,8 @@ def __init__(
         encode_type="joint",
         name_sep="_",
         search_sorted=False,
+        num_buckets=None,
+        hash_type=None,
```
benfred, line 128: Whats the hash_type argument for?  Can you also update the docstring for the num_buckets argument?

benfred (commented):
Left some feedback below - can you also add a unittest for this?

nvtabular/ops/categorify.py
rnyak: @benfred will remove hash_type. I added it for a reason but it was not required afterwards.

nvtabular/ops/categorify.py
rnyak: @benfred list support is not complete yet.  Could you please clarify `I don't think we need the encode_list_column call just yet here` ? where should we define it? in the HashBucket op it was defined as follows, and that's where I got `_encode_list_column` script.

```
for col, nb in num_buckets.items():
            new_col = f"{col}_{self._id}"
            if is_list_dtype(gdf[col].dtype):
                encoded = _encode_list_column(gdf[col], gdf[col].list.leaves.hash_values() % nb)
            else:
                encoded = gdf[col].hash_values() % nb

            new_gdf[new_col] = encoded
```

Doesn't HashBucket work for list columns already?

nvtabular/ops/categorify.py
rnyak: @benfred I guess passing a list of columns should not be a case. The user should not send a list of columns. Only 3 options are accepted: dict, tuple or int. Otherwise a TypeError will raise.   
```           
 else:
                raise TypeError(
                    "`num_buckets` must be dict, iterable, or int, got type {}".format(
                        type(num_buckets)
                    )

**Comment by rnyak** [comment]
@benfred based on your comment for this section:

```
            elif isinstance(num_buckets, (tuple, list)):
                assert columns is not None
                assert len(columns) == len(num_buckets)
                self.num_buckets = {col: nb for col, nb in zip(columns, num_buckets)}

Not sure about tuple here. I dont think it is really accepted. `int`, 'dict` and `list` should be accepted. see HashBucket op also. Better to remove `tuple` here.

nvtabular/ops/categorify.py
benfred: HashBucket works for list columns already - I  meant that the 'encode_list_column' call should be done once, after all the encoding is done (it creates a new list column - which you don't want to operate on here).  

This function should be something like:
```
            if is_list_dtype(gdf[col].dtype):
                encoded =  gdf[col].list.leaves.hash_values() % nb
            else:
                encoded = gdf[col].hash_values() % nb

            new_gdf[new_col] = encoded
```

and the encode_list_column call should be the last thing that happens in the encode function

nvtabular/ops/categorify.py
rnyak: @benfred  I am modifying the conditions but I dont think `if / elif / else` chain would work here :/  The scenarios are a bit complicated to use `else`.

nvtabular/ops/categorify.py
```diff
@@ -201,23 +215,58 @@ def __init__(
...
+                    on_host=self.on_host,
+                    stat_name=self.stat_name,
+                    name_sep=self.name_sep,
+                ),
+                SetBuckets(
+                    columns=self.column_groups or self.columns,
+                    num_buckets=self.num_buckets,
+                    freq_limit=self.freq_threshold,
+                ),
+            ]
```
rjzamora, line 269: Just a nit-pick, but it's probably best to avoid so much code duplication.

```suggestion
        stats = [
            GroupbyStatistics(
                columns=self.column_groups or self.columns,
                concat_groups=self.encode_type == "joint",
                cont_names=[],
                stats=[],
                freq_threshold=self.freq_threshold,
                tree_width=self.tree_width,
                out_path=self.out_path,
                on_host=self.on_host,
                stat_name=self.stat_name,
                name_sep=self.name_sep,
            )
        ]
        if self.num_buckets:
            stats += [
                SetBuckets(
                    columns=self.column_groups or self.columns,
                    num_buckets=self.num_buckets,
                    freq_limit=self.freq_threshold,
                )
            ]
        return stats
```

nvtabular/ops/categorify.py
```diff
@@ -310,6 +360,17 @@ def _get_embedding_order(cat_names):
 def get_embedding_sizes(workflow):
     mh_cols = None
     cols = _get_embedding_order(workflow.columns_ctx["categorical"]["base"])
+    if "buckets" in workflow.stats.keys() and "freq_limit" not in workflow.stats.keys():
```
rjzamora, line 363: It might be helpful to add a brief comment to explain each of these cases.

rjzamora (commented):
Thanks for the work here @rnyak ! I took a quick pass an left a few minor comments.

nvtabular/ops/categorify.py
```diff
@@ -104,6 +105,18 @@ class Categorify(DFOperator):
         for multi-column groups.
     search_sorted : bool, default False.
         Set it True to apply searchsorted algorithm in encoding.
+    num_buckets : int, list of int, or dictionary:{column: num_hash_buckets}
+        Column-wise modulo to apply after hash function. Note that this
+        means that the corresponding value will be the categorical cardinality
+        of the transformed categorical feature. If given as an int, that value
+        will be used as the number of "hash buckets" for every feature. If
+        a list is provided, it must be of the same length as `columns` (which
```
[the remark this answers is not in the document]
rnyak: @rjzamora I actually agree on that. It would be confusing. I am fine removing `list` option (it was taken from HashBucket op) to avoid confusions/problems.

nvtabular/ops/categorify.py
```diff
@@ -201,23 +215,58 @@ def __init__(
...
+            columns = list(num_buckets)
+            self.num_buckets = num_buckets
+        elif isinstance(num_buckets, list):
+            assert columns is not None
+            assert len(columns) == len(num_buckets)
+            self.num_buckets = {col: nb for col, nb in zip(columns, num_buckets)}
+        elif isinstance(num_buckets, int) or num_buckets is None:
+            self.num_buckets = num_buckets
+            print(self.num_buckets)
+        elif num_buckets == 0 or 0 in num_buckets:
```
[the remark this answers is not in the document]
rnyak: @rjzamora good question. this is covering such case: `num_buckets={"author": 10, "engaging_user": 20, "platform":0}` or `num_buckets=0`, and raising the ValueError below otherwise:
```            raise ValueError(
                "For hashing num_buckets should be an int > 1, otherwise set it to None."
            )
```
Does it make sense? I am open to suggestions.

nvtabular/ops/categorify.py
rnyak: Thanks!

nvtabular/ops/categorify.py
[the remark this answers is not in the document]
rjzamora: I see - But you wouldn't make it into this code block for either of the cases you mentioned, because `num_buckets` would be an `int` or `dict`.  Perhaps you don't want this to be an `elif` (or want to check for zero values in the blocks above)?

rnyak:
@benfred I used hashcrossed op's technique for hashing when we use `encode_typpe="combo"` within categorify op. But I am not sure that this logic is what we want for hashing, because this will map `("User_A", "User_B")` and `("User_B", "User_A")` to the same bucket whereas without hashing, we map/encode them to different categorical numbers(ints) .

```
for column in columns:
                val ^= gdf[column].hash_values()  # or however we want to do this aggregation
            val = val % bucket_size

File: nvtabular/ops/categorify.py [hunk 11]
```diff
@@ -775,15 +836,41 @@ def _encode(
...
+                            gdf, buckets, tuple(selection_l), encode_type="combo"
+                        )
+                    else:
+                        na_sentinel = _hash_bucket(gdf, buckets, selection_l, encode_type="combo")
+        # apply frequency hashing
+        if freq_threshold and buckets and storage_name in buckets:
+            merged_df = codes.merge(
+                value, left_on=selection_l, right_on=selection_r, how="left"
+            ).sort_values("order")
+            merged_df.reset_index(drop=True, inplace=True)
```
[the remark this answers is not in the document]
rnyak: @benfred this is necessary because index of merged df from line 856-858 sort of messed, thus, yields wrong results after doing `merged_df["labels"].fillna(cudf.Series(na_sentinel + max_id + 1), inplace=True)`. Therefore, resetting index for merged_df before solves that issue.

nvtabular/ops/categorify.py
```diff
@@ -775,15 +836,41 @@ def _encode(
...
-        labels = labels.values
+                if buckets and c in buckets and encode_type == "joint":
+                    na_sentinel = _hash_bucket(gdf, buckets, c)
+                elif buckets and encode_type == "combo":
+                    if len(selection_l) > 1:
+                        na_sentinel = _hash_bucket(
+                            gdf, buckets, tuple(selection_l), encode_type="combo"
+                        )
+                    else:
+                        na_sentinel = _hash_bucket(gdf, buckets, selection_l, encode_type="combo")
```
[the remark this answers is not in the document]
rnyak: @benfred if I do this change , it breaks `_hash_bucket` function calculation, so it would not work. 

1. selection_l is an array like ['author'] that's why I send `c` which is `author` to hash_bucket for `joint` encode type.
2. tuple(selection_l) is also required check based on how combo is designed. 

I am working to find a way to incorporate your change. Let's see.

benfred (approved):

PR #385, opened 2020-10-28
TF and PT dataloader list support

alecgunny:
Moving MH support from #365 into backend then adding TensorFlow support on top of it. Still needs tests, benchmarking, embedding compatibility on the TF side, and I'm not sure how I feel about the naming conventions adopted (multi-hot columns get `__values` and `__nnz` appended to the end of their TF dict representation), but functionality should be there.

Note that while PT is using offsets representation, TF is using nnz since RaggedTensor embedding layer will need that. Eventually we should migrate to offsets when we build custom EmbeddingBag op, so this is controllable with a keyword.

The other key functionality I've added is moving all the conversion from gdf to framework tensors into a `make_tensor` method on the dataloaders. This way, you can transform "arbitrary" datarames (to the extent they match the schema used to initialize the dataloader) into framework tensors. The intended use case is if you have some test dataset which hasn't yet been transformed by the preprocessing workflow, you can iterate through it, apply the workflow, then send it into a framework for prediction in one foul swoop.


alecgunny:
Switching to ready for review so more complicated testing can be done by third parties while unit tests and tf embedding compatibility get written.

alecgunny:
- [ ] Tests for TF mh data loading
- [ ] TF multi-hot embedding layers
- [ ] Tests for TF multi-hot embedding layers
- [ ] Updating examples to remove `Scalar` prefix from embedding layers

alecgunny:
Opening up the discussion of what to do about empty rows on the data loader side. Is this something we want to have addressed before merging, or even before 0.3? @benfred

benfred:
rerun tests

jperez999 (approved):

nvtabular/framework_utils/tensorflow/layers/embedding.py
```diff
@@ -260,14 +308,18 @@ class ScalarLinearFeatures(tf.keras.layers.Layer):
     ----------
     feature_columns : list of tf.feature_column
         feature columns describing the inputs to the layer
+    sparse_aggregation : str in ("mean", "sqrtn", "sum")
+        how to aggregate embeddings from multi-valent categorical features.
+        See `tf.nn.embedding_lookup_sparse` documentation for details on
+        how this is used.
```
benfred, line 314: Is this sparse aggregation parameter used (in the base class?). 

I'd prefer not to rely on kwargs for parameters lists when possible, can we add this parameter to the list of params?

nvtabular/framework_utils/tensorflow/layers/embedding.py
```diff
@@ -307,10 +359,11 @@ def call(self, inputs):
         numeric_inputs = []
         for feature_column in self.feature_columns:
             if isinstance(feature_column, fc.NumericColumn):
-                numeric_inputs.append(inputs[feature_column.key])
+                numeric_inputs.append(_handle_continuous_feature(inputs, feature_column))
             else:
                 table = self.embedding_tables[feature_column.key]
-                x = x + tf.gather(table, inputs[feature_column.key][:, 0])
+                embeddings = _categorical_embedding_lookup(table, inputs, feature_column.key, "sum")
```
benfred, line 365: Should "sum" be self.sparse_aggregation here (assuming we set it in the constructor) ? 
```suggestion
                embeddings = _categorical_embedding_lookup(table, inputs, feature_column.key, self.sparse_aggregation)
```

alecgunny: This kwarg is removed, just removed it from the documentation too. Should have never been added, linear layers always use sum aggregation by definition

nvtabular/framework_utils/tensorflow/layers/embedding.py
alecgunny: No I added this kwarg temporarily but removed it, just pushed a change to remove it from the doc too

benfred (approved):

PR #392, opened 2020-11-02
Adding dtypes option to Workflow

albert17:
1. Adds dtypes option for Workflow in functions `apply `and `write_to_dataset`. 
    1.1. Should I also add it to `update_stats` ?

2. Should I remove the existing dtypes option from Dataset? No, it makes sense to keep it.

3. Should I remove the existing dtype option for Categorify? Not sure if it will be used, but so far we will keep it and we will see in the future.

### DIFF
--- a/nvtabular/workflow.py
+++ b/nvtabular/workflow.py
@@ -20,7 +20,6 @@ import warnings
 
 import dask
 import dask_cudf
-import numpy as np
 import yaml
 from fsspec.core import get_fs_token_paths
 
@@ -512,7 +511,7 @@ class BaseWorkflow:
 
                 start_write = time.time()
                 # Special dtype conversion
-                gdf = _set_dtypes(dtypes)
+                gdf = _set_dtypes(gdf, dtypes)
                 writer.add_data(gdf)
                 self.timings["write_df"] += time.time() - start_write
 
@@ -825,7 +824,8 @@ class Workflow(BaseWorkflow):
 
         # Iterate through dataset, apply ops, and write out processed data
         if apply_ops:
-            for gdf in dataset.to_iter(shuffle=(shuffle is not None)):
+            columns = self.columns_ctx["all"]["base"]
+            for gdf in dataset.to_iter(shuffle=(shuffle is not None), columns=columns):
                 self.apply_ops(gdf, output_path=output_path, writer=writer, dtypes=dtypes)
 
         # Close writer and write general/specialized metadata
--- a/nvtabular/workflow.py
+++ b/nvtabular/workflow.py
@@ -20,6 +20,7 @@ import warnings
 
 import dask
 import dask_cudf
+import numpy as np
 import yaml
 from fsspec.core import get_fs_token_paths
 
@@ -511,9 +512,7 @@ class BaseWorkflow:
 
                 start_write = time.time()
                 # Special dtype conversion
-                if dtypes:
-                    _meta = _set_dtypes(gdf._meta, dtypes)
-                    gdf = gdf.map_partitions(_set_dtypes, dtypes, meta=_meta)
+                gdf = _set_dtypes(dtypes)
                 writer.add_data(gdf)
                 self.timings["write_df"] += time.time() - start_write
 
@@ -761,8 +760,8 @@ class Workflow(BaseWorkflow):
             Number of IO threads to use for writing the output dataset.
             For `0` (default), no dedicated IO threads will be used.
         dtypes : dict
-            Dictionary containing desired datatypes for colums.
-            Keys are column names, values datatypes.
+            Dictionary containing desired datatypes for output columns.
+            Keys are column names, values are datatypes.
         """
 
         # Check shuffle argument
@@ -887,6 +886,12 @@ class Workflow(BaseWorkflow):
             for idx, _ in enumerate(self.phases[:end]):
                 self.exec_phase(idx, record_stats=record_stats, update_ddf=(idx == (end - 1)))
             self._base_phase = 0  # Re-Set _base_phase
+
+        if dtypes:
+            ddf = self.get_ddf()
+            _meta = _set_dtypes(ddf._meta, dtypes)
+            self.set_ddf(ddf.map_partitions(_set_dtypes, dtypes, meta=_meta))
+
         if output_format:
             output_path = output_path or "./"
             output_path = str(output_path)
@@ -896,7 +901,6 @@ class Workflow(BaseWorkflow):
                 shuffle=shuffle,
                 out_files_per_proc=out_files_per_proc,
                 num_threads=num_io_threads,
-                dtypes=dtypes,
             )
 
     def write_to_dataset(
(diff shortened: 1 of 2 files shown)


nvtabular/workflow.py
```diff
@@ -508,6 +510,10 @@ def apply_ops(self, gdf, start_phase=None, end_phase=None, writer=None, output_p
                     writer.need_cal_col_names = False
 
                 start_write = time.time()
+                # Special dtype conversion
+                if dtypes:
+                    _meta = _set_dtypes(gdf._meta, dtypes)
+                    gdf = gdf.map_partitions(_set_dtypes, dtypes, meta=_meta)
```
rjzamora, line 516: I believe `apply_ops` is used for online preprocessing, where you are not actually dealing with a `dask_cudf.DataFrame` object.  I *think* this should probably be as simple as `gdf = _set_dtypes(dtypes)`

nvtabular/workflow.py
```diff
@@ -753,6 +760,9 @@ def apply(
         num_io_threads : integer
             Number of IO threads to use for writing the output dataset.
             For `0` (default), no dedicated IO threads will be used.
+        dtypes : dict
+            Dictionary containing desired datatypes for colums.
```
rjzamora, line 764: Maybe specify that these are the desired datatypes for **output** columns?

nvtabular/workflow.py
```diff
@@ -947,6 +966,10 @@ def ddf_to_dataset(
         if output_format not in ("parquet", "hugectr"):
             raise ValueError("Only parquet/hugectr output supported with Dask.")
         ddf = self.get_ddf()
+        # Special dtype conversion
+        if dtypes:
+            _meta = _set_dtypes(ddf._meta, dtypes)
+            ddf = ddf.map_partitions(_set_dtypes, dtypes, meta=_meta)
```
rjzamora, line 972: If we are going to do a `map_partitions` call, we might as well move this up into `build_and_process_graph` in case the user is not actually writing out a dataset (but wants to use the workflow for a dataloader, or some other processing).

If we are only planning to use the `dtypes` when we write out a transformed dataset, then we should probably say so in the docstring.  Also, ff this is the case, we may want to pass the `dtypes` argument all the way down to the reader to avoid adding an extra task for every partition in ddf.

tests/unit/test_workflow.py
```diff
@@ -442,3 +442,62 @@ def test_chaining_3():
     assert all(
         x in result.columns for x in ["ad_id_count", "ad_id_clicked_sum_ctr", "ad_id_clicked_sum"]
     )
+
+
+@pytest.mark.parametrize("shuffle", [nvt.io.Shuffle.PER_WORKER, nvt.io.Shuffle.PER_PARTITION, None])
+@pytest.mark.parametrize("use_client", [True, False])
+def test_workflow_apply(client, use_client, tmpdir, shuffle):
```
rjzamora, line 449: Maybe add an `apply_online` parameter here to test the online case.

rjzamora (changes requested):
Thanks @albert17 !

This is looking good, but it seems that the `apply_online` case may be using dask when it shouldn't.  Let me know if there are questions with any of my comments.

nvtabular/workflow.py
albert17: Moved up to `build_and_process_graph `

tests/unit/test_workflow.py
albert17: My bad, it was not being covered. I have added the parameter, and I am running into an error that seems unrelated to this dtypes issue. 

 Digging into that I have found that [when calling to ](https://github.com/NVIDIA/NVTabular/blob/main/nvtabular/workflow.py#L816) `dataset.to_iter` no columns are passed, which it looks like it will never work for me.

nvtabular/workflow.py
rjzamora: Sorry @albert17 ! ! Typo on my end.  I meant to suggest `gdf = _set_dtypes(gdf, dtypes)`

albert17: Got it! Thanks!

tests/unit/test_workflow.py
albert17: @rjzamora have we notices these problem before?

tests/unit/test_workflow.py
```diff
@@ -442,3 +442,64 @@ def test_chaining_3():
...
+        dict_dtypes[col] = np.int64
+
+    processor.apply(
+        dataset,
+        apply_offline=apply_offline,
+        output_path=out_path,
+        shuffle=shuffle,
+        out_files_per_proc=out_files_per_proc,
+        dtypes=dict_dtypes,
+    )
```
rjzamora, line 500:
```suggestion
    if not apply_offline:
        processor.apply(
            dataset,
            output_format=None,
            record_stats=True,
        )

    processor.apply(
        dataset,
        apply_offline=apply_offline,
        record_stats=(apply_offline==True),
        output_path=out_path,
        shuffle=shuffle,
        out_files_per_proc=out_files_per_proc,
        dtypes=dict_dtypes,
    )
```

tests/unit/test_workflow.py
rjzamora: >@rjzamora have we noticed this problem before?

Good catch!! The code you referenced above should probably be changed to:

```python
        # Iterate through dataset, apply ops, and write out processed data
        if apply_ops:
            columns = self.columns_ctx["all"]["base"]
            for gdf in dataset.to_iter(shuffle=(shuffle is not None), columns=columns):
                self.apply_ops(gdf, output_path=output_path, writer=writer, dtypes=dtypes)
```

Along with [my other suggestion](https://github.com/NVIDIA/NVTabular/pull/392/files#r516341260) - The tests *should* pass.

albert17: Got it now!

tests/unit/test_workflow.py
albert17: Thanks. That improves the coverage

albert17:
@rjzamora all changes applied.

tests/unit/test_workflow.py
```diff
@@ -442,3 +442,71 @@ def test_chaining_3():
...
+    processor.add_cont_feature([ops.FillMissing(), ops.Clip(min_value=0), ops.LogOp()])
+    processor.add_cat_preprocess(ops.Categorify())
+
+    processor.finalize()
+    # Force dtypes
+    dict_dtypes = {}
+    for col in cont_columns:
+        dict_dtypes[col] = np.float32
+    for col in cat_columns:
+        dict_dtypes[col] = np.float32
```
rjzamora, line 489: Just checking - Is the default type different than `"float32"` for all columns? (I want to make sure we are checking that all columns are changed by the `dtypes` argument)

rjzamora (approved):
Looks good @albert17 - Thanks again for working on this!

benfred (approved):

tests/unit/test_workflow.py
albert17: All the columns were supposed to be int64, but I think it makes sense to be more explicit. I have forced in the last update some of the columns to be float64 and some other int32

Issue #383, opened 2020-10-27
[FEA] Simplify API for adding ops to a workflow


rjzamora:
**Is your feature request related to a problem? Please describe.**
After a few offline discussions (and my own personal experience), I feel that the current `add_preproc`/`add_feature` API is  causing unnecessary confusion for new users.   For example, it seems a bit confusing that `add_preprocess` **should** be used to add `DFOperator`s (which have 1+ `StatOperator` dependencies), **unless* those ops are transforming columns that do not exist in the raw dataset.  Overall, I don’t think there is a real need for two different APIs at all.

I suggest we consolidate the API to a single `add_operators` (or something like that), and deprecate the existing API in a future release.

**Describe the solution you'd like**
A consolidated API for adding **all** operator types (like `add_operators`).  Operator-dependencies can still be captured by adding a `list` of ops as a single pipeline.

**Describe alternatives you've considered**
The obvious alternative is to carry on with the existing API.  This is fine, but likley to cause confusion for new users.

benfred:
This should be handled by the API overhaul (#472)

benfred:
New API is merged ! We no longer have add_preprocess / add_feature anymore

Issue #377, opened 2020-10-23
[BUG] LambdaOp + Categorify: keyerror


PerkzZheng:
**Describe the bug**
I was trying to do feature crossing with `LambdaOp`, which created two new cols C1_C2, C3_C4.
After that, I was trying to categorify the two new cols. However, it seems to be that `categorify` op couldn't find the two new cols.

**Steps/Code to reproduce bug**
```python
proc = nvt.Workflow(cat_names=CATEGORICAL_COLUMNS,cont_names=CONTINUOUS_COLUMNS,label_name=LABEL_COLUMNS,client=client)

proc.add_cont_feature([ops.FillMissing(),ops.LogOp()])

feature_cross_list  = "C1_C2,C3_C4"
CROSS_COLUMNS = []
## Feature Crossing of C1 and C2, C3 and C4
if feature_cross_list:
    feature_pairs = [pair.split("_") for pair in feature_cross_list.split(",")]
    for pair in feature_pairs:
        col0 = pair[0]
        col1 = pair[1]
        col0_1 = col0+'_'+col1
        CROSS_COLUMNS.append(col0+'_'+col1)
        ## LambdaOp will automatically add new column with the name of col_name + "_" + op_name for differentiation
        proc.add_feature(ops.LambdaOp(op_name=col1,f=lambda col, gdf: col + gdf[col1], columns=[col0], replace=False))

freq_limit = 6
proc.add_preprocess(ops.Categorify(freq_threshold=freq_limit, columns=CATEGORICAL_COLUMNS + CROSS_COLUMNS,out_path=stats_path))
```

**Expected behavior**
It should categorify the two new cols as expected.

**Environment details (please complete the following information):**
 - Environment location: NVTabular 0.2 docker (NGC)
 - Method of NVTabular install: [conda, Docker, or from source]
   - `docker pull nvcr.io/nvidia/nvtabular:0.2`
 

**Error Message**
```python
distributed.worker - WARNING -  Compute Failed 
Function:  _top_level_groupby
args:      (            I1     I2    I3     I4    I5  ...          C23         C24        C25        
 C26 label
0          2.0   45.0   2.0   33.0   1.0  ...   -675152885 -1468599200  -10139646  -317696227   1.0  
1         55.0   <NA>  10.0   <NA>  <NA>  ...  -2036572784  -748184992  809724924  -317696227   0.0  
[...]

Issue #372, opened 2020-10-21
[FEA] Standardize a Workflow's representation of its output columns


alecgunny:
In a robust training pipeline, online preprocessing `Workflow`s will need to know the columns output by the preprocessing `Workflow`, and the model will need to know the columns output by the online `Workflow` (or the preprocessing `Workflow` if there is no online preprocessing) in order to build the appropriate input tensors. Moreover, they will need to know which columns are list-like (or sparse or multi-hot or whatever you want to call it) in order to know what _types_ of inputs to build. Creating a standard and intuitive representation of these columns that can be accessed from, saved by, and loaded to a `Workflow` will be invaluable in making pipelines simpler and more robust.

I understand that right now there is a method `Workflow.create_final_cols` that builds a representation like this. It would be good to better document what that representation is and possibly even build an object for it so that its attributes can be better codified and e.g. tab completed. I also wonder, depending on the amount of work it takes to generate this final representation, if this might be better done by a `property` like `Workflow.output_columns` that just computes this "final" representation based on whatever `Op`s are currently in the `Workflow`.

Once we have something like this, I can even envision a TensorFlow `framework_util` which can map from this representation to a set of input tensors (and even take care of constructing the `SparseTensor` multi-hot embedding inputs from `values` and `nnz` input tensors). Such a pipeline might look like:

```
input_columns = load_column_schema("path/to/columns/exported/by/preprocessing/workflow/columns.extension")
online_proc = Workflow(input_columns)
online_proc.add_cat_preprocess(HashBucket(...))

inputs = framework_utils.tensorflow.build_inputs(online_proc.output_columns) # probably return a dict of {name: tensor}
outputs = build_whatever_model(**model_params)(inputs)
model = tf.keras.Model(inputs=list(inputs.values()), outputs=outputs)
```

benfred:
We're part of the way there with the API overhaul - we now have a complete representation of the output column names at least, but we still need to be able to get the dtypes reliably.

benfred:
tracked in https://github.com/NVIDIA/NVTabular/issues/943 now

PR #393, opened 2020-11-02
various fixes for diff issues

jperez999:
This will refactor the workflow process for ingesting operators including simplified api and allowing multiple operators of same kind. Ordering will have priority this will allow chaining to follow a more user friendly convention. Reductions to phases will still be conducted before application phase. This also fixes tensorflow API 2 gpu memory usage util and band aid for torch tensor convergence issue.
#383 #377 #372


jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
Not ready... Stat Operators need to be individually verified to ensure they append new stats and not overwrite other stats created by others operators of the same type... i.e. moments.

jperez999:
Good to go now. we handle issue 401.

nvtabular/workflow.py
```diff
@@ -869,7 +1014,7 @@ def build_and_process_graph(
 
         # Reorder tasks for two-phase workflows
         # TODO: Generalize this type of optimization
-        self.reorder_tasks()
+        #         self.reorder_tasks()
```
benfred, line 1017: Lets delete if we don't need this anymore (rather than just comment out).

Also can we delete the function definition itself? I believe this is the only place reorder_tasks is called

nvtabular/workflow.py
```diff
@@ -74,6 +76,187 @@ def __init__(self, cat_names=None, cont_names=None, label_name=None, config=None
...
+            extra_cols = []
+            cur_extra_cols = [] if len(current) < 2 else current[1]
+            full_list = current[0] + cur_extra_cols
+            for col in full_list:
+                if col not in target_cols:
+                    extra_cols.append(col)
+            current = self.create_full_col_ctx_entry(op, target_cols, extra_cols, parent=parent)
+        self.reduce(self.columns_ctx["full"])
+
+    def reduce(self, full_dict):
```
benfred, line 102: can  we make it clear that these new methods on the workflow object are private - so users don't need to care about them. L

Methods like reduce/transform/analyze_placement/detect_cols_collision etc should also have a '_' by them ?

benfred (approved):
+1 otherwise

PR #404, opened 2020-11-05
Hash operator

albert17:
Avoid using cudf hashing function for integers(int32/int64)

### DIFF
--- a/nvtabular/ops/hash_bucket.py
+++ b/nvtabular/ops/hash_bucket.py
@@ -14,7 +14,9 @@
 # limitations under the License.
 #
 import cudf
+import numpy as np
 from cudf.utils.dtypes import is_list_dtype
+
 from nvtx import annotate
 
 from .categorify import _encode_list_column
@@ -122,7 +124,7 @@ class HashBucket(TransformOperator):
             new_col = f"{col}_{self._id}"
             if is_list_dtype(gdf[col].dtype):
                 encoded = _encode_list_column(gdf[col], gdf[col].list.leaves.hash_values() % nb)
-            elif gdf[col].dtype == np.int32 or  gdf[col].dtype == np.int64:
+            elif gdf[col].dtype == np.int32 or gdf[col].dtype == np.int64:
                 encoded = gdf[col] % nb
             else:
                 encoded = gdf[col].hash_values() % nb
--- a/nvtabular/ops/hash_bucket.py
+++ b/nvtabular/ops/hash_bucket.py
@@ -16,7 +16,6 @@
 import cudf
 import numpy as np
 from cudf.utils.dtypes import is_list_dtype
-
 from nvtx import annotate
 
 from .categorify import _encode_list_column
--- a/nvtabular/ops/hash_bucket.py
+++ b/nvtabular/ops/hash_bucket.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 #
 import cudf
-import numpy as np
 from cudf.utils.dtypes import is_list_dtype
 from nvtx import annotate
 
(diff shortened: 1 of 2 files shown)


nvtabular/ops/hash_bucket.py
```diff
@@ -122,6 +123,8 @@ def op_logic(self, gdf: cudf.DataFrame, target_columns: list, stats_context=None
             new_col = f"{col}_{self._id}"
             if is_list_dtype(gdf[col].dtype):
                 encoded = _encode_list_column(gdf[col], gdf[col].list.leaves.hash_values() % nb)
+            elif gdf[col].dtype == np.int32 or gdf[col].dtype == np.int64:
```
benfred, line 126: This strict dtype checking will miss a bunch of other variants (like uint8 etc). Can we switch this up? I think there are multiple ways of doing this - one way that would work is to check the 'kind' character flag to see if it's a signed/unsigned int:

```suggestion
            elif gdf[col].dtype.kind in ["u", "i"]:
```

benfred (changes requested):
This doesn't work on list elements, and needs a unittest. Can we switch this up a bit more? I think we should define a 'hash_values' function and use that in place of the cudf hash_values function.

something like
```python
def _hash_values(col):
    #  TODO (alberto) add a comment on why we are doing this and link back to the original issue
    if col.dtype.kind in ["u", "i"]:
        return col
    return col.hash_values()
```

Having this as a separate function will also help out Ronay as she incorporates some hash operations into the Categorify op

alecgunny:
Bringing this up here for posterity, but feel free to ping me Ben to discuss more since there's definitely properties of hashing that I'm unfamiliar with, but I'm not sure it makes sense to just avoid hashing ints altogether (with no ability to override this behavior). Doesn't this rely on some distributional properties of the sample integer values? E.g. if I have some unexpected periodic behavior in my data, the sampling of integers won't be uniform over N and so neither will the modulo. I can imagine use cases where that uniform sampling behavior is what you want, and it seems counterintuitive to not have it as the default in a hash op?

benfred:
> Bringing this up here for posterity, but feel free to ping me Ben to discuss more since there's definitely properties of hashing that I'm unfamiliar with, but I'm not sure it makes sense to just avoid hashing ints altogether (with no ability to override this behavior). Doesn't this rely on some distributional properties of the sample integer values? E.g. if I have some unexpected periodic behavior in my data, the sampling of integers won't be uniform over N and so neither will the modulo. I can imagine use cases where that uniform sampling behavior is what you want, and it seems counterintuitive to not have it as the default in a hash op?

The big advantages of this change is that it guarantees that there will be no collisions during the hash_values (which is pretty much why the python 'hash' function works like this for ints too  - like in python ```hash(10) == 10```).  

However - I think you bring up a good point here about the distribution since we're also taking the mod after the hash. Like if we have inputs that are all divisible by some common factor we're going to have pretty terrible outputs here if the num_buckets param also has that as a common factor (and asking people to choose a prime number for num_buckets would be putting too high a burden on our users). As an example, if all the inputs are even numbers and the num_buckets parameter is also even - we'll only use half the buckets. 

@albert17 what do you think here? Should we just change the hugectr input to do a lambdaop w/ the mod instead ?

albert17:
> > Bringing this up here for posterity, but feel free to ping me Ben to discuss more since there's definitely properties of hashing that I'm unfamiliar with, but I'm not sure it makes sense to just avoid hashing ints altogether (with no ability to override this behavior). Doesn't this rely on some distributional properties of the sample integer values? E.g. if I have some unexpected periodic behavior in my data, the sampling of integers won't be uniform over N and so neither will the modulo. I can imagine use cases where that uniform sampling behavior is what you want, and it seems counterintuitive to not have it as the default in a hash op?
> 
> The big advantages of this change is that it guarantees that there will be no collisions during the hash_values (which is pretty much why the python 'hash' function works like this for ints too - like in python `hash(10) == 10`).
> 
> However - I think you bring up a good point here about the distribution since we're also taking the mod after the hash. Like if we have inputs that are all divisible by some common factor we're going to have pretty terrible outputs here if the num_buckets param also has that as a common factor (and asking people to choose a prime number for num_buckets would be putting too high a burden on our users). As an example, if all the inputs are even numbers and the num_buckets parameter is also even - we'll only use half the buckets.
> 
> @albert17 what do you think here? Should we just change the hugectr input to do a lambdaop w/ the mod instead ?

I think that "we're going to have pretty terrible outputs here if the num_buckets param also has that as a common factor" sounds more than reasonable to not do that. 

Asking the user to introduce a primer number sounds too much, and convertng the `num_buckets ` input to the closest prime number is not an easy computation, and it could produce unexpected behaviors for the user.

If we all agree then, I will update it so the mod operation does a lambdaop too.

alecgunny:
Yeah I think doing stuff like forcing the user to use prime bucket counts for integer inputs is way too restrictive to address the use case that this op was intended for and that I think most users will want to use it for.

@albert17 that LambdaOp won't fix your problems, but doing a LambdaOp with just the modulo (without the hash) will. I guess that's ultimately my point: while all of this behavior would be clever to put here, there is a much simpler solution that doesn't overcomplicate the behavior of an op that _most_ users (especially those coming from TF) will see as distinct.

albert17:
Closing since we don't want to change the mod op behavior, we want the HugeCTR Notebook to use the operation LambdaOp(lambda col, gdf: gdf[col].hash_values() % bucket_size) instead of Mod operation.

Issue #405, opened 2020-11-05
[BUG] `KerasSequenceValidater` doesn't compute global metrics exactly


alecgunny:
`KerasSequenceValidater` computes metrics in an online fashion, using weighted averaging of batch-wise metric scores to compute the average over the validation dataset. Obviously, for metrics like AUC that measure _global_ properties, this won't be exact. With large enough batches, it should be a good enough approximation in most cases, but it still might be worth considering how to do this with some sort of online state update to compute these types of metrics exactly.

alecgunny:
It looks like Keras metrics have an `update_state` method that may solve this. I'll try to do this in a PR

PR #406, opened 2020-11-06
Exact validation computation in TensorFlow

alecgunny:
Changing `KerasSequenceValidater` to leverage metric `update_state` method in order to do exact compute of global metrics.


alecgunny:
Addresses #405

benfred (approved):

PR #408, opened 2020-11-06
Adding xdeepfm outer product layer

alecgunny:
I don't have a test yet, but the layer itself is a bit unintuitive so I'm not quite sure how to build one besides replicating the logic in the layer itself (maybe finding a ground-truth implementation to compare against?).


benfred:
rerun tests

benfred (approved):

Issue #409, opened 2020-11-06
[BUG] Outbrains example notebook is broken


benfred:
**Describe the bug**
The outbrains workflow is broken in the main branch and throws an exception like ```ValueError: The following columns were not found in the dataset...``` 

. This seems to be related to the changes in https://github.com/NVIDIA/NVTabular/pull/334 . This change uses dask_cudf.read_parquet to create the dask dataframe object, but passes the list of columns explicitly that need read in : https://github.com/NVIDIA/NVTabular/blob/8e176b2bb3888cce7bacadeec2266fcb8fd1a32f/nvtabular/io/parquet.py#L97-L100

Since the outbrains flow has several ops that generate new columns (and are included in the cat/conts) this fails with an error that these columns aren't in the dataset.

<details><summary>Click to see full error thrown by the outbrains notebook</summary>
<p>

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-13-b133e2b51cbf> in <module>
      2 valid_dataset = nvt.Dataset(OUTPUT_BUCKET_FOLDER+'valid_gdf.parquet', part_mem_fraction=0.12)
      3 
----> 4 workflow.apply(train_dataset, record_stats=True, output_path=output_train_dir, shuffle=True, out_files_per_proc=5)
      5 workflow.apply(valid_dataset, record_stats=False, output_path=output_valid_dir, shuffle=False, out_files_per_proc=5)

/rapids/notebooks/benf/NVTabular/nvtabular/workflow.py in apply(self, dataset, apply_offline, record_stats, shuffle, output_path, output_format, out_files_per_proc, num_io_threads, dtypes)
    782                 out_files_per_proc=out_files_per_proc,
    783                 num_io_threads=num_io_threads,
--> 784                 dtypes=dtypes,
    785             )
    786         else:

/rapids/notebooks/benf/NVTabular/nvtabular/workflow.py in build_and_process_graph(self, dataset, end_phase, output_path, record_stats, shuffle, output_format, out_files_per_proc, apply_ops, num_io_threads, dtypes)
    885             self._base_phase = 0  # Set _base_phase
    886             for idx, _ in enumerate(self.phases[:end]):
--> 887                 self.exec_phase(idx, record_stats=record_stats, update_ddf=(idx == (end - 1)))
    888             self._base_phase = 0  # Re-Set _base_phase
    889 

[...]

PR #410, opened 2020-11-07
Fix for workflows that generate new columns

benfred:
Generating a new column in the workflow and then using that in a future op
wasn't working properly (#409). This adds a quick fix to revert to the previous
behaviour of not limiting columns in the read_parquet call.


benfred:
This PR seems to have triggered a different bug (https://github.com/NVIDIA/NVTabular/issues/411) I put in a fix in the last commit there - though I'm not sure if we should also push for a fix in dask_cudf for this

jperez999 (approved):

PR #425, opened 2020-11-10
fix schema validation swap and removed excess sorting

jperez999:
fixes the schema validation swap of conts and cat columns... This was causing a need for an unnecessary extra sorting of continuous columns downstream.

### DIFF
--- a/nvtabular/loader/tensorflow.py
+++ b/nvtabular/loader/tensorflow.py
@@ -215,7 +215,7 @@ class KerasSequenceLoader(tf.keras.utils.Sequence, DataLoader):
         # sort the ccolumns to avoid getting incorrect output
         # (https://github.com/NVIDIA/NVTabular/issues/412)
         cat_names = _get_embedding_order(cat_names)
-        cont_names = cont_names
+        cont_names = _get_embedding_order(cont_names)
 
         assert devices is None or len(devices) == 1  # TODO: figure out multi-gpu support
         devices = devices or [0]
--- a/nvtabular/loader/tensorflow.py
+++ b/nvtabular/loader/tensorflow.py
@@ -66,7 +66,7 @@ def _validate_dataset(paths_or_dataset, batch_size, buffer_size, engine, reader_
     return Dataset(files, engine=engine, **reader_kwargs)
 
 
-def _validate_schema(feature_columns, cont_names, cat_names):
+def _validate_schema(feature_columns, cat_names, cont_names):
     _uses_feature_columns = feature_columns is not None
     _uses_explicit_schema = (cat_names is not None) or (cont_names is not None)
     if _uses_feature_columns and _uses_explicit_schema:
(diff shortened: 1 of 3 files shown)


nvtabular/loader/tensorflow.py
```diff
@@ -215,7 +215,7 @@ def __init__(
         # sort the ccolumns to avoid getting incorrect output
         # (https://github.com/NVIDIA/NVTabular/issues/412)
         cat_names = _get_embedding_order(cat_names)
-        cont_names = _get_embedding_order(cont_names)
```
benfred, line 218: I don't think we can get rid of the sorting here ... if we add a similar test for the ordering for the continuous columns, this change will cause incorrect results.


Can we revert this change and add this patch to the unittest?
```
diff --git a/tests/unit/test_tf_dataloader.py b/tests/unit/test_tf_dataloader.py
index 90dca67..cfccc27 100644
--- a/tests/unit/test_tf_dataloader.py
+++ b/tests/unit/test_tf_dataloader.py
@@ -32,12 +32,13 @@ tf_dataloader = pytest.importorskip("nvtabular.loader.tensorflow")
 
 def test_tf_catname_ordering(tmpdir):
     df = cudf.DataFrame(
-        {"cat1": [1] * 100, "cat2": [2] * 100, "cat3": [3] * 100, "label": [0] * 100}
+        {"cat1": [1] * 100, "cat2": [2] * 100, "cat3": [3] * 100, "label": [0] * 100,
+         "cont3": [3.0] * 100, "cont2": [2.0] * 100, "cont1": [1.0] * 100 }
     )
     path = os.path.join(tmpdir, "dataset.parquet")
     df.to_parquet(path)
     cat_names = ["cat3", "cat2", "cat1"]
-    cont_names = []
+    cont_names = ["cont3", "cont2", "cont1"]
     label_name = ["label"]
 
     data_itr = tf_dataloader.KerasSequenceLoader(
@@ -54,6 +55,10 @@ def test_tf_catname_ordering(tmpdir):
         assert list(X["cat2"].numpy()) == [2] * 10
         assert list(X["cat3"].numpy()) == [3] * 10
 
+        assert list(X["cont1"].numpy()) == [1.0] * 10
+        assert list(X["cont2"].numpy()) == [2.0] * 10
+        assert list(X["cont3"].numpy()) == [3.0] * 10
+
 
 # TODO: include use_columns option
 # TODO: include parts_per_chunk test
```

benfred (changes requested):

tests/unit/test_tf_dataloader.py
```diff
@@ -264,3 +268,5 @@ def test_validater(tmpdir, batch_size):
     true_auc = roc_auc_score(labels, predictions)
     estimated_auc = logs[auc_key]
     assert np.isclose(true_auc, estimated_auc, rtol=1e-6)
+
+
```
benfred, line 272: CI didn't seem to pick this up (?) - but I think you need to run black on this

benfred (approved):

PR #426, opened 2020-11-10
RecSys2020 updates

bschifferer:
RecSys2020 workflow updates after internal review


examples/recsys2020.ipynb
```diff
@@ -16,15 +16,26 @@
     "\n",
     "The [RecSys](https://recsys.acm.org/) conference is the leading data science conference for recommender systems and organizes an annual competiton in recommender systems. The [RecSys Challenge 2020](https://recsys-twitter.com/), hosted by Twitter, was about predicting interactions of ~200 mio. tweet-user pairs. NVIDIA's team scored 1st place. The team explained the solution in the [blogpost](https://medium.com/rapids-ai/winning-solution-of-recsys2020-challenge-gpu-accelerated-feature-engineering-and-training-for-cd67c5a87b1f) and published the code on [github](https://github.com/rapidsai/deeplearning/tree/main/RecSys2020).\n",
     "\n",
+    "### Downloading the dataset\n",
+    "\n",
+    "The dataset has to be downladed from the original source, provided by Twitter. You need to create an account on the [RecSys Challenge 2020 website](https://recsys-twitter.com/). Twitter needs to (manually) approve your account, which can take a few days. After your account is approved, you can download the data [here](https://recsys-twitter.com/data/show-downloads). We will use only the `training.tsv` file, as we cannot make submissions anymore.\n",
```
benfred, line 21:
```suggestion
    "The dataset has to be downloaded from the original source, provided by Twitter. You need to create an account on the [RecSys Challenge 2020 website](https://recsys-twitter.com/). Twitter needs to (manually) approve your account, which can take a few days. After your account is approved, you can download the data [here](https://recsys-twitter.com/data/show-downloads). We will use only the `training.tsv` file, as we cannot make submissions anymore.\n",
```

examples/recsys2020.ipynb
```diff
@@ -36,17 +47,19 @@
     "import glob\n",
     "import gc\n",
     "\n",
-    "import cupy\n",
-    "import cupy as cp\n",
-    "import cudf\n",
-    "import rmm\n",
-    "import dask\n",
-    "import dask_cudf\n",
+    "import cupy                # CuPy is an implementation of NumPy-compatible multi-dimensional array on GPU\n",
```
benfred, line 50: Do we need to import cupy twice (as cupy and cp)?  Can we remove this import and then change any instances of cupy to cp later?

We definitely don't need the same comment explaining what cupy is twice even if we leave the imports as is

examples/recsys2020.ipynb
```diff
@@ -300,76 +462,131 @@
...
-    "\n",
-    "1. Extract weekday from the timestamp column.\n",
-    "2. Count the number of tokens in the columns hashtags, domains, links.\n",
-    "3. Fill in na/missing values in target columns + hashtags, domains, links.\n",
-    "4. Change data type to uint32 to save memory/storage.\n",
-    "5. Apply splitmedia function.\n",
-    "6. Encode columns in continous integer to save memory. Some categorical columns contain long hashes of type String as values to preserve the privacy of the users (e.g. userId, language, etc.). Long hashes of type String requires signficant amount of memory to store. We encode/map the Strings to continous Integer to save significant memory.\n",
-    "7. Change data type to uint32 to save memory/storage."
+    "NVTabular `ops` can be registered in two different ways to a NVTabular `workflow`: Feature or Preprocess.\n",
+    "**Features are exectued before Preprocess `ops`.**\n",
```
benfred, line 466:
```suggestion
    "**Features are executed before Preprocess `ops`.**\n",
```

examples/recsys2020.ipynb
```diff
@@ -300,76 +462,131 @@
...
+    "proc.add_feature(\n",
+    "    nvt.ops.FillMissing(columns=label_name + ['hashtags', 'domains', 'links'])\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We encode columns in continous integer to save memory. Some categorical columns contain long hashes of type String as values to preserve the privacy of the users (e.g. userId, language, etc.). Long hashes of type String requires signficant amount of memory to store. We encode/map the Strings to continous Integer to save significant memory. Finally, we optimize the datatype."
```
benfred, line 540: A couple typos here . Also do we really want to go with 'continuous integer' to describe our categorical encoding? Maybe we should describe as 'small contiguous integers' instead?

```suggestion
    "We encode columns as a continuous integer to save memory. Some categorical columns contain long hashes of type String as values to preserve the privacy of the users (e.g. userId, language, etc.). Long hashes of type String requires significant amount of memory to store. We encode/map the Strings to continuous Integer to save significant memory. Finally, we optimize the datatype."
```

examples/recsys2020.ipynb
```diff
@@ -300,76 +462,131 @@
...
+   "source": [
+    "proc.add_preprocess([\n",
     "    nvt.ops.Categorify(\n",
     "        columns=['media', 'language', 'tweet_type', 'tweet_id', \n",
     "                 'a_user_id', 'b_user_id', 'hashtags', 'domains', 'links']\n",
     "    ),\n",
-    "    # Change data type to uint32 to save memory/storage\n",
     "    nvt.ops.LambdaOp(\n",
-    "        op_name='astypeint32_2',\n",
+    "        op_name='astypeint32',\n",
```
benfred, line 555: Do we even need this lambda op? Alberto has added support for setting the dtypes we write out columns as - can we use that instead? https://github.com/NVIDIA/NVTabular/pull/392

examples/recsys2020.ipynb
```diff
@@ -407,28 +624,35 @@
...
+      "part.10.parquet  part.22.parquet  part.34.parquet  part.46.parquet\r\n",
+      "part.11.parquet  part.23.parquet  part.35.parquet  part.47.parquet\r\n",
+      "part.12.parquet  part.24.parquet  part.36.parquet  part.48.parquet\r\n",
+      "part.13.parquet  part.25.parquet  part.37.parquet  part.5.parquet\r\n",
+      "part.14.parquet  part.26.parquet  part.38.parquet  part.6.parquet\r\n",
+      "part.15.parquet  part.27.parquet  part.39.parquet  part.7.parquet\r\n",
+      "part.16.parquet  part.28.parquet  part.4.parquet   part.8.parquet\r\n",
+      "part.17.parquet  part.29.parquet  part.40.parquet  part.9.parquet\r\n",
+      "part.18.parquet  part.3.parquet   part.41.parquet\r\n",
+      "part.19.parquet  part.30.parquet  part.42.parquet\r\n"
```
benfred, line 655: did you manage to get this working on windows ? (asking about the \r\n fields at the end there ). Doesn't require any code changes here, just wondering =)

examples/recsys2020.ipynb
```diff
@@ -730,7 +1060,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## 4. Train"
+    "## Training out model"
```
benfred, line 1063: This sentence reads weird to me - maybe either 'Training our model' or 'Training the model  instead' ?
```suggestion
    "## Training our model"
```

examples/recsys2020.ipynb
```diff
@@ -742,108 +1072,46 @@
...
-   "source": [
-    "train.columns"
-   ]
-  },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We use following input features:"
+    "Some columns are only used for feature engineering. Therefore, we define the columns, we want to ignore for training."
```
benfred, line 1087:
```suggestion
    "Some columns are only used for feature engineering. Therefore, we define the columns we want to ignore for training."
```

examples/recsys2020.ipynb
```diff
@@ -853,51 +1121,26 @@
...
-    "wait(valid)"
-   ]
-  },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Our experiments show that we require only 10% of the training dataset. Our feature engineering, such as TargetEncoding, uses the training datasets and leverage the information of the full dataset. The actual model training does not require all the data, anymore."
+    "Our experiments show that we require only 10% of the training dataset. Our feature engineering, such as TargetEncoding, uses the training datasets and leverage the information of the full dataset. The actual model training does not require all the data, anymore.<br><br>\n",
+    "We sample the training dataset to 10% of the size and drop all columns, which we do not want to use."
```
benfred, line 1129: Why are we sampling down to 10% ? Is this to run faster or to save memory usage?

benfred (changes requested):

examples/recsys2020.ipynb
bschifferer: Thanks, I added 'small' to it. I think the gist is that long Strings are inefficient to represent categorical values. I hope this is communicated.

examples/recsys2020.ipynb
bschifferer: Thanks, I kept the import cupy as cp

examples/recsys2020.ipynb
bschifferer: Fixed, thanks

examples/recsys2020.ipynb
bschifferer: I havent tried NVTabular on windows yet. I am sure, if this should be different on windows? The associated output, which has '\r\n' is produced by the cell !ls $BASE_DIR/preprocess. I just list the output files to see what is produced. I think running the notebook on Windows may produce it with '\n'

examples/recsys2020.ipynb
bschifferer: Thanks, fixed

examples/recsys2020.ipynb
bschifferer: Yes - during the competition we tried out 10%, 20% and 50% and we did not see an improvement in the performance by using more data. The reason is, that all the feature engineering (like Target Encoding) is done using the full dataset. Training the model on the features does not require all data anymore

examples/recsys2020.ipynb
bschifferer: Thanks, fixed

examples/recsys2020.ipynb
bschifferer: Yes, that is a good point. I updated the notebook using #392

examples/recsys2020.ipynb
```diff
@@ -109,63 +280,53 @@
...
-    "client.run(_rmm_pool)"
+    "## Preparing our dataset"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The original data format had multiple inefficiencies, resulting in requiring more disk space and memory:\n",
+    "1. The file format is `.tsv`, an uncompressed, column-separated format. The `parquet` file format stores tabular data in a column-oriented way and compresses the data. This can save significant amount of disk spaces and therefore, requires less i/o operations and is faster.\n",
```
benfred, line 311:
```suggestion
    "1. The file format is `.tsv`, an uncompressed, column-separated format. The `parquet` file format stores tabular data in a compressed column-oriented format. This saves a significant amount of disk space which results in fewer i/o operations and faster execution.\n",
```

benfred (approved):

PR #435, opened 2020-11-16
update Outbrain nb

rnyak:
The following changes are applied in this notebook wrt the NVT 0.3 docs review:

- Notebook name changed to Outbrain.ipynb
- Imports are organized
- TIMESTAMP_DELTA is described
- Longer cosine similarity definition and motivation
- TE definition and motivation
-  Smooth_ctr calculation with JoinGroupby + LambdaOP replaced with Target Encoding 
-  Model shape was changed as deep_hidden_units=[1024, 512, 256]


examples/wnd_outbrain/Outbrain.ipynb
```diff
@@ -0,0 +1,746 @@
+{
```
benfred, line 1: This looks like an entirely new notebook (and the previous E2E-nvt.ipynb still also exists in the same notebook).

Can you rename the notebook instead? I'd like to see a diff of the changes, rather than review the entire notebook from scratch.

benfred (changes requested):
With changing the notebook filename, you will also need to update the symlink in the docs

examples/wnd_outbrain/Outbrain.ipynb
```diff
@@ -439,91 +425,46 @@
...
-    "print(tf.__version__)\n",
-    "from tensorflow.python.feature_column import feature_column_v2 as fc\n",
     "\n",
     "# we can control how much memory to give tensorflow with this environment variable\n",
     "# IMPORTANT: make sure you do this before you initialize TF's runtime, otherwise\n",
     "# TF will have claimed all free GPU memory\n",
     "#os.environ['TF_MEMORY_ALLOCATION'] = \"0.8\" # fraction of free memory\n",
+    "from tensorflow.python.feature_column import feature_column_v2 as fc\n",
+    "from tensorflow.keras.mixed_precision import experimental as mixed_precision\n",
+    "from tensorflow.python.feature_column import feature_column_v2 as fc\n",
```
benfred, line 464: duplicated import:
```suggestion
    "from tensorflow.keras.mixed_precision import experimental as mixed_precision\n",
    "from tensorflow.python.feature_column import feature_column_v2 as fc\n",
```

examples/wnd_outbrain/Outbrain.ipynb
```diff
@@ -439,91 +425,46 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We can save the stats from the workflow and load it anytime, so we can run training without doing preprocessing."
+    "We select certain categorical and numerical features that are processed and generated via the NVTabular workflow to train our W&D TF model. Below, we'll create NUMERIC_COLUMNS list to feed our WnD model."
```
benfred, line 428: I think we should be consistent with our abbreviations - we should either use W&D or WnD, but not both. 
```suggestion
    "We select certain categorical and numerical features that are processed and generated via the NVTabular workflow to train our W&D TF model. Below, we'll create NUMERIC_COLUMNS list to feed our W&D model."
```

benfred (commented):
aside from a couple of minor things - this looks good

benfred (approved):

PR #437, opened 2020-11-17
Fixes binary duplicated names

albert17:
When using HugeCTR Format, we were having duplicated filenames when using dask, updated to work in the same way that parquet is workig.

I think for version 0.4, we should be improving the HugeCTR writer, to look similar to Parquet, and support not only Binary but also Norm format.

### DIFF
--- a/nvtabular/io/hugectr.py
+++ b/nvtabular/io/hugectr.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 import os
+from uuid import uuid4
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


benfred (approved):

benfred:
rerun tests

benfred:
rerun tests

nvtabular/io/hugectr.py
```diff
@@ -23,7 +23,10 @@
 class HugeCTRWriter(ThreadedWriter):
     def __init__(self, out_dir, **kwargs):
         super().__init__(out_dir, **kwargs)
-        self.data_paths = [os.path.join(out_dir, f"{i}.data") for i in range(self.num_out_files)]
+        if self.use_guid:
+            self.data_paths = [os.path.join(out_dir, f"{i}i.{guid()}.data") for i in range(self.num_out_files)]
+        else:
+            self.data_paths = [os.path.join(out_dir, f"{i}.data") for i in range(self.num_out_files)]
```
benfred, line 29: Looks like black doesn't like this : 
```suggestion
        if self.use_guid:
            self.data_paths = [
                os.path.join(out_dir, f"{i}i.{guid()}.data") for i in range(self.num_out_files)
            ]
        else:
            self.data_paths = [
                os.path.join(out_dir, f"{i}.data") for i in range(self.num_out_files)
            ]
```

also you might want to install the precommit hook here to catch this automatically:

```
#  from https://pre-commit.com/ - inside the nvtabular go:
pip install precommit
precommit install
```

benfred (changes requested):
I don't think this is working - flake8 is complaining about:
```
./nvtabular/io/hugectr.py:28:47: F821 undefined name 'guid'
```

albert17:
Sorry, CI was not working  yesterday night and I didn't realized. I have updated.

benfred:
> Sorry, CI was not working yesterday night and I didn't realized. I have updated.

No worries - installing the precommit hook will take about 1 minute and would catch issues like this though before hitting CI =) (follow the steps here https://pre-commit.com/ - the config file is already checked in).

benfred (approved):

PR #441, opened 2020-11-19
[BUGFIX] wget command in movielens

bschifferer:
Movielens notebook had a bug for the wget command. If the filename to download is equal with the filename in the output parameter, there was a naming conflict. Fixed by downloading and using a separate mv command to store it in the final folder.


benfred:
rerun tests

benfred (approved):

Issue #432, opened 2020-11-13
[BUG] Error when normalizing numeric column with negatives - ValueError: Number of Input Columns must be same replacement Dataframe


gabrielspmoreira:
**Describe the bug**
This error occurs when I include a specific float32 column (NetworkTestLatency) in the workflow and apply the `nvt.ops.Normalize()`.
The column has negatives (-1.0) and it seems the error is related to computing `sqrt()` for negatives, which is invalid.

```
/opt/conda/envs/rapids/lib/python3.7/site-packages/pandas/core/series.py:726: RuntimeWarning: invalid value encountered in sqrt
  result = getattr(ufunc, method)(*inputs, **kwargs)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<timed eval> in <module>

/nvtabular0.3/NVTabular/nvtabular/workflow.py in apply(self, dataset, apply_offline, record_stats, shuffle, output_path, output_format, out_files_per_proc, num_io_threads, dtypes)
    869                 out_files_per_proc=out_files_per_proc,
    870                 num_io_threads=num_io_threads,
--> 871                 dtypes=dtypes,
    872             )
    873         else:

/nvtabular0.3/NVTabular/nvtabular/workflow.py in build_and_process_graph(self, dataset, end_phase, output_path, record_stats, shuffle, output_format, out_files_per_proc, apply_ops, num_io_threads, dtypes)
    968             self._base_phase = 0  # Set _base_phase
    969             for idx, _ in enumerate(self.phases[:end]):
--> 970                 self.exec_phase(idx, record_stats=record_stats, update_ddf=(idx == (end - 1)))
    971             self._base_phase = 0  # Re-Set _base_phase
    972 

/nvtabular0.3/NVTabular/nvtabular/workflow.py in exec_phase(self, phase_index, record_stats, update_ddf)
    755         _ddf = self.get_ddf()
    756         if transforms:
--> 757             _ddf = self._aggregated_dask_transform(_ddf, transforms)
    758 
    759         stats = []

/nvtabular0.3/NVTabular/nvtabular/workflow.py in _aggregated_dask_transform(self, ddf, transforms)
    724         for transform in transforms:
    725             columns_ctx, cols_grp, target_cols, logic, stats_context = transform
--> 726             meta = logic(meta, columns_ctx, cols_grp, target_cols, stats_context)
    727         return ddf.map_partitions(self.__class__._aggregated_op, transforms, meta=meta)
    728 

[...]

PR #444, opened 2020-11-20
update moments.py for a normalize bug

rnyak:
This small change is done to solve the potential problem where pandas does not upcast square of sum of columns from float32 to float64. This causes `inf` value for the `var` variable,  which then results in an error when we merge the `original` gdf and `new_gdf`.

It also addresses the issue https://github.com/NVIDIA/NVTabular/issues/432.


benfred (approved):

benfred:
rerun tests

PR #450, opened 2020-11-22
AWS Test Script

albert17:
AWS Python Test script using boto3.

@benfred Issues:

1. In which directory should I locate the script? `unit`, `integration`? Temporary placed under `test/cloud`
2. I can only use NGC containers (`nvcr.io/nvidia/nvtabular:0.2`), because the ones in gitlab require VPN and it is not possible to connect to VPN in AWS)
    2.1. Because of that I can only tests on V100 machines, since NGC container does not support A100. This could be updated when there is a 0.3 container in NGC with no major problems
3. Real Blocker: I don't have permission to execute commands remotely in AWS. I have also tested with aws-cli command getting the same error. Getting the following error:
```
  File "tests/cloud/test_aws.py", line 59, in <module>
    InstanceIds=instances,
  File "/home/albertoa/.local/lib/python3.6/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/albertoa/.local/lib/python3.6/site-packages/botocore/client.py", line 676, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (AccessDeniedException) when calling the SendCommand operation: User: arn:aws:sts::843263297212:assumed-role/AWSOS-AD-Engineer/c42028bc-65e6-4990-961f-98961913fe14 is not authorized to perform: ssm:SendCommand on resource: arn:aws:ec2:us-east-1:843263297212:instance/i-09537947f08c99565
```


albert17:
Closing PR. We have decided to use Dask Cloud Support instead of scripts/terraform.

PR #466, opened 2020-11-27
Allow unary functions in LambdaOp

benfred:
The LambdaOp operator used to take a function with two arguments col (the current
column being applied against) and gdf (the entire dataframe). The majority
of use cases don't need the dataframe parameter, making using lambdaop more
onerous than needed.

Fix by allowing both the 1 argument case (column only) or 2 argument case
(column/dataframe) by using pythons inspect.signature function to figure
out how many argument the lambda takes.


bschifferer (commented):
Changes look good. Can we add a test case to use col and gdf?

benfred:
> Changes look good. Can we add a test case to use col and gdf?

Thanks! There are some tests that still take both the col/gdf , they just don't show up in this PR because the code is unchanged. For instance one of these tests is here:
https://github.com/NVIDIA/NVTabular/blob/59b515946d2cbc62277f3cf0c86f3c6b56090087/tests/unit/test_workflow.py#L457-L462

bschifferer:
> > Changes look good. Can we add a test case to use col and gdf?
> 
> Thanks! There are some tests that still take both the col/gdf , they just don't show up in this PR because the code is unchanged. For instance one of these tests is here:
> https://github.com/NVIDIA/NVTabular/blob/59b515946d2cbc62277f3cf0c86f3c6b56090087/tests/unit/test_workflow.py#L457-L462

Sorry, I checked only the LambdaOp unittests and havent seen there any.

bschifferer (approved):

PR #481, opened 2020-12-08
LogOp changed np.log to cudf.log() 

bschifferer:
LogOp used np.log() - which looks as a cpu version is used. Changed to using cudf log


bschifferer:
run tests

benfred:
rerun tests

benfred (approved):

PR #484, opened 2020-12-08
Add validate_dataset and regenerate_dataset methods

### LINKED ISSUE #447: [FEA] Parquet dataset creation/sanitation utility
NVTabular is designed with a specific **type** of dataset in mind.  Ideally the dataset will have the following characteristics:  

1. Comprises 1+ parquet files
2. Each parquet file consists of row-groups around 128MB in size
3. Each parquet file is large enough to map onto an entire `dask_cudf.DataFrame` partition.  This typically means >=1GB.
4. All parquet files should be located within a "root" directory, and that directory should contain a global `"_metadata"` file.
    - **Note**: This `"_metadata"` file allows the `dask_cudf` client to produce a DataFrame collection much faster, because all metadata can be accessed from a single file.  When this file is **not** present, the client needs to aggregate footer metadata from all files in the dataset.   Having this file also avoids the dreaded "inconsistent schema" error described in #429 , because that error only occurs when metadata is being aggregated manually (i.e. when `"_metadata"` doesn't exist).

**My suggestion is to add a dedicated `ensure_optimal_dataset` utility to NVTabular** (the name is not important).  By default, this utility would just return suggestions to the user.  For example, "Optimal file format is Parquet, please specify a path to `output_directory` to generate an optimal version of this dataset", or "Dataset is missing a `_metadata` file, please set replace_metadata_file=True to add one."  However, as hinted in the suggestion examples, the utility should also be able to generate a fresh "optimized" parquet dataset.

**Notes**:

- The recent addition of [dask#6851](https://github.com/dask/dask/pull/6851) and [cudf#6796](https://github.com/rapidsai/cudf/pull/6796) means that NVTabular will not need to implement its own logic to generate a `"_metadata"` file for an existing dataset.  However, since `dask.dataframe`/`dask_cudf.to_parquet` does not support an option to specify the desired size of each output file, NVTabular *would* need to implement its own logic to map input dataset files/chunks onto output parquet files.
- It is always possible for an input Dataset to be impossible to process in GPU memory (e.g giant compressed csv files, or giant single-row-group parquet files).  Therefore, the utility will likely need a CPU-memory path.
- It should be possible to support hive-partitioned datasets with a utility like this.

rjzamora:
Closes #447 

This PR adds two methods to the `Dataset` API: `validate_dataset` and `regenerate_dataset`.  For both methods, the actual logic is located within the underlying engine (within methods of the same name).

**`validate_dataset`**

The purpose of the `validate_dataset` method is to validate that a `Dataset` object meets the minimal requirements for efficient NVTabular processing.  The return value for the method is a `bool`, but warnings will also be raised to provide the user with feedback/suggestions.  For now, the dataset must be backed by parquet files to pass the validation.  If the user tries to "validate" a non-parquet dataset, they will just get a return value of `False` (and a warning message explaining that Parquet is the recommended format).

If the user asks to validate a parquet-backed dataset, the `ParquetDatasetEngine` will check the following:

-  The raw dataset directory should contain a global "_metadata" file.  If this file is missing, ``add_metadata_file=True`` can be passed to generate a new one automatically within `validate_dataset`.
-  If there is no _metadata file, the parquet schema must be consistent for all row-groups/files in the raw dataset. Otherwise, a new _metadata file must be generated to avoid errors at IO time.
- The row-groups should be no larger than the maximum size limit (``row_group_max_size``).
- For multi-file datasets, the files should be no smaller than the minimum size limit (``file_min_size``).

The `ParquetDatasetEngine.validate_dataset` method will suggest the user try using `regenerate_dataset` if/when any of the above checks do not pass. For example: 

```python
ds = nvt.Dataset(path, engine="parquet", part_size="1GB")
ds.validate_dataset()
```

```
/.../NVTabular/nvtabular/io/parquet.py:305: UserWarning: File part.0.parquet is smaller than the desired dataset partition size (1000000000). Consider using the regenerate_dataset utility to rewrite your dataset with a smaller number of (larger) files.
...
False
```

**`regenerate_dataset`**

If the user does not have a "valid" (optimal) dataset for NVTabular, `regenerate_dataset` can be used to produce a proper parquet dataset with appropriate/desired properties.  Although the user may specify a desired `output_format`, only `"parquet"` (the default) is supported.  The `ParquetDatasetEngine` backend will handle the following args/kwargs:

```
output_path : string
    Root directory path to use for the new (regenerated) dataset.
columns : list[string], optional
    Subset of columns to include in the regenerated dataset.
file_size : int or string, optional
    Desired size of each output file.
part_size : int or string, optional
    Desired partition size to use within regeneration algorithm.
    Note that this is effectively the size of each contiguous write
    operation in cudf.
cats : list[string], optional
    Categorical column list.
conts : list[string], optional
    Continuous column list.
labels : list[string], optional
    Label column list.
storage_options : dict, optional
    Storage-option kwargs to pass through to the `fsspec` file-system
    interface.
```

For example, an existing datset at `path` can be regenerated in the `new_path` directory with code like:

```python
ds = nvt.Dataset(path, engine="parquet", part_size="1GB")
ds.regenerate_dataset(
    new_path, file_size="10GB", cats=["id", "name"], conts=["x", "y"], labels=["label"]
)
```

Note that the `regenerate_dataset` call will write the `_metadata`, `_file_list.txt`, and `_metadata.json` files in the `new_path` directory.  Therfore, to ensure that `_metadata.json` is written "correctly," the user must specify the `cats`, `conts`, and `labels`.

**TODO**:
- [x] Add `validate_dataset`
    - [x] Add API
    - [x] Add test(s)
- [x] Add `regenerate_dataset`
    - [x] Add API
    - [x] Add test(s)
- [ ] Update Dataset docs
    - [ ] `validate_dataset`
    - [ ] `regenerate_dataset`


rjzamora:
@benfred @EvenOldridge - This PR is now "up to date."  Please let me know if there are any realistic use cases I can/should test with.

nvtabular/io/parquet.py
```diff
@@ -109,6 +132,458 @@ def to_ddf(self, columns=None):
...
+        labels=None,
+        storage_options=None,
+    ):
+        """Regenerate an NVTabular Dataset for efficient processing.
+
+        Example Usage::
+
+            dataset = Dataset("/path/to/data_pq", engine="parquet")
+            dataset.regenerate_dataset(
+                out_path, part_size="1MiB", file_size="10MiB"
```
benfred, line 353: Our recommendations here are around 128MB for part_size and 10GB or so for file_size - should we reflect that here?

nvtabular/io/parquet.py
```diff
@@ -109,6 +132,458 @@ def to_ddf(self, columns=None):
...
+                warnings.warn(msg)
+                if require_metadata_file:
+                    meta_valid = False
+
+        # Return True if we have a parquet dataset with a _metadata file (meta_valid)
+        # and the row-groups and file are appropriate sizes (size_valid)
+        return meta_valid and size_valid
+
+    @classmethod
+    def regenerate_dataset(
```
benfred, line 335: Whats the advantage of this regenerate_dataset method over something like ```nvtabular.Dataset(input_paths).to_parquet(output_paths)``` ?

rjzamora: Yeah, good question - `Dataset.to_parquet` will produce `out_files_per_proc * N` output files (where `N` is the number of worker processes).  If you are using >1 worker, it is not possible to preserve the original dataset ordering.  The problem is that `to_parquet` relies on a `ParquetWriter` object being cached on the worker in betweeen IO tasks.  We avoid this problem in `regenerate_dataset`, because  we use CPU memory for the read.  After reading, we also aggregate the output files in CPU memory, allowing us to preserve the original ordering of the data (even when written in parallel).  This is because the "regenerate" version will always use CPU memory for the read.

With the above in mind, I agree that we could probably expand `to_parquet` to cover our needs.

benfred (approved):

PR #502, opened 2020-12-16
[REVIEW] Update Outbrain nb for new api

rnyak:


rnyak:
@benfred I updated the text also. But We have two issues here:

1. If you look at the printed put accuracy, you'll see that it only prints out train auc va,ue, not `val_acc` value. Which is a strange behavior that @bschifferer is also observing. we are using docker container, nvt 0.3 image and we belive there is a bug. 
2. I manually created EMBEDDING_TABLE_SHAPES since the new emb_size_rule is giving pretty high emb_dim. Let me know if that's fine.

benfred (approved):

PR #504, opened 2020-12-17
New API doc strings for operators

bschifferer:
New API doc strings for operators:

Missing: Column Similarity and LambdaOp
Two ops require a parameter description for cat_cache

### DIFF
--- a/nvtabular/ops/target_encoding.py
+++ b/nvtabular/ops/target_encoding.py
@@ -93,8 +93,11 @@ class TargetEncoding(StatOperator):
         while low-cardinality columns can likely use `tree_width=1`.
         If passing a dict, each key and value should correspond to the column
         name and width, respectively. The default value is 8 for all columns.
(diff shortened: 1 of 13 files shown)


nvtabular/ops/target_encoding.py
```diff
@@ -94,6 +93,8 @@ class TargetEncoding(StatOperator):
         while low-cardinality columns can likely use `tree_width=1`.
         If passing a dict, each key and value should correspond to the column
         name and width, respectively. The default value is 8 for all columns.
+    cat_cache : ToDo Describe
+        Text
```
benfred, line 97: Lets take this from categorify:
```suggestion
    cat_cache : {"device", "host", "disk"} or dict
        Location to cache the list of unique categories for
        each categorical column. If passing a dict, each key and value
        should correspond to the column name and location, respectively.
        Default is "host" for all columns.
```

benfred (approved):

PR #512, opened 2020-12-18
data generator phase 1

jperez999:
builds a dataset given, user supplies number of cont columns, cat columns, cardinality list and distribution type. Also added distribution verification capability to the generator.


jperez999:
rerun tests

jperez999:
rerun tests

nvtabular/tools/data_gen.py
```diff
@@ -0,0 +1,442 @@
...
+    def __init__(self, name, dtype, distro=None):
+        self.name = name
+        self.dtype = dtype
+        self.distro = distro
+
+    def tupel(self):
+        tupel = []
+        for attr, val in self.__dict__.items():
+            tupel.append(val)
+        return tupel
```
benfred, line 336: Whats this method trying to do - and what does 'tupel' mean here (is it supposed to be 'tuple' ?).  

I think this could be more cleanly expressed like :
```suggestion
    def tupel(self):
        return list(self.__dict__.values())
```

nvtabular/tools/data_gen.py
```diff
@@ -0,0 +1,442 @@
...
+from scipy import stats
+from scipy.stats import powerlaw, uniform
+
+from ..io.dataset import Dataset
+from ..utils import device_mem_size
+
+
+class UniformDistro:
+    def create_col(self, num_rows, dtype=np.float32, min_val=0, max_val=1):
+        ser = cudf.Series(cupy.random.uniform(min_val, max_val, size=num_rows))
```
benfred, line 18: The dtype parameter for this method seems to be ignored - should this be 
```suggestion
        ser = cudf.Series(cupy.random.uniform(min_val, max_val, size=num_rows), dtype=dtype)
```

?

nvtabular/tools/data_gen.py
```diff
@@ -0,0 +1,442 @@
...
+        """
+        num_cols = len(conts_rep)
+        df = cudf.DataFrame()
+        for x in range(num_cols):
+            name, dtype, distro, min_val, max_val = conts_rep[x][:5]
+            width = conts_rep[x][-1]
+            dist = distro or self.dist
+            for y in range(width):
+                ser = dist.create_col(size, min_val=min_val, max_val=max_val)
+                ser = ser.astype(dtype)
```
benfred, line 73:
```suggestion
                ser = dist.create_col(size, min_val=min_val, max_val=max_val, dtype=dtype)
```

nvtabular/tools/data_gen.py
```diff
@@ -0,0 +1,442 @@
...
+        min_val=0,
+        max_val=1,
+        mean=None,
+        std=None,
+        per_nan=None,
+        width=1,
+        distro=None,
+    ):
+        super().__init__(name, dtype, distro)
+        #         self.dtype = dtype
```
benfred, line 353: just delete - this is set in the base class afaict
```suggestion
```

nvtabular/tools/data_gen.py
```diff
@@ -0,0 +1,442 @@
...
+        max_entry_size=None,
+        avg_entry_size=None,
+        per_nan=None,
+        multi_min=None,
+        multi_max=None,
+        multi_avg=None,
+        distro=None,
+    ):
+        super().__init__(name, dtype, distro)
+        #         self.dtype = dtype
```
benfred, line 378:
```suggestion
```

nvtabular/tools/data_gen.py
```diff
@@ -0,0 +1,442 @@
...
+        self.per_nan = None
+        self.multi_min = multi_min
+        self.multi_max = multi_max
+        self.multi_avg = multi_avg
+
+
+class LabelCol(Col):
+    def __init__(self, name, dtype, cardinality, distro=None):
+        super().__init__(name, dtype, distro)
+        #         self.dtype = dtype
```
benfred, line 392:
```suggestion
```

nvtabular/tools/data_gen.py
```diff
@@ -0,0 +1,442 @@
...
+    for section, vals in schema.items():
+        cols[section] = []
+        for col_name, val in vals.items():
+            v_dict = {"name": col_name}
+            v_dict.update(val)
+            if distros and col_name in distros:
+                dis = distros[col_name]
+                new_distr = distro_types[dis["name"]](**dis["params"])
+                v_dict.update({"distro": new_distr})
+            cols[section].append(executor[section](**v_dict).tupel())
```
benfred, line 441: So - it seems like the only use for ContCol/CatCol/LabelCol classes is to get created from a dictionary - and then get immediately called 'tupel' on and aren't used afterwards.

I think we should either remove the ContCol/CatCol/LabelCol classes - and just store the dictionary here (```cols[section].append(v_dict)```) or we should return the class and not call tupel (```cols[sections].append(executor[section](**v_dict)```).   I think either way will be safer than relying on the dictionary order of the class in create_conts/create_cats/create_cols

nvtabular/tools/data_gen.py
```diff
@@ -0,0 +1,442 @@
...
+        )
+        return cudf.Series(col)
+
+    def create_df(
+        self,
+        size,
+        conts_rep=None,
+        cats_rep=None,
+        labs_rep=None,
+        dist=PowerLawDistro(),
```
benfred, line 171: Do we need to have 3 different ways of specifying the random distribution here? Like we have the 'dist' member on the class, the 'dist' parameter on the method and the 'distro' parameter in the conts_rep  / cats_rep / labs_rep.  

Especially because it doesn't seem likethe 'dist' on this method is even used.

nvtabular/tools/data_gen.py
```diff
@@ -0,0 +1,442 @@
...
+            children=(offs, encoded),
+        )
+        return cudf.Series(col)
+
+    def create_df(
+        self,
+        size,
+        conts_rep=None,
+        cats_rep=None,
+        labs_rep=None,
```
benfred, line 170: Can we simplify the API a bit here? Rather than pass in conts_rep/cats_rep/labs_rep - can we just pass in the dataset schema and then call _get_cols_from_schema to get those 3?

benfred (approved):

PR #494, opened 2020-12-14
Uses cudf 0.17

albert17:
Errors found when using cudf 0.17:

1. `KeyError: None` in `tests/unit/test_dask_nvt.py::test_dask_workflow_api_dlrm`
    - This is due to a Dask bug. @rjzamora  has already submitted a [PR](https://github.com/dask/dask/pull/6969); and until we update to the next Dask version as a work around we can just avoid the flag `index=False` in NVTabular

2. `AssertionError: DataFrame.index` are different in `tests/unit/test_io.py::test_dask_dataset`
    - This error is happening when reading multiple parquet files. Trying to fix it adding the flag `check_index=False` since we don't care about matching indices in NVTabular. That solved the Parquet error, but creates errors for csv. @rjzamora does not get the csv errors and it works for him.

3. `assert <NA> is None` in `test_ops.py::test_difference_lag`
    - Having a frame with `dtype: bool`, we were doing the following operation `mask[mask == False] = None`. The resulting dataframe contains `<NA>` values instead of `None`. This is expected, cudf has stopped using None and instead is using `<NA>` to be coherent with pandas (`cudf.NA`).

4. `ValueError: Length mismatch: Expected axis has` in test_workflow.py, test_torch_dataloader.py, test_ops.py, and test_io.py
    - We are getting many errors like `ValueError: Length mismatch: Expected axis has 10 elements, new values have 2 elements`. @rjzamora was guessing to be a cudf bug for reading BytesIO Objects, he has raisen [an issue in cudf](https://github.com/rapidsai/cudf/issues/7011). Keith Kraus guess that it may be related to the boolean masking change if "doing something like gdf[gdf['col'] == x] where any value of gdf['col'] which is <NA> will return False for the purpose of boolean masking". 

**TODO: Fix errors**:
- [X] Fix `KeyError: None`
- [ ] Fix `AssertionError: DataFrame.index`
- [X] Fix `assert <NA> is None`
- [ ] Fix `ValueError: Length mismatch: Expected axis has`

PR #520, opened 2021-01-07
Uses cudf 0.17

albert17:
New PR after #494 was closed and cannot be reopened (new_api branch does not exist anymore).

@rjzamora for info.

Errors found when using cudf 0.17:

1. `KeyError: None` in `tests/unit/test_dask_nvt.py::test_dask_workflow_api_dlrm`
    - This is due to a Dask bug. @rjzamora  has already submitted a [PR](https://github.com/dask/dask/pull/6969); and until we update to the next Dask version as a work around we can just avoid the flag `index=False` in NVTabular

2. `AssertionError: DataFrame.index` are different in `tests/unit/test_io.py::test_dask_dataset`
    - This error is happening when reading multiple parquet files. Trying to fix it adding the flag `check_index=False` since we don't care about matching indices in NVTabular. That solved the Parquet error, but creates errors for csv.

3. `assert <NA> is None` in `test_ops.py::test_difference_lag`
    - Having a frame with `dtype: bool`, we were doing the following operation `mask[mask == False] = None`. The resulting dataframe contains `<NA>` values instead of `None`. This is expected, cudf has stopped using None and instead is using `<NA>` to be coherent with pandas (`cudf.NA`).

4. `assert False` in `test_lamdaop`

5. `ValueError: Can only compare identically-labeled DataFrame` in `test_log` 

6. `RuntimeError: cuDF failure` in `test_normalize_minmax`, `test_fill_median`, and `test_normalize`
    - `RuntimeError: cuDF failure at: /opt/conda/envs/rapids/conda-bld/libcudf_1607613794356/work/cpp/src/binaryop/binaryop.cpp:695: Column sizes don't match`

 
**TODO: Fix errors**:
- [X] Fix `KeyError: None`
- [X] Fix `AssertionError: DataFrame.index`
- [X] Fix `assert <NA> is None`
- [X] Fix `assert False`
- [X] Fix `ValueError: Can only compare identically-labeled DataFrame`
- [X] Fix `RuntimeError: cuDF`

### DIFF
--- a/tests/unit/test_ops.py
+++ b/tests/unit/test_ops.py
@@ -22,6 +22,7 @@ import numpy as np
 import pandas as pd
 import pytest
 from cudf.tests.utils import assert_eq
+from dask.dataframe import assert_eq as assert_eq_dd
 from pandas.api.types import is_integer_dtype
 
 import nvtabular as nvt
@@ -41,6 +42,7 @@ def test_normalize_minmax(tmpdir, df, dataset, gpu_memory_frac, engine, op_colum
     processor = nvtabular.Workflow(cont_features)
     processor.fit(dataset)
     new_gdf = processor.transform(dataset).to_ddf().compute()
+    new_gdf.index = df.index  # Make sure index is aligned for checks
     for col in op_columns:
         col_min = df[col].min()
         assert col_min == pytest.approx(processor.column_group.op.mins[col], 1e-2)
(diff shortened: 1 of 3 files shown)


albert17:
@benfred it works now for cudf 0.17.  

@rjzamora did most of the work, thank you very much!

tests/unit/test_ops.py
```diff
@@ -640,16 +657,16 @@ def test_difference_lag():
     processor.fit(dataset)
     new_gdf = processor.transform(dataset).to_ddf().compute()
 
-    assert new_gdf["timestamp_difference_lag_1"][0] is None
+    assert new_gdf["timestamp_difference_lag_1"][0] is cudf.NA
```
rjzamora, line 660: Perhaps you can add a `hasattr(cudf, "NA")` check for backward cudf compatibility?

albert17: That makes sense. Added. Thanks Rick!

benfred (approved):
LGTM! Thanks @albert17 and @rjzamora !

PR #521, opened 2021-01-08
Data inspect

albert17:
Opening because #510 was closed after new_api branch was merged into main.

@jperez999 I applied the feedback you told me, and I also added unit test. I have seen your data generation branch, I can follow your packaging style when you get that PR Merged.

Pending for the future:
1. Fix dask-cudf dtypes for lists: We need cudf support, there is an issue created (https://github.com/rapidsai/cudf/issues/7157)
2. When list are supported, test them in the unit tests. @benfred can I modify the [testing dataset](https://github.com/NVIDIA/NVTabular/blob/main/tests/conftest.py#L82) to add one column that is a list?


jperez999 (changes requested):
Remove all of those examples/dataset_inspector assets. put the inspector example as a comment within the command line tool when you move it into the code base. This will allow the docs to pick that up and turn it into an example within the documentation that users can reference.

albert17:
rerun tests

albert17:
rerun tests

jperez999:
rerun tests

albert17:
rerun tests

nvtabular/tools/data_gen.py
```diff
@@ -405,11 +404,13 @@ def _get_cols_from_schema(schema, distros=None):
...
             dtype:
             cardinality:
-            % NaNs:
+            per_nan:
     """
     cols = {}
-    executor = {"conts": ContCol, "cats": CatCol, "labs": LabelCol}
+    executor = {"conts": ContCol, "cats": CatCol, "labels": LabelCol}
     for section, vals in schema.items():
+        if section == "num_rows":
```
jperez999, line 412: what is this check for?

jperez999 (changes requested):

jperez999 (approved):

nvtabular/tools/dataset_inspector.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def __get_stats(self, ddf, ddf_dtypes, col, data, col_type, key_names):
+        """
+        Parameters
+        -----------
+        ddf : dask.dataframe.DataFrame
+            Dask dataframe with the data
+        ddf : dask.dataframe.DataFrame
+            Dask dataframe with the correct dtypes
+        col: string
```
benfred, line 55: Can we use the numpy docstring guide here for https://numpydoc.readthedocs.io/en/latest/format.html#method-docstrings (see parameters).
```suggestion
        col: str
```

nvtabular/tools/dataset_inspector.py
```diff
@@ -0,0 +1,191 @@
...
+            Col to process
+        data: Dictionary
```
benfred, line 57:
```suggestion
        data: dict

```

nvtabular/tools/dataset_inspector.py
```diff
@@ -0,0 +1,191 @@
...
+            Dictionary to store the output stats
+        col_type: tring
```
benfred, line 59:
```suggestion
        col_type: str

```

nvtabular/tools/dataset_inspector.py
```diff
@@ -0,0 +1,191 @@
...
+            Column type (i.e cat, cont, label)
+        key_names: Dictionary
```
benfred, line 61:
```suggestion
        key_names: dict

```

nvtabular/tools/dataset_inspector.py
```diff
@@ -0,0 +1,191 @@
...
+            data[col]["std"] = ddf[col].std().compute()
+
+        # Get percentage of nan for all
+        data[col]["per_nan"] = 100 * (1 - ddf[col].count().compute() / len(ddf[col]))
+
+    def inspect(self, path, dataset_format, columns_dict, output_file):
+        """
+        Parameters
+        -----------
+        path: str, list of str, or <dask.dataframe|cudf|pd>.DataFrame
```
benfred, line 125: Why not use a nvtabular.Dataset object here? That wraps all the functionality you need (takes dask dataframe/path/cudf dataframe/ pandas dataframe)

nvtabular/tools/dataset_inspector.py
```diff
@@ -0,0 +1,191 @@
...
+            return float(obj)
+        elif isinstance(obj, np.bool_):
+            return bool(obj)
+        elif isinstance(obj, np.ndarray):
+            return obj.tolist()
+        else:
+            return super(NpEncoder, self).default(obj)
+
+
+class DatasetInspector:
```
benfred, line 42: Is there any reason we aren't using a nvtabulat.ops.StatOperator interface here?

nvtabular/tools/dataset_inspector.py
```diff
@@ -0,0 +1,191 @@
...
+        for col in labels:
+            self.__get_stats(ddf, ddf_dtypes, col, data["labels"], "label", key_names)
+
+        # Write json file
+        with fsspec.open(output_file, "w") as outfile:
+            json.dump(data, outfile, cls=NpEncoder)
+
+        # Stop Dask Cluster
+        client.shutdown()
+        cluster.close()
```
benfred, line 191: we should use a context manager for this - to make sure we get shutdown/close on exceptions

benfred (changes requested):

nvtabular/tools/dataset_inspector.py
jperez999: So you thinking we should use the categorify operator for cardinality? Or are you saying we should create stat ops for the stats we are collecting?

nvtabular/tools/dataset_inspector.py
jperez999: Dataset is the nvtabular.io.Dataset i think based on imports.

nvtabular/tools/dataset_inspector.py
benfred: The latter.

I'm thinking that if we use a StatOperator class to collect these statistics - not only do we standardize the API, we also gain the ability to collect these statistics as part of a workflow which gives us a bunch of extra functionality for free. (like we can use these statistics for the ml ops work, we can measure dataset statistics on the output of a workflow etc).

nvtabular/tools/dataset_inspector.py
jperez999: The paths get turned into a nvt dataset. We ask for the paths so that the class can be used from command line to generate a json file.

nvtabular/tools/dataset_inspector.py
jperez999: ahh big picture type stuff... got it.

benfred: also see  https://github.com/NVIDIA/NVTabular/issues/533

nvtabular/tools/dataset_inspector.py
albert17: Sorry about that, I will use the guide.

nvtabular/tools/dataset_inspector.py
albert17: We are using the Dataset object here: `dataset = Dataset(path, engine=dataset_format)`.

nvtabular/tools/dataset_inspector.py
albert17: This is something I was thinking about, and I was discussing with Ronay because it would work great for #533 as Ben is pointing.

I am gonna create a dataset stats operator. Initially as `datastats.py `, please let me know what it would be a better name.

albert17:
@benfred I applied all the changes.

@jperez999 It looks like CI is broken.

albert17:
rerun tests

nvtabular/tools/dataset_inspector.py
benfred: Whats the advantage here to taking a path/format and then immediately converting to a Dataset? Can' we just pass the dataset object in directly?

nvtabular/tools/dataset_inspector.py
```diff
@@ -0,0 +1,128 @@
...
+            for col in columns_dict[col_type]:
+                data[col_type][col] = {}
+                data[col_type][col]["dtype"] = output[col]["dtype"]
+                if col_type != "conts":
+                    data[col_type][col]["cardinality"] = output[col]["cardinality"]
+                if col_type != "labels":
+                    data[col_type][col][key_names["min"][col_type]] = output[col]["min"]
+                    data[col_type][col][key_names["max"][col_type]] = output[col]["max"]
+                    data[col_type][col][key_names["mean"][col_type]] = output[col]["mean"]
+                if col_type == "conts":
```
benfred, line 122: Nitpick: the key_names structure is confusing - and removing it in favour of an if/elif will reduce the number of lines slightly:

```suggestion
                if col_type == "cats":
                    data[col_type][col]["min_entry_size"] = output[col]["min"]
                    data[col_type][col]["max_entry_size"] = output[col]["max"]
                    data[col_type][col]["avg_entry_size"] = output[col]["mean"]
                elif col_type == "conts":
                    data[col_type][col]["min_val"] = output[col]["min"]
                    data[col_type][col]["max_val"] = output[col]["max"]
                    data[col_type][col]["mean"] = output[col]["mean"]
```

nvtabular/ops/data_stats.py
```diff
@@ -0,0 +1,103 @@
...
+            self.col_names.append(col)
+            # Get dtype for all
+            dtype = ddf_dtypes[col].dtype
+            self.col_dtypes.append(dtype)
+
+            # Identify column type
+            if np.issubdtype(dtype, np.float):
+                col_type = "cont"
+            else:
+                col_type = "cat"
```
benfred, line 50: Can we assume that all integers are categorical columns? What if the column is an integer representing something like the users age?

Also - do we need a cat/col/label breakdown at all here? Can we just calculate different statistics based off the dtype of the column?

nvtabular/ops/data_stats.py
```diff
@@ -0,0 +1,103 @@
...
+
+            # Get Percentage of NaNs for all
+            dask_stats[col]["per_nan"] = 100 * (1 - ddf[col].count() / len(ddf[col]))
+
+        return dask_stats
+
+    def fit_finalize(self, dask_stats):
+        for i, col in enumerate(self.col_names):
+            # Add dtype
+            dask_stats[col]["dtype"] = str(self.col_dtypes[i])
```
benfred, line 82: nitpick:  I'd rather zip than enumerate here: 

```suggestion
        for col, dtype in zip(self.col_names, self.col_dtypes):
            # Add dtype
            dask_stats[col]["dtype"] = str(dtype)
```

nvtabular/ops/data_stats.py
```diff
@@ -0,0 +1,103 @@
...
+#
+import dask_cudf
+import numpy as np
+from nvtx import annotate
+
+from .operator import ColumnNames
+from .stat_operator import StatOperator
+
+
+class DataStats(StatOperator):
```
benfred, line 24: Can we rename this? Maybe something like ColumnStats ?

benfred (commented):
Couple minor things here - but aside from that looks good.

albert17:
@benfred Changes applied. 

I have added more configuration options for the Dask cluster, and I got that part out of the inspector to the script. I think it makes more sense to just pass a client to the Inspector rather than dealing inside wit this.

benfred:
rerun tests

benfred (approved):

benfred:
rerun tests

jperez999 (approved):

PR #536, opened 2021-01-27
Workflow serialization

benfred:
The existing load_stats/save_stats code only saved the calculated statistics for a workflow, but
didn't save all the operators that were used as well as the processing graph. This makes it
difficult to recreate the workflow exactly for cases like inferences

This PR adds support for saving and loading the entire workflow object. Since the operators
being applied could use lambdas or be custom operators written in jupyter notebooks - we're
using cloudpickle for serialization (cloudpickle is also used by dask).

### DIFF
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -281,15 +281,13 @@ class Categorify(StatOperator):
         for col in categories:
             self.categories[col] = categories[col]
 
-    def save(self):
-        return [self.categories, self.mh_columns]
-
-    def load(self, data):
-        self.categories, self.mh_columns = data
-
     def clear(self):
         self.categories = {}
         self.mh_columns = []
+
+    def set_storage_path(self, new_path, copy=False):
+        self.categories = _copy_storage(self.categories, self.out_path, new_path, copy=copy)
+        self.out_path = new_path
 
     @annotate("Categorify_transform", color="darkgreen", domain="nvt_python")
     def transform(self, columns: ColumnNames, gdf: cudf.DataFrame) -> cudf.DataFrame:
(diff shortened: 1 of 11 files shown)


rnyak:
@benfred I tested with an example as below. One thing I notice, we get two `categories` folder. One is created in the `cwd`, after categorify op, and the other one is after we save the `workflow` in the `output/workflow` folder.  Is that fine?
```
import os
import numpy as np
import glob
import pandas as pd
import cudf

from cudf.tests.utils import assert_eq

import nvtabular as nvt
from nvtabular import ColumnGroup, Dataset, Workflow

df = cudf.DataFrame({'author': ["User_A","User_E","User_B","User_C","User_A","User_B","User_B","User_C","User_B","User_C"],
                    'engaging_user': ["User_B", "User_B", "User_A", "User_D", "User_B", "User_k", "User_A", "User_D", "User_D", "User_D"], 
                    'platform':[3, 3, 12, 17, 3, 1, 1, 0, 1, 12]})

df_test = cudf.DataFrame({"author": ["User_K","User_M","User_M","User_d","User_K","User_K","User_B","User_C","User_B","User_C"],"engaging_user": [
"User_B", "User_b", "User_A", "User_D", "User_B", "User_c", "User_A", "User_D", "User_D", "User_b"], "platform":[3, 12, 19, 15, 3, 7, 1, 0, 0, 7]})

PREPROCESS_DIR = os.path.join("./examples/data")
PREPROCESS_DIR_TRAIN = os.path.join(PREPROCESS_DIR, 'train')
PREPROCESS_DIR_VALID = os.path.join(PREPROCESS_DIR, 'valid')

cat_names=['author', 'engaging_user', "platform"]
cat_features = cat_names >> nvt.ops.Categorify(freq_threshold={"author": 3, "engaging_user": 2, "platform": 0}, num_buckets={"author": 10, "engaging_user": 100})

# create column group where we apply TE to multiple columns at once- multi-column joint encoding.
te_cats = [['author', 'engaging_user'], 'author', 'engaging_user']

te_features = te_cats >> nvt.ops.TargetEncoding(
    "platform", 
    kfold=1, 
    p_smooth=20
)
workflow = nvt.Workflow(cat_features + te_features)

train_dataset = nvt.Dataset(df)
valid_dataset = nvt.Dataset(df_test)

workflow.fit(train_dataset)
workflow.transform(train_dataset).to_parquet(PREPROCESS_DIR_TRAIN, shuffle=False)
expected= cudf.read_parquet(PREPROCESS_DIR_TRAIN+'/*.parquet')

# save the workflow (including categorical mapping parquet files)
# and then verify we can load the saved workflow after moving the directory
out_path = os.path.join(PREPROCESS_DIR, "output", "workflow")
workflow.save(out_path)

moved_path = os.path.join(PREPROCESS_DIR, "output", "workflow2")
shutil.move(out_path, moved_path)
workflow2 = Workflow.load(moved_path)

# also check that when transforming our input we get the same results after loading
transformed = workflow2.transform(train_dataset).to_ddf().compute()
assert_eq(expected, transformed)
```

rnyak (approved):

oyilmaz-nvidia (approved):
I guess the code that generates the config.pbtxt is not included in this PR. Noticed couple of issues there. But, this PR looks good.

Let's merge this PR since we need to move forward in the inference task. But, I think we should check later if "cloudpickle.load" function moves the required data to GPU memory automatically or not. If not, we might be moving the data each time we call transform function.

benfred:
@rnyak - I've created a new issue to track the multiple copies of the categorical mappings. https://github.com/NVIDIA/NVTabular/issues/542 . This is intended right now - but I agree that we don't need to have multiple copies written out.

@oyilmaz-nvidia We don't actually store anything on the GPU in the workflow - so this should be all good .

oyilmaz-nvidia:
@benfred Maybe, keeping some of the values/tables on GPU memory can help boost the performance for the inference. Obviously, there will be limitations because of the limited amount of memory. But, let's chat about this later.

PR #540, opened 2021-01-27
Adds Cloud Storage Support

albert17:
1. Adds GCS/S3 Deleting files support to the benchmark script
2. Adds path with "/" at the end check in the input and output paths to the benchmark script
3. Adds writing into cloud support to `io/parquet.py`
4. Adds read stored stats from cloud to `worker.py`


nvtabular/io/parquet.py
```diff
@@ -693,7 +700,8 @@ def _write_pq_metadata_file(md_list, fs, path):
         metadata_path = fs.sep.join([path, "_metadata"])
         _meta = cudf.io.merge_parquet_filemetadata(md_list) if len(md_list) > 1 else md_list[0]
         with fs.open(metadata_path, "wb") as fil:
-            _meta.tofile(fil)
+            for m in _meta:
+                fil.write(m)
```
rjzamora, line 704:
```suggestion
            fil.write(bytes(_meta))
```
I think we should write the entire metadata blob at once (rather than looping over every byte).

examples/dask-nvtabular-criteo-benchmark.py
```diff
@@ -20,9 +20,11 @@
 import time
 import warnings
 
+import boto3
 import rmm
 from dask.distributed import Client, performance_report
 from dask_cuda import LocalCUDACluster
+from google.cloud import storage
```
rjzamora, line 27: `boto3` and `google.cloud` shouldn't be dependencies to run the benchmark.  Need a `try`/`except` for optional libraries.

examples/dask-nvtabular-criteo-benchmark.py
```diff
@@ -38,6 +40,39 @@ def setup_rmm_pool(client, pool_size):
     return None
 
 
+def setup_dirs(BASE_DIR, dask_workdir, output_path, stats_path):
+    # GCP Storage
+    if "gs://" in BASE_DIR:
+        # Get client and bucket
+        storage_client = storage.Client()
```
rjzamora, line 47: Here, we should raise an error if `google.cloud.storage` is not installed (and tell the user they need that library to handle `"gs://"`.  Same idea for aws below.

examples/dask-nvtabular-criteo-benchmark.py
```diff
@@ -76,16 +111,11 @@ def main(args):
         os.environ["UCX_TLS"] = UCX_TLS
 
     # Cleanup output directory
-    BASE_DIR = args.out_path
+    BASE_DIR = args.out_path[:-1] if args.out_path[-1] == "/" else args.out_path
```
rjzamora, line 114: (just a nitpick) I know the use of `BASE_DIR` is not your choice, and I this is not something I feel "strongly" about, but it seems that this should be `base_dir` throughout (since it is not reall a "constant" nor an environment variable).

rjzamora (commented):
Great work here @albert17 !  I left a few small comments.

rjzamora (approved):

benfred (approved):

Issue #543, opened 2021-01-29
[BUG] Categorify problems when using num_buckets


albert17:
@rnyak @benfred 

I am updating the Criteo-HugeCTR Notebook example to use the new API: https://github.com/NVIDIA/NVTabular/blob/764b8960f8bb66ec98d785497f62c50c1457efc8/examples/hugectr/criteo-hugectr.ipynb

Previously, I was `proc.add_cat_preprocess([ops.Categorify(out_path=stats_path), ops.LambdaOp(op_name="MOD10M", f=lambda col: col % num_buckets)])` for the categoricals. 

1. If try to use LambdaOp with the new API: `cat_features = CATEGORICAL_COLUMNS >> Categorify(out_path=stats_path) >> LambdaOp(f=lambda col: col % num_buckets)` I get an empty dict when I try to get the embeddings sizes `get_embedding_sizes(workflow)`.

2. If I try to use `num_buckets` within Categorify `cat_features = CATEGORICAL_COLUMNS >> Categorify(out_path=stats_path, num_buckets=num_buckets)` I get an error due to num_buckets being an int (it seems to be a bug not considering the int case)
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-18-671dd4884770> in <module>
----> 1 embeddings = get_embedding_sizes(workflow)
      2 print(embeddings)
/NVTA/nvtabular/ops/categorify.py in get_embedding_sizes(workflow)
    387         current = queue.pop()
    388         if current.op and hasattr(current.op, "get_embedding_sizes"):
--> 389             output.update(current.op.get_embedding_sizes(current.columns))
    390 
    391             if hasattr(current.op, "get_multihot_columns"):
/NVTA/nvtabular/ops/categorify.py in get_embedding_sizes(self, columns)
    357 
    358     def get_embedding_sizes(self, columns):
--> 359         return _get_embeddings_dask(self.categories, columns, self.num_buckets, self.freq_threshold)
    360 
    361     def get_multihot_columns(self):
/NVTA/nvtabular/ops/categorify.py in _get_embeddings_dask(paths, cat_names, buckets, freq_limit)
    414         path = paths.get(col)
    415         num_rows = cudf.io.read_parquet_metadata(path)[0] if path else 0
--> 416         if buckets and col in buckets and freq_limit[col] > 1:
    417             num_rows += buckets.get(col, 0)
    418         if buckets and col in buckets and not freq_limit[col]:
[...]

PR #544, opened 2021-01-29
[REVIEW] Fix get_emb_size bug when feeding num_buckets as an int to Categorify

rnyak:
This PR fixes the bug reported in this issue https://github.com/NVIDIA/NVTabular/issues/543.


_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-cayman
```
benfred, line 1: Did you mean to check this file in?

benfred (changes requested):
Can you add a unit test that tests out get_embedding_sizes with an int numbuckets passed to categorify?

_config.yml
rnyak: it was mistakenly generated. Idk how.

benfred (approved):

PR #550, opened 2021-02-02
Remove warnings when running unittests

benfred:
When running our unittests, we get a bunch of warnings saying things like

*  Port 8787 is already in use. Perhaps you already have a cluster running?
*  DeprecationWarning: The default dtype for empty Series will be 'object' instead of 'float64' in a future version. 
*  Environment variables with the 'NUMBAPRO' prefix are deprecated and consequently ignored,
etc

Clean these all up - either by cleaning up the root cause or by adding to the setup.cfg pytest warnings filter where that isn't possible.

### DIFF
--- a/nvtabular/tools/data_gen.py
+++ b/nvtabular/tools/data_gen.py
@@ -279,7 +279,7 @@ class DatasetGen:
 
     def get_batch(self, row_size):
         # grab max amount of gpu memory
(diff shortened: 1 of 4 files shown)


benfred:
rerun tests

jperez999 (approved):

PR #552, opened 2021-02-02
Perfomance improvments for triton python poc

benfred:
* Don't use a dask dataframe, call _transform_partition on a single dataframe instead
* Don't copy individual columns inside _transform_partition, instead select from the dataframe
* call '_merge_add_nodes' to (slightly) optimize the execution graph

With this changes the time to process a small batch of the rossmann dataset (with a hashbucket op
instead of categorify) is reduce from ~40ms to ~20ms.  Of this 20ms, about 75% is spent in
operators and the rest (5ms) is spent in remaining workflow overhead.

Also fix issue with serializing workflows - so that we continue to keep the input/output dtypes.
Instead of only serializing the columngroup, serialize the entire workflow but exclude the
dask client

### DIFF
--- a/nvtabular/column_group.py
+++ b/nvtabular/column_group.py
@@ -240,7 +240,7 @@ def _merge_add_nodes(graph):
     # lets take a copy to avoid mutating the input
     import copy
 
-    graph = copy.deepcopy(graph)
+    graph = copy.copy(graph)
 
     queue = [graph]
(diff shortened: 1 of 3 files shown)


benfred:
rerun tests

oyilmaz-nvidia (approved):
LGTM.

PR #554, opened 2021-02-03
Normalize performance improvements

benfred:
POC normalize perf improvements

### DIFF
--- a/nvtabular/ops/normalize.py
+++ b/nvtabular/ops/normalize.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 import cudf
+import cupy
 import dask_cudf
 from nvtx import annotate
 
@@ -52,18 +53,17 @@ class Normalize(StatOperator):
         return _custom_moments(ddf[columns])
 
     def fit_finalize(self, dask_stats):
-        for col in dask_stats.index:
-            self.means[col] = float(dask_stats["mean"].loc[col])
-            self.stds[col] = float(dask_stats["std"].loc[col])
+        self.means_cupy = cupy.array(dask_stats["mean"].values)
+        self.std_cupy = cupy.array(dask_stats["std"].values)
 
     @annotate("Normalize_op", color="darkgreen", domain="nvt_python")
     def transform(self, columns: ColumnNames, gdf: cudf.DataFrame) -> cudf.DataFrame:
-        new_gdf = cudf.DataFrame()
-        for name in columns:
-            if self.stds[name] > 0:
-                new_gdf[name] = (gdf[name] - self.means[name]) / (self.stds[name])
-                new_gdf[name] = new_gdf[name].astype("float32")
-        return new_gdf
+        arr = cupy.fromDlpack(gdf[columns].astype("float32").to_dlpack())
+        arr -= self.means_cupy
+        arr /= self.std_cupy
+        ret = cudf.from_dlpack(arr.toDlpack())
+        ret.columns = columns
+        return ret
 
     def clear(self):
         self.means = {}
--- a/nvtabular/ops/normalize.py
+++ b/nvtabular/ops/normalize.py
@@ -47,6 +47,7 @@ class Normalize(StatOperator):
         super().__init__()
         self.means = {}
         self.stds = {}
+        self.means_cupy = self.stds_cupy = None
 
     @annotate("Normalize_fit", color="green", domain="nvt_python")
     def fit(self, columns: ColumnNames, ddf: dask_cudf.DataFrame):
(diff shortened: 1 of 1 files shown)


nvtabular/ops/normalize.py
```diff
@@ -46,24 +47,31 @@ def __init__(self):
...
 
     @annotate("Normalize_op", color="darkgreen", domain="nvt_python")
     def transform(self, columns: ColumnNames, gdf: cudf.DataFrame) -> cudf.DataFrame:
-        new_gdf = cudf.DataFrame()
-        for name in columns:
-            if self.stds[name] > 0:
-                new_gdf[name] = (gdf[name] - self.means[name]) / (self.stds[name])
-                new_gdf[name] = new_gdf[name].astype("float32")
-        return new_gdf
+        arr = cupy.fromDlpack(gdf[columns].astype("float32").to_dlpack())
```
rjzamora, line 66: Is there a good reason to do `to_dlpack`/`fromDlpack` instad of using `values` directly? That is, doesn't this produce the same result: `arr = gdf[columns].astype("float32").values`

My preference for the `values` approach (if possible/performant) is that it should work for both `pandas` and `cudf`

nvtabular/ops/normalize.py
```diff
@@ -46,24 +47,31 @@ def __init__(self):
...
         self.stds = {}
+        self.means_cupy = self.stds_cupy = None
 
     @annotate("Normalize_fit", color="green", domain="nvt_python")
     def fit(self, columns: ColumnNames, ddf: dask_cudf.DataFrame):
         return _custom_moments(ddf[columns])
 
     def fit_finalize(self, dask_stats):
+        self.means_cupy = cupy.array(dask_stats["mean"].values)
+        self.stds_cupy = cupy.array(dask_stats["std"].values)
```
rjzamora, line 58: Perhaps we could do this lazily (and cache the result) when `transform` is first called?  This would make it easier to dispatch between cpu and gpu-specific code, because we would know the type of the gdf.

nvtabular/ops/normalize.py
```diff
@@ -46,24 +47,31 @@ def __init__(self):
...
-                new_gdf[name] = new_gdf[name].astype("float32")
-        return new_gdf
+        arr = cupy.fromDlpack(gdf[columns].astype("float32").to_dlpack())
+        arr -= self.means_cupy
+        arr /= self.stds_cupy
+        ret = cudf.from_dlpack(arr.toDlpack())
+        if isinstance(ret, cudf.Series):
+            ret = cudf.DataFrame({columns[0]: ret})
+        else:
+            ret.columns = columns
```
rjzamora, line 73: (with cpu/gpu dispatching in mind) Maybe something like this will work:  `type(gdf)(arr, columns=columns)`

rjzamora (commented):
This PR is a great idea @benfred! It definitely makes sense to batch the numerical columns together whenever possible.

Since I was just working on #561 before I looked at this, I left a few comments focused on generalizing the logic to work with both cpu and gpu backends - Let me know what you think.

nvtabular/ops/normalize.py
benfred: thats a good idea

nvtabular/ops/normalize.py
benfred: Thanks for the review!

> Is there a good reason to do to_dlpack/fromDlpack instad of using values directly? That is, doesn't this produce the same result: arr = gdf[columns].astype("float32").values


Yeah-  this code here is just a POC and I don't think we should merge as is (especially with your most recent PR that adds cpu support =).  The motivating idea here is to speed up small batch sizes (for instance when doing inference). This PR leads to about a 4x speedup in the normalize transform time by doing all the columns at once.

I agree that we should make this work with both the cpu/gpu versions - and the 'values' method you proposed is an elegant way of handling that. However, the dlpack version is about 2x faster than the .values version if you include the 'astype' cast - and about 6x faster without it:

```python
df = cudf.DataFrame({"x": [1.,2.,3.,4.,5.], "y":[10.,20.,30.,40., 50.]})

# With converting to float32:

min(timeit.Timer(lambda: df.astype("float32").values).repeat(repeat=100, number=1))
# 0.00040847109630703926
min(timeit.Timer(lambda: cupy.fromDlpack(df.astype("float32").to_dlpack())).repeat(repeat=100, number=1)) 
# 0.000211270060390234

# Without converting to float32:
min(timeit.Timer(lambda: df.values).repeat(repeat=100, number=1))
#  0.00016679614782333374

min(timeit.Timer(lambda: cupy.fromDlpack(df.to_dlpack())).repeat(repeat=100, number=1))
# 4.502898082137108e-05
```
Its *probably* still worth it do as you suggested - the total times here are small - but its something we should keep in mind.

rjzamora: Wow - Thats very interesting!  Thanks for clarifying that there is indeed an advantage of dlpack to consider here.

jperez999:
rerun tests

benfred:
we're not going this route - closing.

Issue #534, opened 2021-01-22
[FEA] Implement CPU-only version of NVTabular


rjzamora:
**Is your feature request related to a problem? Please describe.**
NVTabular currently requires that cudf be installed, and that the user has 1+ NVIDIA GPUs in their system.  This makes it difficult for users without direct/local access to GPU-equipped systems to iteratively develop and test new models/workflows.  It also makes it difficult to directly compare GPU and CPU performance.  

**Describe the solution you'd like**
All components of NVTabular should support "CPU Mode", which should be able to function without using `cudf` in any way.  An end-to-end solution likely involves the following features/improvements:

- [x] The `Dataset` API should handle a `cpu` boolean argument with a default value of `False` if `cudf` is installed (and `True` otherwise).  The underlying dataset engines should honor this option when the internal `ddf` is generated (using pandas-backed `dask.dataframe` if `cpu=True`, and `dask_cudf` otherwise). **[Covered in #556]**
- [x] `Dataset` should offer `to_gpu` and `to_cpu` methods to allow the user to explicitly move between host and device memory. **[Covered in #556]**
- [x] All operators should use a dispatching approach based on the location of the data that is passed to it.  That is, a pandas-backed partition should stay on the CPU, and a cudf-backed partition should stay on the GPU.  **This is likely to be the lions share of the work**, but it doesn't need to be done in one sweeping PR.  Once the `Dataset` work is done, the ops can be knocked out in parallel.
    - [x] Bucketize [#874]
    - [x] Categorify [#566]
    - [x] Categorify 
    - [x] Clip [#566]
    - [x] ColumnSimilarity [#880]
    - [x] ColumnSimilarity 
    - [x] DataStats [#884]
    - [x] DifferenceLag  [#884]
    - [x] Dropna [#884]
    - [x] FillMissing [#566] [#883]
    - [x] FillMedian [#566]
    - [x] Filter [#879]
    - [x] HashBucket [#884]
    - [x] HashedCross [#884]
    - [x] JoinExternal [#843]
    - [x] JoinGroupby [#877]
    - [x] LambdaOp [#883] [Mostly covered in #566]
    - [x] ListSlice 
    - [x] LogOp [#566]
    - [x] Normalize [#561]
    - [x] NormalizeMinMax [#883] [#561] 
    - [x] Rename [#883]
    - [x] TargetEncoding [#876]
[...]

PR #556, opened 2021-02-03
Add cpu argument to Dataset API

rjzamora:
Addresses the first two items in #534 

This PR introduces the `cpu` argument to the `Dataset` API (and to the `DatasetEngine` backends).  It also introduces the `to_cpu` and `to_gpu` methods.  Note that none of the operators (nor `Dataset.to_parquet`) support a pandas-backed partition at the moment, so any use of `cpu=True` or `to_cpu()` will result in a user warning.  Overall, this is just the first step towards a "CPU version" of NVTabular.


nvtabular/io/dataset.py
```diff
@@ -18,9 +18,13 @@
 import random
 import warnings
 
-import cudf
+try:
+    import cudf
+    import dask_cudf
+except ImportError:
+    cudf = None
+    dask_cudf = None
```
rjzamora, line 26: Not sure if we should wait for "full" cpu support before covering the case that cudf/dask_cudf is not installed...

benfred: Yeah - I think fully handling the case where there is no cudf/dask_cudf will require changes to almost every module here and is probably out of scope of the 0.4 release.  I think we can wait on this for the 'full' support in 0.5

benfred:
rerun tests

nvtabular/io/csv.py
```diff
@@ -36,8 +36,23 @@ def __init__(self, paths, part_size, storage_options=None, **kwargs):
...
+        if cpu:
+            if columns:
+                return dd.read_csv(self.paths, blocksize=self.part_size, **self.csv_kwargs)[columns]
+            return dd.read_csv(self.paths, blocksize=self.part_size, **self.csv_kwargs)
+        else:
+            if columns:
+                return dask_cudf.read_csv(self.paths, chunksize=self.part_size, **self.csv_kwargs)[
+                    columns
+                ]
+            return dask_cudf.read_csv(self.paths, chunksize=self.part_size, **self.csv_kwargs)
```
benfred, line 52: super minor nitpick: I think it'd be cleaner to minimize the number of branches here - maybe something like this instead?

```suggestion
        if cpu:
            ddf = dd.read_csv(self.paths, blocksize=self.part_size, **self.csv_kwargs)
        else:
            ddf = dask_cudf.read_csv(self.paths, chunksize=self.part_size, **self.csv_kwargs)
        if columns:
            ddf = ddf[columns]
        return ddf
```

nvtabular/io/parquet.py
```diff
@@ -109,18 +110,33 @@ def num_rows(self):
...
+            except ValueError:
+                return dd.read_parquet(
+                    self.paths,
+                    engine="pyarrow",
+                    columns=columns,
+                    index=None if columns is None else False,
+                    gather_statistics=False,
+                    split_row_groups=self.row_groups_per_part,
+                    storage_options=self.storage_options,
+                )
```
benfred, line 139: What's the cause of this fallback? Is this due to different versions of dask? If so - can you add a comment here explaining why this is necessary

(I'm not against having this code here - just want to make sure I understand why its there =)

benfred (approved):
lgtm

nvtabular/io/parquet.py
rjzamora: Right - Older versions of Dask will not support the `"pyarrow-dataset"` engine.  This code is indeed confusing, so I tried to clean it up a bit.

benfred (approved):

PR #558, opened 2021-02-07
Support to_parquet for cpu-backed Dataset

rjzamora:
Followup to #556 (fixes a small bug from that PR)

This focus of this PR is to update `Dataset.to_parquet` to honor the `cpu` attribute.  If the user initializes a `Dataset` with the `cpu=True` option, and then calls `to_parquet`, the algorithm will now keep the data in host memory.

This PR also:

- Fixes a small bug introduced in  #556
- Fixes a metadata-collection bug for `Shuffle.PER_WORKER` (which may have been around for a while)
- Removed `_memory_usage` (which shouldn't be necessary for cudf>=0.17)


HowItWorks.md
```diff
@@ -188,24 +188,11 @@ The `row_group_size` is the number of rows that are stored in each row group (in
...
-    size = 0
-    for col in df._data.columns:
-        if cudf.utils.dtypes.is_list_dtype(col.dtype):
-            for child in col.base_children:
-                size += child.__sizeof__()
-        else:
-            size += col._memory_usage(deep=True)
-    size += df.index.memory_usage(deep=True)
-    return size
-
```
rjzamora, line 204: I removed `_memory_usage` from the code - So, also removing here.  My understanding is that we can assume the user is on 0.18?

nvtabular/io/parquet.py
```diff
@@ -661,14 +685,63 @@ def _write_thread(self):
...
+
+    def _bytesio_to_disk(self):
+        md = {}
+        for bio, path in zip(self.data_bios, self.data_paths):
+            df = self._read_parquet(bio)
+            bio.close()
+            if self.shuffle == Shuffle.PER_WORKER:
+                df = _shuffle_df(df)
+            md[path] = self._to_parquet(df, path)
+        return md
```
rjzamora, line 705: Notice that we cannot use the original metadata for the the data written to `bio`, because the row-groups may be different when we re-write the whole file in one go.

nvtabular/io/parquet.py
```diff
@@ -677,17 +750,70 @@ def _close_writers(self):
...
+        if has_list_column:
+            # Need to check if pandas supports chunked parquet writing
+            # with list columns.
+            filename = self._get_filename(len(self.data_paths))
+            data.to_parquet(filename)
+            self.data_paths.append(filename)
+        else:
+            table = pa.Table.from_pandas(data)
+            writer = self._get_or_create_writer(idx, schema=table.schema)
+            writer.write_table(table)
```
rjzamora, line 797: Ah - We need to pass the desired number of rows in a row-group with the `row_group_size` argument here. Otherwise, we will get one large row-group per write.

rjzamora: Okay - For now, we can just use `memory_usage` on the first partition processed by the `CPUParquetWriter` to approximate a "good" `row_group_size` argument (and cache the result for subsequent writes).

nvtabular/io/shuffle.py
```diff
@@ -47,10 +49,20 @@ def _check_shuffle_arg(shuffle):
...
+    # This should be available in numpy>=1.20
+    if isinstance(df, pd.DataFrame):
+        arr = np.arange(size)
+        np.random.shuffle(arr)
+    else:
+        arr = cp.arange(size)
+        # Note that np.random.shuffle "should" Work for both gpu
+        # and cpu (via NEP-18), but it seems the cupy API is
+        # still needed here for correct behavior.
+        cp.random.shuffle(arr)
```
benfred, line 67: Yeah - it seems like this should work, but fails horribly.

I just tried it out using np.random.shuffle on a cupy array - and while it didn't throw an exception it mangled the results pretty badly:

```python
arr = cupy.arange(10)
# arr = array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

np.random.shuffle(arr)
# arr = array([0, 1, 1, 3, 1, 5, 6, 3, 1, 2]) 
```

I don't think this is a nep-18 problem though - I think its a legit bug in cupy.  For non-numpy arrays, numpy shuffle just swaps entries:

https://github.com/numpy/numpy/blob/9f12028b9453c17b72b26355fd503e512af96a5d/numpy/random/_generator.pyx#L4455-L4458

We can get the same bug happening here without involving numpy at all


``` python
x = cupy.arange(10)
# x = array([0., 1., 9., 3., 4., 5., 6., 7., 8., 9.])

# swap the first last entries
x[0], x[9] = x[9], x[0]
# x = array([9, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```

This seems to be because of https://github.com/cupy/cupy/issues/2824

benfred (approved):

nvtabular/io/parquet.py
```diff
@@ -661,14 +686,63 @@ def _write_thread(self):
...
+        fn = sink.split(self.fs.sep)[-1]
+        return df.to_parquet(sink, metadata_file_path=fn, compression=None, index=False)
+
+    def _write_table(self, idx, data, has_list_column=False):
+        if has_list_column:
+            # currently cudf doesn't support chunked parquet writers with list columns
+            # write out a new file, rather than stream multiple chunks to a single file
+            filename = self._get_filename(len(self.data_paths))
+            data.to_parquet(filename)
+            self.data_paths.append(filename)
```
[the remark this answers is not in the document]
rjzamora: I do think it is best to avoid obsolete workarounds wherever possible - I removed the `has_list_columns` logic throughout

nvtabular/io/shuffle.py
rjzamora: Whoa - good find! I added a note about the cupy bug

nvtabular/io/parquet.py
```diff
@@ -677,23 +744,101 @@ def _close_writers(self):
...
+            self.md_collectors[path][0].set_file_path(fn)
+        return self.md_collectors
+
+
+def _write_pq_metadata_file_cudf(md_list, fs, path):
     """ Converts list of parquet metadata objects into a single shared _metadata file. """
     if md_list:
         metadata_path = fs.sep.join([path, "_metadata"])
         _meta = cudf.io.merge_parquet_filemetadata(md_list) if len(md_list) > 1 else md_list[0]
+        # import pdb; pdb.set_trace()
```
benfred, line 823: We should remove this
```suggestion
```

nvtabular/io/parquet.py
```diff
@@ -677,23 +744,101 @@ def _close_writers(self):
...
         with fs.open(metadata_path, "wb") as fil:
-            _meta.tofile(fil)
+            # _meta.tofile(fil)
```
benfred, line 825:
```suggestion
```

benfred (approved):

benfred:
rerun tests

benfred:
Rerun tests

benfred:
Rerun tests

benfred:
rerun tests

nvtabular/io/parquet.py
rjzamora: Oops - Sorry!  (forgot I was testing this while reviewing Alberto's PR)

rjzamora:
Hmm - Another CI failure I am unable to reproduce locally :/

Issue #194, opened 2020-08-05
Categorify Op Improvements


EvenOldridge:
Some additional requirements for the categorify op:
 - Ability to select a fixed cardinality embedding rather than a frequency limit (same options for low frequency; mapping to a single value and hashing)
 - quotient remainder hashing (for the entire set, for a frequency limit, or for a fixed cardinality embedding)

alecgunny:
Related to #78 .

For the q/r hashing, it might make sense to add on to that implementation by adding a `hash_type` kwarg that takes `"quotient"` or`"remainder"` as possible values, that way users who want to do both can add 2 `Categorify` ops using a different value for that kwarg in each.

The problem with that, though, is that we would be `Categorify`-ing values _above_ the frequency threshold twice, which is a waste. The alternative would be to also accept a `"both"` argument or something to do both in the same op, but curious if we have the API support for the same `Op` to output 2 columns.

EvenOldridge:
I think hash type should definitely be an input option.  There are a number of hashing mechanisms beyond this that we'll want to cover so making that part compartmentalized would be good.

benfred:
Quotient remainder moved to https://github.com/NVIDIA/NVTabular/issues/573

PR #560, opened 2021-02-08
add fix_emb_size param in Categorify op for fixed embedding table size

rnyak:
This PR is created to address `Ability to select a fixed cardinality embedding rather than a frequency limit (same options for low frequency; mapping to a single value and hashing)` in the issue https://github.com/NVIDIA/NVTabular/issues/194.


rnyak:
@benfred I am having an issue with `_get_embeddings_dask` func.. It does not generate correct embeddings, when I set `max_size > len(df), and num_buckets > 1`.  I could not make it to generate correct results for this scenario.

I uploaded an example here. you can run and see.  Please also try the following params. 

```
cat_features = cat_names >> nvt.ops.Categorify(
    max_size={"author": 8, "engaging_user": 12}, num_buckets=5
)
```
I am thinking may be I need to send raw  unique cardinality values to `_get_embeddings_dask` so that, we can check if `max_size` < `len(df)` then it wont consider `num_buckets` at all.

benfred:
Hey - I think there were two issues there:

1) You were checking the max_size against the length of the dataframe - and only applying nlargest if it was smaller. this caused a problem with the num_buckets - 1 check (like with buckets we could still need to take nlargest)
2) The logic in the get_embeddings_dask was a little more complicated than I think it needed to be .

I pushed a change there that I *think* should handle both of these in that last commit - can you take a look?

rnyak:
> Hey - I think there were two issues there:
> 
> 1. You were checking the max_size against the length of the dataframe - and only applying nlargest if it was smaller. this caused a problem with the num_buckets - 1 check (like with buckets we could still need to take nlargest)
> 2. The logic in the get_embeddings_dask was a little more complicated than I think it needed to be .
> 
> I pushed a change there that I _think_ should handle both of these in that last commit - can you take a look?

@benfred thanks for your help. Unfortunately we still have the problem. if you can check the latest committed code and run the example again, you will see that we don't get the expected emb_size for  a case like `cat_features = cat_names >> nvt.ops.Categorify(max_size={"author": 12, "engaging_user": 8}, num_buckets=5)`. 

My first idea of setting `max_size` as boolean (True, False) actually would work, and it'd be simpler than this case. what do you think?

benfred:
I'm not sure I understand - with the code I pasted I get  ```{'author': (12, 16), 'engaging_user': (8, 16)}``` with ```nvt.ops.Categorify(max_size={"author": 12, "engaging_user": 8}, num_buckets=5)``` . Is that not what we should have?

rnyak:
> I'm not sure I understand - with the code I pasted I get `{'author': (12, 16), 'engaging_user': (8, 16)}` with `nvt.ops.Categorify(max_size={"author": 12, "engaging_user": 8}, num_buckets=5)` . Is that not what we should have?

No, because the `author` column has cardinality of 

> I'm not sure I understand - with the code I pasted I get `{'author': (12, 16), 'engaging_user': (8, 16)}` with `nvt.ops.Categorify(max_size={"author": 12, "engaging_user": 8}, num_buckets=5)` . Is that not what we should have?

My understanding from that was,  since, the `author` column has 8 unique values, so cardinality =8, and since the original `cardinality < max_size`,  at the end with `max_size` param which is used for fix_embedding size,  the user should get emb_size maximum 9 (8+1).  I remember we talked on that at our Monday meeting, and we decided if the original cardinalities are less than the max_size, they wont be touched, and left as they are? so in Alberto's case, if he uses `max_size` together with `num_buckets` param, he will end up with emb_size `original cardinality + num_buckets + 1` for features with only 3 cardinality.  If that's fine,  then we can proceed.

benfred:
> My understanding from that was, since, the author column has 8 unique values, so cardinality =8, and since the original cardinality < max_size, at the end with max_size param which is used for fix_embedding size, the user should get emb_size maximum 9 (8+1). I

So - I think the 'max' size for the embedding table should be specified by max_size here - so it should never be any larger than 12. So if you pass in ```max_size={"author": 12}```  that max value should be 12 here.

However there are 8 unique values, and 5 values for a hash bucket and 1 value for unknown - so the max cardinality here is really 14, so I think having the values returned as 12 is correct in this case . I think we need to keep the hash_bucket sizes here, because we don't know the distribution of unique values in the validation set - just in the train set. If you bump up the 'max_size' to a 100 - it still returns 14 which I think is correct

nvtabular/ops/categorify.py
```diff
@@ -149,6 +149,16 @@ class Categorify(StatOperator):
...
+    max_size : int or dictionary:{column: max_size_value}, default 0
+        If this param is set, freq_threshold should not be given. This will use
+        max_size value as topK value for each categorical feature. For exp, if max_size
+        is set to 1000, the first 1000 most frequent categoricals will be encoded, and the rest
+        will be mapped to a single bucket. To map the rest to a number of buckets, set
+        num_buckets param as indicated, i.e., num_buckets > 1. In that case, topK value
+        will be `max_size - num_buckets -1`. Note that num_buckets value should be smaller than the
+        max_size value. Final embedding table size will be `max_size` for each categorical feature.
+        If max_size is set as bigger than the number of unique features in a column, then each
+        feature's cardinality is considered for embedding table size.
```
benfred, line 161: * For the docstring - I think we should start with the most important thing about this (not that this doesn't work with freq_threshold)
* nitpick: if you set to max_size to 1000 only the first 999 most frequent items are used
* "for exp," -> "for example,"

```suggestion
    max_size : int or dictionary:{column: max_size_value}, default 0
        This parameter allows you to set the maximum size for an embedding table for each column.
        For example, if max_size is set to 1000 only the first 999 most frequent values for each column
        will be be encoded, and the rest will be mapped to a single value (0). To map the rest to a number
        of buckets,  you can set the num_buckets parameter > 1. In that case, topK value will be
        `max_size - num_buckets -1`.  Setting the max_size param means that freq_threshold should not 
        be given.  If the num_buckets parameter is set,  it must be smaller than the max_size value.
```

benfred (approved):

PR #562, opened 2021-02-09
[REVIEW] Add dependencies to `pyproject.toml` for security scanning

### LINKED ISSUE #471: [FEA] Use poetry to list python dependencies
We should use https://python-poetry.org/ to define python dependencies instead of listing in requirements.txt. While our primary package manager is still conda, having poetry definitions will simplify some processes around security scanning.

karlhigley:
This adds the dependencies listed in `requirements.txt` and `requirements-dev.txt` to the `pyproject.toml` file, so that they can be picked up by `poetry` during security scanning of the dependencies. The updated `pyproject.toml` file passes `poetry check`.

Resolves #471.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,32 @@
+[tool.poetry]
+name = "nvtabular"
+version = ""
+description = ""
+
+[tool.poetry.dependencies]
+python = "^3.7"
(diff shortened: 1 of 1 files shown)


benfred (approved):

benfred:
rerun tests

PR #563, opened 2021-02-10
Always generate filelist.txt

### LINKED ISSUE #524: [BUG] _file_list.txt is not written out in the validation output folder
**Describe the bug**
After we perform `workflow.transform()` both for train and validation set, we can see there is `_file_list.txt` is generated in the training output folder, but it is missing in the validation output folder.

`_file_list.txt` is required for HugeCTR training. There should be a file under train directory and another under valid directory and the file path is provided in the json file:

https://github.com/NVIDIA/NVTabular/blob/main/examples/hugectr/dlrm_fp32_64k.json#L35

**Expected behavior**
 `_file_list.txt` should be written out for validation also.

**Environment details (please complete the following information):**
 - Environment location: [Bare-metal, Docker, Cloud(specify cloud provider)] :Conda
 - Method of NVTabular install: [conda, Docker, or from source]: Conda + pip install

benfred:
Don't default to using ddf.to_parquet() - this causes some needed
dataset information like filelist.txt to not be written out. Instead
always use the same codepath for writing out parquet datasets.

For users that want to keep the dask-cudf to_parquet call - this can
still be achieved with ```dataset.to_ddf().to_parquet(path)```.

Closes #524

### DIFF
--- a/nvtabular/io/dataset.py
+++ b/nvtabular/io/dataset.py
@@ -448,32 +448,22 @@ class Dataset:
 
         fs = get_fs_token_paths(output_path)[0]
         fs.mkdirs(output_path, exist_ok=True)
-        if shuffle or out_files_per_proc or cats or conts or labels:
 
-            # Output dask_cudf DataFrame to dataset
-            _ddf_to_dataset(
-                ddf,
-                fs,
-                output_path,
-                shuffle,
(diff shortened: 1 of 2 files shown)


benfred:
rerun tests

rjzamora (approved):

PR #564, opened 2021-02-10
Fix flaky unittest

benfred:
We occasionally had unittest failures saying ```ValueError: Only one class present in y_true. ROC
AUC score is not defined in that case```. This was caused by labels being randomly generated
as all true or false in the test_validater test.  Fix by using a random state generated with a
constant seed, rather than picking randomly each run.


karlhigley (approved):
I feel like there has to be a better way than fixing the random state, but removing test flakiness seems like a high priority.

PR #566, opened 2021-02-10
Support criteo/dlrm operations with cpu-backed Dataset

rjzamora:
Establishes CPU support for all operations used in Criteo/DLRM example (i.e. `Categorify`, `LogOp`, `FillMissing`, `Clip`, `Normalize`).  Note that this contributes to the work requested in #534

### DIFF
--- a/nvtabular/dispatch.py
+++ b/nvtabular/dispatch.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import itertools
 from typing import Union
 
 import cudf
@@ -20,6 +21,7 @@ import cupy as cp
 import numpy as np
 import pandas as pd
 import pyarrow.parquet as pq
+from cudf.utils.dtypes import is_list_dtype
 
 DataFrameType = Union[pd.DataFrame, cudf.DataFrame]
 
@@ -36,6 +38,20 @@ def _series_has_nulls(s):
         return s.isnull().values.any()
     else:
         return s._column.has_nulls
+
+
+def _is_list_dtype(s):
+    if isinstance(s, pd.Series):
+        return pd.api.types.is_list_like(s)
+    else:
+        return is_list_dtype(s)
+
+
+def _flatten_list_column(s):
+    if isinstance(s, pd.Series):
+        return pd.DataFrame({s.name: itertools.chain(*s)})
+    else:
+        return cudf.DataFrame({s.name: s.list.leaves})
 
 
 def _concat_columns(args: list):
--- a/nvtabular/dispatch.py
+++ b/nvtabular/dispatch.py
@@ -39,6 +39,11 @@ def _arange(size, like_df=None):
 def _hash_series(s):
     """Row-wise Series hash"""
     if isinstance(s, pd.Series):
+        # Using pandas hashing, which does not produce the
+        # same result as cudf.Series.hash_values().  Do not
+        # expect hash-based data transformations to be the
+        # same on CPU and CPU.  TODO: Fix this (maybe use
+        # murmurhash3 manually on CPU).
         return hash_object_dispatch(s).values
     else:
         if _is_list_dtype(s):
--- a/nvtabular/dispatch.py
+++ b/nvtabular/dispatch.py
@@ -21,6 +21,7 @@ import cupy as cp
 import numpy as np
 import pandas as pd
 import pyarrow.parquet as pq
+from cudf.core.column import as_column, build_column
 from cudf.utils.dtypes import is_list_dtype
 from dask.dataframe.utils import hash_object_dispatch
 
@@ -28,6 +29,7 @@ DataFrameType = Union[pd.DataFrame, cudf.DataFrame]
 
 
 def _arange(size, like_df=None):
+    """Dispatch for numpy.arange"""
     if isinstance(like_df, pd.DataFrame):
         return np.arange(size)
     else:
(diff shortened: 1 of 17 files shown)


nvtabular/dispatch.py
```diff
@@ -13,14 +13,74 @@
...
+
+def _hash_series(s):
+    """Row-wise Series hash"""
+    if isinstance(s, pd.Series):
+        return hash_object_dispatch(s).values
+    else:
+        if _is_list_dtype(s):
+            return s.list.leaves.hash_values()
+        else:
+            return s.hash_values()
```
benfred, line 47: I might be super paranoid here - but the gpu/cpu versions here don't return the same results (different hash functions) and I'm wondering if this will cause issues down the line.

Like if we transform a dataset during training on the gpu - and then try to do the same transforms on the cpu at inference time - and the hash operations will be different so everything will break.

Should we put a warning on the CPU version for now? Ideally we could also do the same hashing operations on the cpu too (cudf does murmurhash3 iirc)

nvtabular/ops/categorify.py
```diff
@@ -649,22 +680,22 @@ def _write_uniques(dfs, base_path, col_group, on_host, concat_groups, name_sep):
         new_cols = {}
         nulls_missing = False
         for col in col_group:
-            if not df[col]._column.has_nulls:
+            if not _series_has_nulls(df[col]):
                 nulls_missing = True
                 new_cols[col] = _concat(
-                    [cudf.Series([None], dtype=df[col].dtype), df[col]],
+                    [df._constructor_sliced([None], dtype=df[col].dtype), df[col]],
```
benfred, line 686: I had no idea about this 'constructor_sliced' property - neat!

nvtabular/ops/stat_operator.py
```diff
@@ -29,7 +29,7 @@ class StatOperator(Operator):
     def __init__(self):
         super(StatOperator, self).__init__()
 
-    def fit(self, columns: ColumnNames, ddf: dask_cudf.DataFrame) -> Any:
+    def fit(self, columns: ColumnNames, ddf: dd) -> Any:
```
benfred, line 32: Is 'dd' the right type here? 
```suggestion
    def fit(self, columns: ColumnNames, ddf: dd.DataFrame) -> Any:
```

nvtabular/workflow.py
```diff
@@ -296,35 +296,46 @@ def _get_stat_ops(nodes):
...
 
 
-def _transform_partition(root_gdf, column_groups):
+def _get_unique(cols):
+    # Need to preserve order in unique-column list
+    unique_cols = []
+    for c in cols:
+        if c not in unique_cols:
+            unique_cols.append(c)
+    return unique_cols
```
benfred, line 305: Its sorta annoying that sets in python aren't ordered - but since dict's are ordered in python, maybe we can take advantage of that ? 

```suggestion
def _get_unique(cols):
    # Need to preserve order in unique-column list
    return list({x: x for x in cols}.keys())
```

benfred (commented):
thanks for this! CPU support is really coming along =)

nvtabular/workflow.py
benfred: also - I think we should probably move this 'unique' functionality to the column_group class (inside flattened_columns rather than here)

rjzamora: >dict's are ordered in python

I didn't realize dicts are ordered?

nvtabular/ops/stat_operator.py
rjzamora: Oops

nvtabular/dispatch.py
rjzamora: Ah - I was actually thinking the same thing, but forgot to leave a note.  I'll think about this a bit and report back here.

nvtabular/ops/categorify.py
rjzamora: Right?  I had two pandas devs confirm that this is the best way to do this when I was working on dask shuffle :)

nvtabular/workflow.py
benfred: yup! since python 3.7 dictionary order is defined to be the same as insertion order: https://softwaremaniacs.org/blog/2020/02/05/dicts-ordered/en/ or https://mail.python.org/pipermail/python-dev/2017-December/151283.html

rjzamora: Nice! I remember finding a serious bug related to a dict **not** being ordered back in the <=3.5 days.  Good to know :)

nvtabular/dispatch.py
rjzamora: Okay - I added a note and a warning for now.  I also tried playing with `mmh3`, but I'm having trouble reproducing cudf hash.

```python
import cudf
import numpy as np
import mmh3

def _hash(val):
    return mmh3.hash_from_buffer(np.array(val))
func = np.frompyfunc(_hash, 1, 1)

print("mmh3:", func(np.arange(2)))
print("cudf:", cudf.Series(np.arange(2)).hash_values())
```
gives...
```
mmh3: [1669671676 1392991556] 
cudf: [  29140149 -247539971] 
```

However, this is likely operator error, since I'm not really sure how to use `mmh3` "correctly"

karlhigley:
It's amazing how small these changes are. Super cool!

nvtabular/dispatch.py
benfred: I've created an issue to track https://github.com/NVIDIA/NVTabular/issues/569 - I think solving this is out of scope of this PR

benfred (approved):

PR #583, opened 2021-02-19
Set up Actions to test conda env creation

karlhigley:
These actions are scheduled to run once a day at 9:30am EST, so that we can catch breaking changes from upstream dependencies near the start of the work day.


karlhigley:
@benfred These have been tested on my fork. Splitting them into separate actions seems to make them more reliable, for whatever reason, and they now allow themselves to be manually triggered from the GH Actions interface.

karlhigley:
rerun tests

.github/workflows/conda-create-cuda10.1.yml
```diff
@@ -0,0 +1,34 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Add conda to system path
+      run: |
+        # $CONDA is an environment variable pointing to the root of the miniconda directory
+        echo $CONDA/bin >> $GITHUB_PATH
+    - name: Create Conda environment with CUDA 10.1
+      run: |
+        conda env create --file conda/environments/nvtabular_dev_cuda10.1.yml --name nvtabular_dev_10.1
```
benfred, line 34: Thanks for this!

It makes sense to run each cuda environment in its own GA job - but I'm wondering if we can cut down on the boilerplate by using a build matrix for this, rather than have as 3 separate files.

Something like:
```
  build-linux-cross:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        target: [cuda11.0, cuda10.2, cuda10.1]

```
And then changing the run step to use the build matrix

```suggestion
        conda env create --file conda/environments/nvtabular_dev_{{ matrix.target }}.yml --name nvtabular_dev_{{ matrix.target }}
```

karlhigley: Yeah, that could work. 👍

benfred (approved):

karlhigley:
rerun tests

benfred:
rerun tests

karlhigley:
rerun tests

karlhigley:
rerun tests

PR #594, opened 2021-02-23
Fix minor formatting, typo, and wording issues in the docs

karlhigley:
There are few minor edits for clarity here, but mostly this PR adjusts formatting so that code renders in code blocks and doesn't run into the right margin.

### DIFF
--- a/nvtabular/workflow.py
+++ b/nvtabular/workflow.py
@@ -158,7 +158,8 @@ class Workflow:
 
     def fit_transform(self, dataset: Dataset) -> Dataset:
         """Convenience method to both fit the workflow and transform the dataset in a single
-        call. Equivalent to calling ``workflow.fit(dataset)`` followed by ``workflow.transform(dataset)``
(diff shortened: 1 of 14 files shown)


nvtabular/workflow.py
```diff
@@ -158,7 +158,7 @@ def fit(self, dataset: Dataset):
 
     def fit_transform(self, dataset: Dataset) -> Dataset:
         """Convenience method to both fit the workflow and transform the dataset in a single
-        call. Equivalent to calling workflow.fit(dataset) followed by workflow.transform(dataset)
+        call. Equivalent to calling ``workflow.fit(dataset)`` followed by ``workflow.transform(dataset)``
```
benfred, line 161: This line is too long as is tripping up the format checker:

```suggestion
        call. Equivalent to calling ``workflow.fit(dataset)`` followed by
        ``workflow.transform(dataset)``
```

benfred (approved):
Thanks for this! looks good

PR #596, opened 2021-02-24
[REVIEW] Update Merlin Inference API TF PoC notebooks 

rnyak:
The notebooks and readme are being updated wrt the bug bash comments.


examples/inference_triton/README.md
```diff
@@ -116,6 +132,6 @@ I0216 15:28:24.027067 71 http_server.cc:2717] Started HTTPService at 0.0.0.0:800
 I0216 15:28:24.068597 71 http_server.cc:2736] Started Metrics Service at 0.0.0.0:8002
 ```
 
-All the models should show "READY" status to indicate that they loaded correctly. If a model fails to load the status will report the failure and a reason for the failure. If your model is not displayed in the table check the path to the model repository and your CUDA drivers.
+All the models should show "READY" status to indicate that they are loaded correctly. If a model fails to load the status will report the failure and a reason for the failure. If your model is not displayed in the table, check the path to the model repository and your CUDA drivers.
```
benfred, line 135: By default triton won't start if the models don't load - maybe we should add a note here?

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

examples/inference_triton/README.md
rnyak: > By default triton won't start if the models don't load - maybe we should add a note here?

Ok will add it.

benfred (approved):

PR #599, opened 2021-02-24
Update dev dependencies to run example notebooks with conda

karlhigley:
With these changes, it should be possible to create a conda env and get set up to run the notebooks from the root of the NVT repo as follows:

> conda env create -f=conda/environments/nvtabular_dev_cuda10.1.yml
> conda activate nvtabular_dev
> python -m ipykernel install --user --name=nvt
> pip install -e .
> jupyter notebook

Open an example notebook and select "nvt" as the kernel.

### DIFF
--- a/conda/environments/nvtabular_dev_cuda10.1.yml
+++ b/conda/environments/nvtabular_dev_cuda10.1.yml
@@ -9,6 +9,8 @@ dependencies:
   - clang-tools=8.0.1
   - cupy>=7.2.0,<9.0.0a
   - cudf>=0.16.*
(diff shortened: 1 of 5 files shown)


benfred (approved):
This looks good to me - we should update to rapids 0.18 now that its out (but we can do in a future PR=)

karlhigley:
As written, this already installs 0.18, but we can tighten the version specifications.

jperez999:
rerun tests

benfred:
rerun tests

benfred:
rerun tests

PR #600, opened 2021-02-24
Ignore unused columns in triton inference server

### LINKED ISSUE #598: [BUG] NVTabular workflow writes more input columns for triton inference server
**Describe the bug**
Using `generate_triton_model` generates the config file for triton inference. It depends on workflow.input_dtypes.
If my workflow does not use all available columns, then the workflow.input_dtypes contains more input columns and triton will fail loading the config.

NVTabular workflow is:
image

Current (fails):
image

Error:
```
E0224 15:58:10.330248 178 model_repository_manager.cc:963] failed to load 'amazonreview_tf' version 1: Internal: unable to create stream: the provided PTX was compiled with an unsupported toolchain.
/nvtabular/nvtabular/workflow.py:236: UserWarning: Loading workflow generated with cudf version 0+untagged.1.gbd321d1 - but we are running cudf 0.18.0a+253.g53ed28e91c. This might cause issues
  warnings.warn(
E0224 15:58:20.534884 178 model_repository_manager.cc:963] failed to load 'amazonreview_nvt' version 1: Internal: Traceback (most recent call last):
  File "/opt/tritonserver/backends/python/startup.py", line 197, in Init
    self.backend.initialize(args)
  File "/models/models/amazonreview_nvt/1/model.py", line 57, in initialize
    self.output_dtypes[name] = triton_string_to_numpy(conf["data_type"])
TypeError: 'NoneType' object is not subscriptable
I0224 15:58:20.535093 178 server.cc:490]
```

This error occurs, when triton tries to load the config for the column `brand`, which is not used in the workflow

How to fix it:
image

Use MovieLens example and do not drop the timestamp column

benfred:
We are currnetly relying on workflow.input_dtypes for figuring out the
columns that are passed to triton. This was including unused columns
from the dataframe/dataset causing issues with our triton integration.

Fix by only recording input_dtypes that are used by the workflow.

Closes #598


jperez999:
rerun tests

benfred:
rerun tests

bschifferer (commented):
I tested my workflow and it works.

bschifferer (approved):

bschifferer:
I ran my workflow with the PR and it worked

Issue #538, opened 2021-01-27
[BUG] Cudf 0.18 nvtabular unit tests failing


jperez999:
Currently 0.18 cudf/nvtabular container is reporting failures in nvtabular/tests/unit/test_dask_nvt.py. These failures are because of an error in cudf that they are aware of here https://github.com/rapidsai/cudf/issues/7199.

benfred:
fixed upstream

PR #623, opened 2021-03-01
Align s3 test with dask_cudf and workaround dask bug

rjzamora:
- This PR modifies `NVTabular/tests/unit/test_s3.py` to align closely with the s3 testing in dask_cudf.  In fact, we are now directly importing the s3 fixtures and context manager logic from dask_cudf's `test_s3.py` file.

- This PR also adds a workaround for #538 (which should be fixed by [dask#7295](https://github.com/dask/dask/pull/7295))


benfred:
rerun tests

benfred (approved):

Issue #616, opened 2021-02-27
[BUG] tritonclient package causes errors in NVT  


jperez999:
After installing the python tritonclient package using the following: `pip install nvidia-pyindex; pip install tritonclient[all]` nvtabular tests no longer succeed failing with NVRTC COMPILATION FAILURES in cudf sub libraries. See below for example failure:
```/opt/conda/envs/rapids/lib/python3.8/site-packages/cudf/core/frame.py:2486: in isnull
    return self.__class__._from_table(Frame(data, self._index))
cudf/_lib/table.pyx:44: in cudf._lib.table.Table.__init__
    ???
/opt/conda/envs/rapids/lib/python3.8/site-packages/cudf/core/column_accessor.py:59: in __init__
    self._data = OrderedColumnDict(data)
/opt/conda/envs/rapids/lib/python3.8/site-packages/cudf/core/frame.py:2484: in <genexpr>
    data_columns = (col.isnull() for col in self._columns)
/opt/conda/envs/rapids/lib/python3.8/site-packages/cudf/core/column/column.py:746: in isnull
    result = result | libcudf.unary.is_nan(self)
/opt/conda/envs/rapids/lib/python3.8/site-packages/cudf/core/column/column.py:1133: in __or__
    return self.binary_operator("or", other)
/opt/conda/envs/rapids/lib/python3.8/site-packages/cudf/core/column/numerical.py:119: in binary_operator
    return _numeric_column_binop(
/opt/conda/envs/rapids/lib/python3.8/contextlib.py:75: in inner
    return func(*args, **kwds)
/opt/conda/envs/rapids/lib/python3.8/site-packages/cudf/core/column/numerical.py:679: in _numeric_column_binop
    out = libcudf.binaryop.binaryop(lhs, rhs, op, out_dtype)
cudf/_lib/binaryop.pyx:209: in cudf._lib.binaryop.binaryop
    ???
[...]

PR #627, opened 2021-03-02
Default to using grpc tritonclient

benfred:


tests/unit/test_triton_inference.py
```diff
@@ -6,10 +6,9 @@
...
 import nvtabular.ops as ops
 
-# triton = pytest.importorskip("nvtabular.inference.triton")
+triton = pytest.importorskip("nvtabular.inference.triton")
 
 
-@pytest.mark.skip(reason="not playing nice with others")
 @pytest.mark.parametrize("engine", ["parquet"])
 def test_generate_triton_model(tmpdir, engine, df):
     import nvtabular.inference.triton as triton
```
jperez999, line 14: remove this import I added for the flake8 stuff. so triton mention below would still be tied to something... with uncommenting above we don't need to have this import here

jperez999 (changes requested):

tests/unit/test_triton_inference.py
benfred: done

jperez999 (approved):

benfred:
rerun tests

benfred:
The issue seems to be because ```tritonclient.http``` imports gevent - and gevent does something to mess up the JIT compile with cudf. Changing to default to grpc here causes the issue in https://github.com/NVIDIA/NVTabular/issues/616 to go away.

PR #660, opened 2021-03-10
Handle explicit output-file mapping in Dataset.to_parquet

rjzamora:
**TODO**: Add detailed description of changes.

### DIFF
--- a/nvtabular/io/dask.py
+++ b/nvtabular/io/dask.py
@@ -16,15 +16,15 @@
 import collections
 
 import cudf
-
 import dask
-import pyarrow as pa
 import pandas as pd
+import pyarrow as pa
 from dask.base import tokenize
 from dask.delayed import Delayed
 from dask.highlevelgraph import HighLevelGraph
 from nvtx import annotate
 
+from nvtabular.dispatch import _to_arrow
 from nvtabular.worker import clean_worker_cache, get_worker_cache
 
 from .shuffle import Shuffle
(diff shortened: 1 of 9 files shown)


rjzamora:
**Update**: As I am finding the Criteo conversion time to be >20% slower than the custom script, I would like to come up with a way to avoid the `to_arrow` step (the cause of the regression) before this PR is integrated.

rjzamora:
**Update 2**: Latest changes avoid the need for in-memory aggregation of partitions for each output file.  Instead, we are essentially dividing the graph of the underlying `ddf` into a distinct subgraph for each output file.  Each subgraph is then executed, and the resulting data written to disk, within a single task.  So, we are passing the subgraphs needed to generate each partition as elements within the graph that is actually processed by the scheduler (Dask-ception if you will).  This approach results in great performance, but it also requires the use of `cloudpickle` to reliably package the subgraphs as task arguments.

nvtabular/io/dask.py
```diff
@@ -28,6 +28,30 @@
 from .writer_factory import _writer_cls_factory, writer_factory
 
 
+class DaskSubgraph:
+    """Simple containter for a Dask subgraph
```
benfred, line 32:
```suggestion
    """Simple container for a Dask subgraph
```

nvtabular/io/dataset.py
```diff
@@ -423,9 +440,25 @@ def to_parquet(
             data processed by each worker.  To improve performace, this option
             currently uses host-memory `BytesIO` objects for the intermediate
             persist stage. The `FULL` option is not yet implemented.
+        preserve_files : bool
+            Whether to preserve the original file-to-partition mapping of
+            the base dataset. This option is only avilable if the base
```
benfred, line 445:
```suggestion
            the base dataset. This option is only available if the base
```

benfred (approved):

PR #661, opened 2021-03-11
Avoid full map_partitions for list-dtype query

rjzamora:
While running the `dask-nvtabular-criteo-benchmark.py` benchmark today, I noticed that the performance degraded dramatically after the API overhaul.  More specifically, the default settings do not even work for a DGX-1 machine with 8x32GB-V100 GPUs.  As far as I can tell, the problem is caused by the new `map_partitions` call in `Categorify`.  This operation is checking every partition to see if there are list-column dtypes, when it is probably sufficient to check a single partition.  This PR changes the problematic `map_partitions` operation to target the first partition only (which seems to be much faster, and results in significantly less spilling).


benfred (approved):

benfred:
rerun tests

PR #670, opened 2021-03-23
Allow transformed columns in JoinGroupby

benfred:
The JoinGroupby op previously could only handle calculating stats on 'cont_cols'
that were unchanged from the original dataset. This change makes it so that
you can use arbitrary columns - including ones that have had some transformations
applied already.

### DIFF
--- a/nvtabular/ops/join_groupby.py
+++ b/nvtabular/ops/join_groupby.py
@@ -18,6 +18,7 @@ import cupy
 import dask_cudf
 from dask.delayed import Delayed
 
+import nvtabular as nvt
 from nvtabular.dispatch import _read_parquet_dispatch
 
 from . import categorify as nvt_cat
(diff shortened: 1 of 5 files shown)


benfred:
rerun tests

nvtabular/ops/join_groupby.py
```diff
@@ -39,14 +40,14 @@ class JoinGroupby(StatOperator):
...
         # Use JoinGroupby to define a NVTabular workflow
         groupby_features = ['cat1', 'cat2', 'cat3'] >> ops.JoinGroupby(
-            out_path=str(tmpdir), stats=['sum','count'], cont_names=['num1']
+            out_path=str(tmpdir), stats=['sum','count'], cont_cols=['num1']
         )
         processor = nvtabular.Workflow(groupby_features)
 
     Parameters
     -----------
-    cont_names : list of str
```
bschifferer, line 48: Why do we rename cont_names to cont_cols?

bschifferer (commented):
Do we need a unit test to capture the edge case?

nvtabular/ops/join_groupby.py
benfred: I'm expanding what it does here - its not just the column names anymore, but can alos be a columngroup or transformed columns now.

benfred:
> Do we need a unit test to capture the edge case?

I've added a unittest as part of this PR (using JoinGroupBy with a normalized continuous column) in [test_joingroupby_dependency](https://github.com/NVIDIA/NVTabular/pull/670/files#diff-e4d18001a1d407273320606db51ffe8ef20709522220cdc6e3cde2070f06b246R696). Is there another test case you

bschifferer:
> > Do we need a unit test to capture the edge case?
> 
> I've added a unittest as part of this PR (using JoinGroupBy with a normalized continuous column) in [test_joingroupby_dependency](https://github.com/NVIDIA/NVTabular/pull/670/files#diff-e4d18001a1d407273320606db51ffe8ef20709522220cdc6e3cde2070f06b246R696). Is there another test case you have in mind?

Ok - sorry didnt see that one :)

bschifferer (approved):

PR #672, opened 2021-03-23
Simplify criteo conversion script

rjzamora:
Leverages recent improvements in #660 to simplify the Criteo-conversion example.

**TODO**:  

- [x] Establish if we need a simple mechanism/option to add a `".parquet"` (or custom) file suffix when `preserve_files=True` or an integer is specified for `output_files`.  **Update**: #675 added the `suffix=` argument to `to_parquet` with a `".parquet"` default.  So, this concern has been addressed.

PR #675, opened 2021-03-24
Add suffix= argument to to_parquet (and writers)

rjzamora:
Adds a simple feature allowing the user to override the default `".parquet"` and `".data"` file extensions for parquet and hugectr output, respectively.  The primary modivation was the `Dataset.to_parquet` method - That method currently allows the the user to specify `preserve_files=True`, but there is no way for them to specify if the file extension should be changed (and how).  With this PR, the extension will be replaced with the value of `suffix` (defaulting to ".parquet" for parquet), unless the user has already specified explicit names with the `output_files` argument.

**Note**: This feature simplifies conversion code in #672 a bit.

### DIFF
--- a/examples/scaling-criteo/01-Download-Convert.ipynb
+++ b/examples/scaling-criteo/01-Download-Convert.ipynb
@@ -257,7 +257,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.9"
+   "version": "3.7.8"
   }
  },
  "nbformat": 4,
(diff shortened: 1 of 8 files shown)


examples/scaling-criteo/01-Download-Convert.ipynb
```diff
@@ -257,7 +257,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.8"
+   "version": "3.7.9"
```
rjzamora, line 260:
```suggestion
   "version": "3.7.8"
```

rjzamora:
rerun tests

benfred (approved):

PR #676, opened 2021-03-26
tf fix for test abs path

jperez999:
This PR fixes the relative paths problem in the test_hvd unit test. So now there is logic that allows running the test from any directory.


jperez999:
rerun tests

jperez999:
rerun tests

benfred (approved):

Issue #641, opened 2021-03-08
[FEA] Sequential / Session-based recommendation and time series support - Group by sorting values by timestamp


gabrielspmoreira:
**Is your feature request related to a problem? Please describe.**
In order to support **session-based / sequence recommendation** and also **time series use cases**, we need NVTabular to be able to group data by some columns (e.g. user id, session id), sorted by another column (e.g. usually timestamp), and aggregate other columns with these different aggregation functions that takes order into account: '**list**', '**first**', '**last**'

**Describe the solution you'd like**
```python
groupby_features = ['user_id', 'session_id', 'timestamp', 'prod_id', 'prod_categ'] >> ops.Groupby(
    groupby_cols=['user_id', 'session_id'], sort_column=['timestamp'], 
    aggregated_cols={'prod_id': 'list',
                   'prod_categ': 'list',
                   'timestamp': ['first', 'last']}
)
processor = nvtabular.Workflow(groupby_features)
```

**Describe alternatives you've considered**
It might be ok is some NVT ops do not support the resulting list columns, like `Categorify` and `Standardize`, because those ops can be done before the grouping.
But the `LambdaOp` should support list columns to allow, for example, to extract the length of the lists or truncate the lists.

**Additional context**
This feature can be accomplished by different data frame frameworks:

***Pandas***
In this example, the preliminary sorting of the rows is respected by the group by, so that the aggregated columns ('item_id') will be sorted by timestamp.

```python
interactions_df.sort_values('timestamp', inplace=True)
sessions_df = interactions_df.groupby('session_id').agg('item_id': 'list')
```
P.s. `cudf / dask_cudf` also supports the 'list' aggregation function. But it does not guarantee that the data frame ordering will be respected, as pandas does.

***PySpark***
```python
import pyspark.sql.functions as F

# This window ensures the sorting of aggregated columns by another column: timestamp
session_window = Window.partitionBy('user_session').orderBy('timestamp')

df.select('user_session',                        
                        F.collect_list('product_idx').over(session_window).alias('product_seq'), 
                        F.first('event_time_ts').over(session_window).alias('session_start_ts'),
                        F.last('event_time_ts').over(session_window).alias('session_end_ts'),
                        ) \
                .groupBy('user_session').agg(
                    # The F.max() is necessary for the list column because F.collect_list() will create an incremental
                    # list of values for each value, sorted by time, and F.max() returns the longest list (containing all items of the sequence).
[...]

Issue #397, opened 2020-11-03
[BUG] Flaky unittest "Cannot safely cast non-equivalent float to int64"


benfred:
**Describe the bug**
unittests ocassionally fail with 'Cannot safely cast non-equivalent float to int64' 

```
__________________________ test_kill_dl[parquet-0.1] ___________________________

tmpdir = local('/tmp/pytest-of-jenkins/pytest-7/test_kill_dl_parquet_0_1_0')
df =      name-cat name-string    id  label         x         y
0     Charlie       Edith  1024   1054  0.763430 -0.231628
...ay   929    972 -0.219996 -0.200242
2160   Ursula         Ray  1045    979 -0.482353  0.136629

[4321 rows x 6 columns]
dataset = <nvtabular.io.dataset.Dataset object at 0x7fd290391a90>
part_mem_fraction = 0.1, engine = 'parquet'

    @pytest.mark.parametrize("part_mem_fraction", [0.000001, 0.1])
    @pytest.mark.parametrize("engine", ["parquet"])
    def test_kill_dl(tmpdir, df, dataset, part_mem_fraction, engine):
        cat_names = ["name-cat", "name-string"]
        cont_names = ["x", "y", "id"]
        label_name = ["label"]
    
        processor = nvt.Workflow(cat_names=cat_names, cont_names=cont_names, label_name=label_name)
    
        processor.add_feature([ops.FillMedian()])
        processor.add_preprocess(ops.Normalize())
        processor.add_preprocess(ops.Categorify())
    
        output_train = os.path.join(tmpdir, "train/")
        os.mkdir(output_train)
    
        processor.apply(
            dataset,
            apply_offline=True,
            record_stats=True,
            shuffle=nvt.io.Shuffle.PER_PARTITION,
>           output_path=output_train,
        )

tests/unit/test_torch_dataloader.py:183: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
nvtabular/workflow.py:784: in apply
    dtypes=dtypes,
nvtabular/workflow.py:887: in build_and_process_graph
    self.exec_phase(idx, record_stats=record_stats, update_ddf=(idx == (end - 1)))
nvtabular/workflow.py:635: in exec_phase
    _ddf = self._aggregated_dask_transform(_ddf, transforms)
[...]

PR #677, opened 2021-03-26
Handle hive-partitioning in NVTabular.dataset.Dataset

### LINKED ISSUE #642: [FEA] Partition output parquet files by a column
For a realistic temporal evaluation protocol, it is common to split users interactions/sessions by time windows (e.g.  hours, days, weeks,  months), so that models can be trained over a time window and evaluated for a future time window.

So, NVTabular should be able to export the parquet files partitioned by a column (which can be a feature extracted from a timestamp column by a `LambdaOp`, like hour, day, week or month)

When the workflow exports the parquet files, they should be organized in folders named after the partition column values (e.g. "interaction_date=2021-03-05").
That is, each folder should contain parquet files whose rows have the same value for the partition column as in their folder name
```python
workflow.transform(dataset).to_parquet('base/path', partition_by='interaction_date')
```

**Additional context**
With PySpark, this can be accomplished as follows

```python
interactions_df.write.partitionBy('session_start_date') \
                     .parquet('base/path')
```
P.s. This issue was extracted from #355  , which is broader in scope, so that it is implemented independently.

rjzamora:
Closes #642 
Addresses global shuffle component of #641 

Thu purpose of this PR is to improve handling of hive-partitioned parquet data in NVTabular.  Since the Dataset API already uses `dask.dataframe.read_parquet`, there is currenlty no "correcness" issue with **reading** hive-partitioned data.  However, (1) there is no convenient mechanism to write hive-partitioned data, and (2) the read stage typically results in many small partitions (rather than a single partition for each input directory).

- **Solution to (1)**:  The `Dataset.to_parquet` method now supports a `partition_on=` argument.  This is designed to match the same option in `dask.dataframe`/`dask_cudf`.  If the user passes a list of 1+ columns with this argument, the output data will be shuffled at IO time into a distinct directory for each unique combination of those `partition_on` column values.  When multiple columns are use for partitioning (e.g. `["month", "day"]`), the directory structure is nested (so that the full path for an output file will look something like `"/month=Mar/day=30/part.0.parquet"`).
- **Solution to (2)**:  Since #641 will need a mechanism to ensure a unique mapping between specified column groups and ddf partitions, this PR adds a `Dataset.partition_by_keys` method to perform a global shuffle on the specified column group (`keys`) and return a new (shuffled) Dataset.  For general Dataset objects, this method will simply call `ddf.shuffle()` under the hood.  For Dataset objects that are backed by hive-partitioned data, however, we use the metadata stored in the file paths to avoid a full shuffle. In the future, this optimization can be pushed even further by directly agregating all IO tasks within the same hive-partition.  However, I suspect that shuch an optimization should be implemented in `dask.dataframe`.

**Example Usage**
```python
import pandas as pd
import dask.dataframe as dd
import dask
import nvtabular as nvt

path = "fake.data"

# Create a sample ddf
ddf = dask.datasets.timeseries(
    start="2000-01-01",
    end="2000-01-03",
    freq="600s",
    partition_freq="6h",
    seed=42,
).reset_index()
ddf['timestamp'] = ddf['timestamp'].dt.round('D').dt.day

# Convert to a Datset and write out hive-partitioned data to disk
keys = ["timestamp", "name"]
nvt.Dataset(ddf).to_parquet(path, partition_on=keys)
```

This will produce a directory structure like:
```
$ find fake.data/ -type d -print
fake.data/
fake.data/timestamp=1
fake.data/timestamp=1/name=Alice
fake.data/timestamp=1/name=Frank
fake.data/timestamp=1/name=Victor
fake.data/timestamp=1/name=George
fake.data/timestamp=1/name=Quinn
fake.data/timestamp=1/name=Kevin
fake.data/timestamp=1/name=Ursula
...
```

Then, you can read the data back in with NVT, and ensure that the ddf partitions are shuffled by `keys`:
```python
ds = nvt.Dataset(path, engine="parquet").shuffle_by_keys(keys)
ds.to_ddf().compute()
```
```
      id         x         y timestamp    name
0    991 -0.750009 -0.587392         1   Alice
1   1022  0.866823 -0.682096         1   Alice
2    991  0.467775  0.683211         1   Alice
3    967  0.534984 -0.931405         1     Bob
4    991 -0.149716 -0.651939         1     Bob
..   ...       ...       ...       ...     ...
25   964  0.843602  0.598580         3  Yvonne
26   961  0.853070 -0.987596         3  Yvonne
27   947  0.934162  0.190069         3  Yvonne
28  1024 -0.107280  0.662606         3  Yvonne
29  1006  0.169090 -0.784889         3   Zelda

[288 rows x 5 columns]
```


rjzamora:
cc @gabrielspmoreira @benfred

gabrielspmoreira:
Sounds great @rjzamora ! This PR will allow incremental training and evaluation of sequential recommender models and time series, as allows splitting data by time windows.

rjzamora:
Is it possible that the dataloader CI failures are being caused by this PR?

benfred:
> Is it possible that the dataloader CI failures are being caused by this PR?

I don't think so - we've had some flaky tests around this for a while now (https://github.com/NVIDIA/NVTabular/issues/397) , but for some reason these errors seem more common now

benfred:
rerun tests

benfred:
rerun tests

benfred (approved):

rjzamora:
rerun tests

karlhigley:
@jperez999 This CI error looks familiar and I think you tried to explain it to me, but I didn't fully understand what's causing it. Thoughts on how to resolve it?

jperez999:
@karlhigley ok so I think... and follow me on this. So the reason for the error is that the median stat operator collects a median value of NA for the continuous columns (sometimes both x and y, sometimes just one of them). When we go to apply the NA value we hit the error we are seeing. I think this is because of the way we setup the dataset and the beginning here https://github.com/NVIDIA/NVTabular/blob/main/tests/conftest.py#L108-L112 This inserts the NA values, and I think we are seeing the random positions coincidentally land on the median indexes.

jperez999:
rerun tests

benfred:
rerun tests

benfred:
@karlhigley  @jperez999  @rjzamora  I think the NA failure in the unittests is unrelated to this PR - I spent some time debugging this and left me thoughts here https://github.com/NVIDIA/NVTabular/pull/687#issuecomment-811553045 . That PR has a 'fix' - but would like to figure out *why* this is happening

benfred:
rerun tests

benfred:
rerun tests

benfred:
rerun tests

benfred:
rerun tests

benfred:
rerun tests

benfred:
Rerun tests

PR #682, opened 2021-03-31
Add `seed_fn` to TF dataloader for consistent seeding across GPUs

karlhigley:
This passes a seed function supplied to the TF dataloader to the underlying dataloader backend, which uses the function to generate a new random seed when an iterator is created. This allows users to inject a function that uses Horovod to choose a consistent seed across processes, so that each epoch can shuffle the dataset partitions in a consistent way across processes while maintaining the ability to shard the dataset across processes by partition numbers.


benfred (approved):
nice!

PR #685, opened 2021-03-31
Add (local) Groupby Operation

### LINKED ISSUE #641: [FEA] Sequential / Session-based recommendation and time series support - Group by sorting values by timestamp

rjzamora:
Addresses most of #641
Pairs nicely with #677

Adds a `Groupby` operator that locally transforms each partition with a groupby aggregation.  The supported aggregations include anything supported by `cudf.groupby(...).agg`, but also include "list", "first", and "last" for list-based aggregations.


nvtabular/ops/groupby.py
```diff
@@ -0,0 +1,229 @@
...
+    # Redirect to _first or _last
+    return _first(x) if kind == "first" else _last(x)
+
+
+def _first(x):
+    # Convert each element of a list column to be the first
+    # item in the list
+    offsets = x.list._column.offsets
+    elements = x.list._column.elements
+    return [elements[offsets[i]] for i in range(0, len(offsets) - 1)]
```
benfred, line 221: I've been wondering if we can do this without iteration over each value - and instead use something like [array broadcasting](https://numpy.org/doc/stable/user/basics.broadcasting.html) for this .

What do you think about something like 


```suggestion
    return elements[offsets[:-1]]
    return elements[offsets[1:].values - 1]
```

for the _first function here

and 
```python
    return elements[offsets[1:].values - 1]
```

to get the last item in _last? This would avoid iterating over the elements 1by1 in python

benfred (approved):

nvtabular/ops/groupby.py
rjzamora: Yeah - Totally makes sense.  Thanks for suggesting this.

rjzamora:
rerun tests

PR #687, opened 2021-04-01
Fit a single stat operator at a time

benfred:
We're seeing unittest failures, usually along the lines of
``` TypeError: float() argument must be a string or a number, not '_NAType'```

This seems to only occur for me when calculating the stats for both FillMedian
and Categorify in a single dask.compute call. Calculating each op independently
doesn't have this problem and seems to work consistently.

### DIFF
--- a/nvtabular/workflow.py
+++ b/nvtabular/workflow.py
@@ -124,27 +124,21 @@ class Workflow:
                 # this shouldn't happen, but lets not infinite loop just in case
                 raise RuntimeError("failed to find dependency-free StatOperator to fit")
 
-            stats, ops = [], []
             for column_group in current_phase:
                 # apply transforms necessary for the inputs to the current column group, ignoring
                 # the transforms from the statop itself
                 transformed_ddf = _transform_ddf(ddf, column_group.parents)
-
                 op = column_group.op
                 try:
-                    stats.append(op.fit(column_group.input_column_names, transformed_ddf))
-                    ops.append(op)
+                    stat = op.fit(column_group.input_column_names, transformed_ddf)
+                    if self.client:
+                        computed = self.client.compute(stat).result()
+                    else:
+                        computed = dask.compute(stat, scheduler="synchronous")[0]
                 except Exception:
                     LOG.exception("Failed to fit operator %s", column_group.op)
                     raise
-
-            if self.client:
-                results = [r.result() for r in self.client.compute(stats)]
-            else:
-                results = dask.compute(stats, scheduler="synchronous")[0]
-
-            for computed_stats, op in zip(results, ops):
-                op.fit_finalize(computed_stats)
+                op.fit_finalize(computed)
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by benfred** [comment]
I spent some time this afternoon debugging the ```TypeError: float() argument must be a string or a number, not '_NAType'``` unittest failure we've seen a bunch here, and had some thoughts

1. This test failure isn't related to None/NaN values in the dataset.  Because of a bug (?) in how we're generating the dataset, the x/y/id fields we're calculating the median for don't have any nulls. I believe this line should be a 'continue' instead of a 'break' https://github.com/NVIDIA/NVTabular/blob/36f7c31fa1711f2ba4fac2bf49d8ccf1c76f051b/tests/conftest.py#L109-L110

2. I can hit this same unittest failure on my local envirionment by running this test in a loop until it fails. I'm just hackily going like 
```python
while True:
    ret = os.system(" pytest test_torch_dataloader.py::test_gpu_dl_break[None-parquet-1000-0.06] ")
    if ret:
        break
```

3. This test failure doesn't seem to be dataset dependant. I can force the dataset to be the same at each test run and some runs will work and some won't work.
4. Calculating the medians on its own always works - the error seems to be because in this test we're calculating FillMEdian/Categorify in a single dask compute call . This PR moves to calculate each op separately and seems to "solve" this unittest failure (though I'm still not sure why)

jperez999 (approved):

rjzamora:
Thanks for the writeup @benfred !

I'm a bit concerned about splitting the statistics into independent `compute` calls, since this *could* lead to a performance regression in cases where the two calculations can share IO-task output (though I am not 100% sure here - and the solution in the PR may be the best way to go).  I wonder if we are accidentally mutating task output in place somewhere - I will try to poke around and help illuminate the root cause.

benfred:
>  since this could lead to a performance regression in cases where the two calculations can share IO-task output (

Agreed - I'd rather not merge this if possible. I can imagine a case where we have two different statoperators operating on the exact same set of input columns: the existing code would handle this in a single pass over the data and this PR would change it to be two passes.

However - we don't have any example workflows where this would happen right now  and I don't think we should see any perf impact on these examples (since our statops are on disjoint sets of columns), so I'm not sure how big a deal this is in practice.

rjzamora:
> However - we don't have any example workflows where this would happen right now and I don't think we should see any perf impact on these examples (since our statops are on disjoint sets of columns), so I'm not sure how big a deal this is in practice.

Right - I totally aggree that we don't really have any concrete examples where we can/should re-use the same IO tasks.  So, if this change produces better stability, it may be worth hypothetical overhead.  With that said, I am hoping that #689 will actaully resolve the CI problem, since it will effectively remove the overlap between stat-operator graphs that do not operate on the same column groups.

benfred:
Closing this in favour of https://github.com/NVIDIA/NVTabular/pull/689

PR #688, opened 2021-04-01
Avoid in-place mutation of df in FillMedian

rjzamora:
Possible fix for recent CI failures.  It is technically not "legal" to modify input data to a task "in place."  It seems that we are doing this in other transforms as well.

Possible alternative to #687

### DIFF
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -342,7 +342,7 @@ class Categorify(StatOperator):
 
     @annotate("Categorify_transform", color="darkgreen", domain="nvt_python")
     def transform(self, columns: ColumnNames, df: DataFrameType) -> DataFrameType:
-        new_df = df.copy(deep=False)
+        new_df = df.copy()
         if isinstance(self.freq_threshold, dict):
             assert all(x in self.freq_threshold for x in columns)
 
--- a/nvtabular/ops/fill.py
+++ b/nvtabular/ops/fill.py
@@ -48,14 +48,15 @@ class FillMissing(Operator):
 
     @annotate("FillMissing_op", color="darkgreen", domain="nvt_python")
     def transform(self, columns, df: DataFrameType) -> DataFrameType:
+        _df = type(df)()
         if self.add_binary_cols:
             for col in columns:
-                df[f"{col}_filled"] = df[col].isna()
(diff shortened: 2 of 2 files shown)


benfred:
If we need to avoid mutating the input - I'd rather do it in _transform_partitions on the workflow than in each op: https://github.com/NVIDIA/NVTabular/blob/36f7c31fa1711f2ba4fac2bf49d8ccf1c76f051b/nvtabular/workflow.py#L310 . 

However, we're failing on the 'fit' function - and its not clear to me how changing the 'transform' functions (which are only called after fit is performed) will fix this problem. There are no operators before the categorify/fillmedian ops in the workflow for this unittest (https://github.com/NVIDIA/NVTabular/blob/36f7c31fa1711f2ba4fac2bf49d8ccf1c76f051b/tests/unit/test_torch_dataloader.py#L101-L104) - so I don't think any transforms will have been called by the time this fails.

rjzamora:
> If we need to avoid mutating the input - I'd rather do it in _transform_partitions on the workflow than in each op

Right, we could always do it in _transform_partition, but it *may* be tough to find a consistent/efficient mechanism at that level (I don’t think you want to pass in a full copy, for example, but maybe you do).

> There are no operators before the categorify/fillmedian ops in the workflow for this unittest

Yes and no - For the example above, the‘fit’ call will execute two graphs, one for Categorify, and one for FillMedian. When I look at the actual graphs, they both start with the same read_parquet task, and then comes _transform_partition in both graphs. So you are right that FillMedian is probably not the problem, but I’m willing to bet that some task coming after read parquet **is**.

rjzamora:
> EDIT: It seems that those _transform_partition calls are just column selection. Sorry - I am used to seeing column selection as a getitem task, and I mistook them for actual transformations. I also misunderstood the general workflow enough to believe that there were transform dependencies within the fit.

Just a somewhat-related side note:  In more-typical Dask workflows, the read-parquet tasks would have slightly different names in the two graphs (Categorify vs. FillMedian), because they are targetting a different set of columns.  However, this doesn't happen in NVTabular, because we are selecting the columns at the cudf/pd.DataFrame level, rather than the ddf level (and so the parquet->getitem optimization doesn't move the column selection into the read-parquet call).

PR #689, opened 2021-04-01
Add ddf column-selection code path to _transform_ddf

rjzamora:
While trying to find the root cause for the CI failures mentioned in #687, I noticed that NVTabular is not selecting columns in a way that allows Dask to push the selection down to read_parquet at graph-optimization time.  This PR adds a simple code path to perform column selection at the ddf level within `_transform_ddf` in cases where the `column_groups` are all simple column lists (no attached ops or parents).

Optimized read-parquet graph layer BEFORE this change (for a 'x', 'y', 'id' selection):
```
{'read-parquet-f6433cfb39b77001e5a8aae0a9a98bce': ParquetSubgraph<name='read-parquet-f6433cfb39b77001e5a8aae0a9a98bce', n_parts=2, columns=['label', 'x', 'id', 'name-cat', 'name-string', 'y']>}
```

Optimized read-parquet graph layer AFTER this change (for a 'x', 'y', 'id' selection):
```
{'read-parquet-9bb9af4d8bdbd3d554c829c1186dcf5a': ParquetSubgraph<name='read-parquet-9bb9af4d8bdbd3d554c829c1186dcf5a', n_parts=2, columns=['x', 'y', 'id']>}
```

Notice that the column list changes in the `ParquetSubgraph` repr.

### DIFF
--- a/nvtabular/workflow.py
+++ b/nvtabular/workflow.py
@@ -293,7 +293,7 @@ def _transform_ddf(ddf, column_groups):
     # Otherwise, Dask will not push the column selection into the
     # IO function.
     if all((c.op is None and not c.parents) for c in column_groups):
-        return ddf[columns]
+        return ddf[list(set(columns))]
 
     # TODO: constructing meta like this loses dtype information on the ddf
     # sets it all to 'float64'. We should propogate dtype information along
--- a/nvtabular/workflow.py
+++ b/nvtabular/workflow.py
@@ -292,8 +292,7 @@ def _transform_ddf(ddf, column_groups):
     # If so, we should perform column selection at the ddf level.
     # Otherwise, Dask will not push the column selection into the
     # IO function.
-    if all((c.op is None and not c.parents) for c in column_groups):
-        return _ddf_column_selection(ddf, column_groups)
+    return ddf[columns]
 
     # TODO: constructing meta like this loses dtype information on the ddf
     # sets it all to 'float64'. We should propogate dtype information along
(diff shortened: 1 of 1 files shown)


rjzamora:
rerun tests

nvtabular/workflow.py
```diff
@@ -288,6 +288,13 @@ def _transform_ddf(ddf, column_groups):
 
     columns = list(flatten(cg.flattened_columns for cg in column_groups))
 
+    # Check if we are only selecting columns (no transforms).
+    # If so, we should perform column selection at the ddf level.
+    # Otherwise, Dask will not push the column selection into the
+    # IO function.
+    if all((c.op is None and not c.parents) for c in column_groups):
+        return _ddf_column_selection(ddf, column_groups)
```
benfred, line 296: We've already gotten the names of all the columns on line 289 above - could we just go something like 
```suggestion
        return ddf[columns]
```
here instead?

benfred:
thanks for this! I'm interested to see how much this improves performance =)

nvtabular/workflow.py
rjzamora: Ah - Good point!  Can't believe I missed that =)

benfred:
This seems to fix the fillmedian unittest failures we were seeing - I've just run this branch on the flaky test ~30 times and haven't seen any failures yet (and used to fail reliably by the 10th retry before).

benfred (approved):

benfred (approved):

PR #694, opened 2021-04-02
Support cpu-based Datasets in Groupby Op

rjzamora:
Small follow-up to #685

- Simplifies algorithm a bit to take advantage of `list` aggregation (now handles both `"list"` and `list` input from user)
- Supports operation on `Dataset` objects created with `cpu=True`


rjzamora:
@benfred - I realized after the Groupby op went in that cudf already supports a `list` aggregation (but not by the name `"list"`).  This PR takes advantage of the existing functionality.  Since the same feature exists in Pandas, I also used this as an opportunity to provide cpu support.

rjzamora:
rerun tests

karlhigley (approved):

benfred (approved):

benfred:
rerun tests

rjzamora:
rerun tests

PR #696, opened 2021-04-02
Tf len fix

jperez999:
Removes infinite iterator behavior. Iterates now accurately on length. This change was necessary because the iterator would be activated before iteration when the keras model ran a peek into the first batch to gain shape and dtype information about batches. With this PR, we a reset every time len is called. Meaning everytime we start iteration we ensure to reset. ***keras uses len and get__item to iterate over dataset.


nvtabular/loader/tensorflow.py
```diff
@@ -238,6 +240,7 @@ def __len__(self):
         """
         # TODO: what's a better way to do this inheritance
         # of the appropriate methods? A Metaclass?
+        DataLoader.stop(self)
```
karlhigley, line 243: It seems a little bit weird to stop the dataloader when computing the length, in the sense that it's an unexpected side effect happening inside a method that's essentially a property of the object. Do TF training loops make a new iter each epoch? Wondering if [the stop inside the `__iter__` method](https://github.com/NVIDIA/NVTabular/blob/2a37c581f5807714732d8ab17245f583dfc1d9e2/nvtabular/loader/backend.py#L245) is sufficient? Or maybe there's another place we can put this?

nvtabular/loader/tensorflow.py
```diff
@@ -246,19 +249,7 @@ def __getitem__(self, idx):
...
-        except StopIteration:
-            # TODO: I would like to do a check for idx == 0
-            # here, but that requires that tf.keras.Model.fit
-            # be called with shuffle=False, and that seems
-            # small enough that it would be too easy to miss
-            # for many users. That said, blind reinitialization
-            # is probably irresponsible, so worth thinking
-            # of something better here
-            # return StopIteration
-            DataLoader.__iter__(self)
```
karlhigley, line 260: It makes sense that we don't want to create a new iterator and return more data here, but do we still want to catch `StopIteration`?

nvtabular/loader/torch.py
```diff
@@ -73,6 +73,7 @@ def __init__(
         device=None,
         global_size=None,
         global_rank=None,
+        drop_last=False,
```
karlhigley, line 76: I think this covers it for Torch, but just double-checking: Do we need to any additional changes here to parallel the TF changes (e.g. with stopping the dataloader)?

tests/unit/test_tf_dataloader.py
```diff
@@ -33,6 +33,55 @@
 tf_dataloader = pytest.importorskip("nvtabular.loader.tensorflow")
 
 
+@pytest.mark.parametrize("batch_size", [10, 9, 8])
+@pytest.mark.parametrize("drop_last", [True, False])
+@pytest.mark.parametrize("num_rows", [100])
+def test_tf_drp_reset(tmpdir, batch_size, drop_last, num_rows):
```
karlhigley, line 39: Should we test the Torch side too? Wondering how different that test would be. If it's mostly the same, maybe the overlapping parts could be a `DataLoader` test?

karlhigley (commented):
Looks good! Left a few comments and questions, which are mostly me trying to wrap my head around the change.

nvtabular/loader/backend.py
```diff
@@ -27,7 +28,7 @@
 
 
 def _num_steps(num_samples, step_size):
-    return (num_samples - 1) // step_size + 1
+    return math.ceil(num_samples / step_size)
```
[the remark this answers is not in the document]
jperez999: Umm... I think math ceil just rounds up on a division that does not end on an integer... Which I think is what we want here. If dividing the number of samples by the batch size gives a XX.xxx that means we need XX + 1 batches. And ceil call enforces that behavior. Or am I wrong? https://docs.python.org/3/library/math.html#math.ceil

nvtabular/loader/tensorflow.py
jperez999: That does not work because the keras sequence class we are using only employs the len and get_item. We are doing this to eliminate the need for the infinite iterator behavior currently employed. https://www.tensorflow.org/api_docs/python/tf/keras/utils/Sequence. Next step would be to change this to a tf dataset instead of a keras sequence, since it is supported now https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit.

nvtabular/loader/tensorflow.py
jperez999: yeah I agree that is why I moved it out of there...?

nvtabular/loader/torch.py
jperez999: No because the pytorch dataloader actually runs using the iterator logic as opposed to get_item and len...

tests/unit/test_tf_dataloader.py
jperez999: This is a good idea... do we want to work on that here, or in another PR?

nvtabular/loader/backend.py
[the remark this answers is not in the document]
karlhigley: Yeah, you're right about `ceil`! I'm more trying to understand the old behavior and why it was written that way than suggesting anything needs to change here.

nvtabular/loader/tensorflow.py
karlhigley: Ah, the possibility of a `StopIteration` still exists but we're going to bubble it out to the calling code? :+1:

tests/unit/test_tf_dataloader.py
karlhigley: Based on your comment above re: Torch, it seems like Torch doesn't have any particular code to test. I could go either way on pursuing a `DataLoader` test here or in a subsequent PR.

nvtabular/loader/tensorflow.py
karlhigley: Okay, cool! I'll write a ticket about that and stick it in a future release.

nvtabular/loader/backend.py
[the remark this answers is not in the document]
karlhigley: I can't see any reason this won't work, so I'm going to resolve this comment thread, but still curious what the heck `(num_samples - 1) // step_size + 1` was about.

karlhigley (approved):
LGTM at this point

tests/unit/test_tf_dataloader.py
jperez999: So for sure I am working on the test for the torch side... because it should be there. But I want to test the actual framework specific code also, So I don't know if this should be a DataLoader level test. But we should have DataLoader level tests for sure.

karlhigley: Yeah, testing it for both frameworks covers the bases nicely. :+1:

tests/unit/test_torch_dataloader.py
```diff
@@ -39,6 +39,53 @@
 GPU_DEVICE_IDS = [d.id for d in numba.cuda.gpus]
 
 
+@pytest.mark.parametrize("batch_size", [10, 9, 8])
+@pytest.mark.parametrize("drop_last", [True, False])
+@pytest.mark.parametrize("num_rows", [100])
+def test_tf_drp_reset(tmpdir, batch_size, drop_last, num_rows):
```
benfred, line 45: Can we go with a more descriptive name for the test?  I assume that 'tf' here means tensorflow - maybe we should rename this to torch here?  Also I'd prefer 'drop' to 'drp' 

```suggestion
def test_torch_drop_reset(tmpdir, batch_size, drop_last, num_rows):
```

(also this test seems to be failing in CI - and I'm not sure why =(

jperez999: yeah its because of the difference in external facing batch tuple format. I am working on it now. But will change name.

PR #697, opened 2021-04-05
PyTorch support for Inference

oyilmaz-nvidia:
This PR includes the code to generate PyTorch config file and the NVTabular+PyTorch ensemble model. 

Even though the export functions for Triton works well, we should work on a nice class structure and design for the export functions in the next releases.


oyilmaz-nvidia:
@benfred @rnyak I still need to add multihot support but I guess we can start the review process since we don't have much time left for the release.

rnyak (approved):
I tested the code on an example nb. It worked for me.

nvtabular/inference/triton/__init__.py
```diff
@@ -492,3 +624,34 @@ def _convert_dtype(dtype):
...
+    if dtype == torch.uint32:
+        return "uint32"
+    if dtype == torch.uint16:
+        return "uint16"
+    if dtype == torch.uint8:
+        return "uint8"
+    if dtype == torch.bool:
+        return "bool"
+    if cudf.utils.dtypes.is_string_dtype(dtype):
+        return model_config.TYPE_STRING
```
benfred, line 656: Is this logic right? Everywhere else is returning a numpy style string , and this is returning a triton enum

nvtabular/inference/triton/__init__.py
```diff
@@ -492,3 +624,34 @@ def _convert_dtype(dtype):
     if cudf.utils.dtypes.is_string_dtype(dtype):
         return model_config.TYPE_STRING
     raise ValueError(f"Can't convert dtype {dtype})")
+
+
+def _convert_torch_dtype(dtype):
```
benfred, line 629: This function doesn't appear to be used anywhere - can we remove?

nvtabular/inference/triton/__init__.py
```diff
@@ -248,7 +328,14 @@ def convert_triton_output_to_df(columns, response):
 
 
 def _generate_nvtabular_config(
```
benfred, line 330: I'm wondering if we should split this function up into 3 (for tf/hugectr/pytorch). There isn't much code in common between the 3 options

nvtabular/inference/triton/__init__.py
```diff
@@ -76,6 +77,76 @@ def export_tensorflow_ensemble(model, workflow, name, model_path, label_columns,
...
+    # generate the PT saved model
+    pt_path = os.path.join(model_path, name + "_pt")
+    pt_model_path = os.path.join(pt_path, str(version), "model.onnx")
+
+    os.makedirs(pt_path, exist_ok=True)
+    os.makedirs(os.path.join(pt_path, str(version)), exist_ok=True)
+
+    pt_config = _generate_pytorch_config(name + "_pt", pt_path, model_info)
+
+    torch.onnx.export(
```
benfred, line 133: nice! glad you managed to get this working with onnx

nvtabular/inference/triton/__init__.py
oyilmaz-nvidia: Sorry, this was an oversight. Will fix this.

nvtabular/inference/triton/__init__.py
benfred: lets track in https://github.com/NVIDIA/NVTabular/issues/708

benfred (approved):

PR #702, opened 2021-04-06
Generate local seeds from global seed before and after shuffling dataset

karlhigley:
This adjusts the dataloader backend to achieve three conditions related to random seeding and shuffling:

- Before generating a shuffle seed collaboratively across workers, each worker should have a different (but deterministic) random seed. Since the random seed may be set globally in the top-level training script, we have to generate local seeds before choosing a shuffle seed.
- After generating a shuffle seed together, all workers should have the same random seed so that they can execute the same shuffle and choose partitions to work on without overlap.
- After shuffling, each worker should have a different deterministic random seed again for training.


nvtabular/loader/backend.py
```diff
@@ -243,6 +243,20 @@ def _gather_indices_for_dev(self, dev):
...
         return self.indices[start : start + per_worker].tolist()
 
+    def _generate_local_seed(self):
+        random_state = cp.random.get_random_state()
+        seeds = random_state.tomaxint(size=self.global_size)
+        local_seed = seeds[self.global_rank]
+        cp.random.seed(local_seed.get())
+
+    def _shuffle_indices(self):
+        self._generate_local_seed()
```
benfred, line 253: This might be a silly question - but why do we need to generate the seed before and after shuffling? Could we just generate this once on dataloader init?

benfred (approved):

PR #704, opened 2021-04-06
Merlin env

jperez999:
change  env to merlin from rapids.


jperez999:
@bschifferer Not sure if I should have updated the docker-compose files... I didnt know if they were locked to the 0.4 tag for a reason?

benfred (approved):

PR #709, opened 2021-04-08
Add a 'download_file' utility function

benfred:
This adds a 'download_file' function that will download example files, optionally
unzip or untar them once downloaded, and display a progress bar in notebooks while
doing so. This replaces the 'wget/unzip' calls in notebooks.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -75,14 +75,14 @@ The container will open a shell when the run command execution is completed. You
 root@2efa5b50b909:
 ```
 
-1. Activate the ```rapids``` conda environment by running the following command:
+1. Activate the ```merlin``` conda environment by running the following command:
    ```
-   root@2efa5b50b909: source activate rapids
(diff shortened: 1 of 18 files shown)


benfred:
rerun tests

benfred:
rerun tests

jperez999 (approved):

rnyak:
@benfred I tested your PR that seems it is working. Somehow, it displays a progress bar in the notebook, but the percentage does not change. something like that:
```
downloading ml-25m.zip: 0.00B [00:00, ?B/s]
unzipping files:   0%|          | 0/8 [00:00<?, ?files/s]
```

benfred:
> Somehow, it displays a progress bar in the notebook, but the percentage does not change. something like that:

Seems like that was an issue with the JS version of tqdm in the TF container. 'Fixed' by defaulting to ascii progress bars only.

Issue #707, opened 2021-04-08
[BUG] Properly preparing NVTabular preprocessed data for HugeCTR inference


vinhngx:
**Describe the bug**
NVTab encodes categorical data is not directly compatible with HugeCTR requirement and would require the value to be shifted - so that they occupy different integer ranges

Concrete conceptual example: provided there are 10 users and 10 items, NVtab encodes users in 0-9 range, and items in 0-9 range. HugeCTR expect users in 0-9 range and items in 10-19 range

**Steps/Code to reproduce bug**
A concrete example of shifting:

https://gitlab-master.nvidia.com/dl/hugectr/hugectr/-/blob/master/notebooks/ecommerce-example.ipynb

**Expected behavior**

**Environment details (please complete the following information):**
 - Environment location: [Bare-metal, Docker, Cloud(specify cloud provider)]
 - Method of NVTabular install: [conda, Docker, or from source]
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used
 

**Additional context**

Issue #710, opened 2021-04-08
[Task] Add a unit test to check the accuracy of the models after deployed for inference


oyilmaz-nvidia:
We are running tests only to see if the whole inference pipeline is working or not. But, we also need to check if there is any change for the expected accuracy. In the bug reported here https://github.com/NVIDIA/NVTabular/issues/707, the accuracy of the NVTabular + HugeCTR is below the expected level even though the whole inference pipeline works.

benfred:
Done with #735

PR #712, opened 2021-04-09
[REVIEW] Add offsets for the categorical columns

oyilmaz-nvidia:
This is the PR to fix the bug reported in this issue; https://github.com/NVIDIA/NVTabular/issues/707


nvtabular/inference/triton/__init__.py
```diff
@@ -463,6 +472,13 @@ def get_column_types(path):
     return json.load(open(os.path.join(path, "column_types.json")))
 
 
+def get_slot_sizes(path):
+    if not path.exists(path):
+        return json.load(open(os.path.join(path, "slot_size_array.json")))
```
benfred, line 477: Why are you checking if the path doesn't exist? Also if the path doesn't exist, how can the slot_size_array.json file inside the path exist?

nvtabular/inference/triton/__init__.py
```diff
@@ -463,6 +472,13 @@ def get_column_types(path):
...
+    else:
+        return None
```
benfred, line 479: The only caller of this function seems to be in model_hugectr.py - maybe we should move there?

Also the code in model_hugectr.py doesn't handle this function returning None - maybe just remove the check to see if the file exists?

nvtabular/inference/triton/model_hugectr.py
```diff
@@ -51,6 +51,9 @@ def initialize(self, args):
         self.workflow = nvtabular.Workflow.load(workflow_path)
         self.model_config = json.loads(args["model_config"])
         self.column_types = get_column_types(workflow_path)
+        self.slot_sizes = get_slot_sizes(workflow_path)
+        self.slot_sizes.insert(0, 0)
+        self.slot_sizes.pop()
```
benfred, line 56: Does this work? The slot sizes file is written out as a dictionary afaict on line 138 of ```__init__.py``` , and this code seems to be expecting a list

benfred (changes requested):
You need to run black on the changes here - the style check is failing because of it.

How have you tested this code? I've left some comments but I'm very doubtful that this PR works.

nvtabular/inference/triton/model_hugectr.py
oyilmaz-nvidia: Forgot to push some of the updates before turning the title to "REVIEW". Just pushed them.

nvtabular/inference/triton/__init__.py
oyilmaz-nvidia: Updated this. And, moved it into model_hugectr.py

nvtabular/inference/triton/__init__.py
oyilmaz-nvidia: Updated it.

benfred:
rerun tests

shashank3959:
Onur and I tested the change, and it seemed to work. 

Using the merlin v0.4 inference container, we observed an issue loading the ensemble model with the following setting in dcn_nvt/config.pbtxt
instance_group {
  count: 1
  kind: KIND_CPU
}


After removing this config, the ensemble loaded successfully.

oyilmaz-nvidia:
@shashank3959 We have created an issue here to add a unit test to check the accuracy. Here is that issue; https://github.com/NVIDIA/NVTabular/issues/710

We'll add that test in the next release.

benfred (approved):

benfred:
rerun tests

vinhngx:
Great work. Thanks for promptly fixing this.

I've got a question. In the following:
output_df[self.column_types["cats"]] + self.slot_sizes

is slot_sizes the actual slot sizes, or accumulative sum of the slot sizes?

I think it needs to be the accumulative sum.

oyilmaz-nvidia:
@vinhngx Created another PR for the accumulative sum. Could you please review this? 
https://github.com/NVIDIA/NVTabular/pull/735

Will add you as a reviewer.

PR #716, opened 2021-04-09
Fix groupby bug caused by empty aggregation lists

rjzamora:
The current groupby logic allows internal aggregation dictionaries to have empty list values, which trips up the column-naming logic a bit.  This PR includes a trivial fix (with a minor test-coverage tweak).


rjzamora:
cc @rnyak

benfred (approved):

PR #722, opened 2021-04-11
Defer importing pytorch in inference

benfred:
We don't have pytorch installed in the TF/hugectr training containers,
meaning we couldn't export these models to triton. Fix by lazily importing
pytorch when generating the pytorch ensemble.


rnyak (approved):
Tested the PR with `merlin-tensorflow-training` 0.5 container, and it works without issue.

PR #723, opened 2021-04-11
Create example notebook for multi-gpu training with PyTorch and Horovod

karlhigley:
This is a parallel example notebook to the TF notebook, with a few changes:
* I replaced the `os` calls with Python equivalents (not just for style, the `os` calls weren't working for me due to running `docker` with `sudo`.)
* I added `client.shutdown()` to free GPU memory from the Dask cluster. It probably works fine without that if you have huge amounts of GPU memory, but it causes an OOM error if you have 11GB.


examples/multi-gpu-movielens/01-03b-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-PyTorch.ipynb
```diff
@@ -0,0 +1,683 @@
...
+    "\n",
+    "zip_path = pathlib.Path(BASE_DIR, 'ml-25m.zip')\n",
+    "\n",
+    "if not zip_path.exists():\n",
+    "    zip_path.parent.mkdir(parents=True, exist_ok=True)\n",
+    "    \n",
+    "    request.urlretrieve(\"http://files.grouplens.org/datasets/movielens/ml-25m.zip\", str(zip_path))\n",
+    "    \n",
+    "    with zipfile.ZipFile(zip_path,\"r\") as zip_ref:\n",
+    "        zip_ref.extractall(BASE_DIR)"
```
benfred, line 110: I added a utility function in https://github.com/NVIDIA/NVTabular/pull/709 to download and unzip files. It's pretty much the same as here, except includes progress bars of download progress.

examples/multi-gpu-movielens/01-03b-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-PyTorch.ipynb
```diff
@@ -0,0 +1,683 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Writing ./trainer.py\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%writefile './trainer.py'\n",
```
benfred, line 484: nice!

benfred (approved):

examples/multi-gpu-movielens/01-03b-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-PyTorch.ipynb
```diff
@@ -0,0 +1,683 @@
...
+    "    shuffle=shuffle,\n",
+    "    out_files_per_proc=out_files_per_proc,\n",
+    ")\n",
+    "workflow.transform(valid_iter).to_parquet(\n",
+    "    output_path=pathlib.Path(BASE_DIR, \"valid\"),\n",
+    "    shuffle=shuffle,\n",
+    "    out_files_per_proc=out_files_per_proc,\n",
+    ")\n",
+    "\n",
+    "client.shutdown()"
```
benfred, line 389: Should we also close the cluster? (like `cluster.close()` here)?

examples/multi-gpu-movielens/01-03b-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-PyTorch.ipynb
karlhigley: This was @bschifferer's suggestion, and a really good one!

examples/multi-gpu-movielens/01-03b-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-PyTorch.ipynb
karlhigley: I rewrote this to use `pathlib` instead of `os` because `os` calls were failing for me (which are also used in the `download_file` util function.) I'm happy to standardize this code and stop having to write it all over the place, but I'm not sure `download_file` addresses my issue. I'll give it a try.

karlhigley: I do like progress bars! :smile:

examples/multi-gpu-movielens/01-03b-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-PyTorch.ipynb
karlhigley: Yes, probably so. Weirdly, that method is not in the Dask Distributed API documentation.

examples/multi-gpu-movielens/01-03b-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-PyTorch.ipynb
benfred: what was the error with the os.path calls? I'm surprised that it doesn't work (like I think pathlib has better ergonomics / and is a more modern way of manipulating paths and since we only work on python 3.6+ we should use - but I didn't realize it was doing any thing fundamentally different than os.path). 

Also, We don't need to change for this release - we can merge as is (and maybe standardize and upgrade download_file etc to also use pathlib etc).

jperez999:
@benfred @karlhigley How are we testing horovod now on pytorch? we just took out the only script we had running a unit test?  I have same question for tf...

karlhigley:
@jperez999 Good catch, that's an error on my part! I've added a commit here that switches the unit test to using the script written out by the notebook (which should be the same.)

jperez999:
@karlhigley are we fixing pytorch and tf unit tests separately or do you plan on handling both in this PR?

karlhigley:
@jperez999 I thought I'd fix each framework's tests in their open PRs. Hoped I could sneak that past our CI, but will that work?

karlhigley:
Looks like it will not work. I'll open a separate PR that fixes both.

PR #733, opened 2021-04-13
set long_description_content_type

benfred:
Our README isn't rendering right at https://pypi.org/project/nvtabular. Fix
by setting long_description_content_type in setup.py


benfred:
rerun tests

karlhigley (approved):

PR #735, opened 2021-04-14
[REVIEW] Add cumulative sum to the slot sizes for NVTabular + HugeCTR ensemble

oyilmaz-nvidia:


vinhngx:
haven't got chance to test this hand-on but the logic looks good to me now.

nvtabular/inference/triton/model_hugectr.py
```diff
@@ -87,7 +87,7 @@ def execute(self, requests: List[InferenceRequest]) -> List[InferenceResponse]:
 
             if "cats" in self.column_types:
                 output_df[self.column_types["cats"]] = (
-                    output_df[self.column_types["cats"]] + self.slot_sizes
+                    output_df[self.column_types["cats"]] + self.offsets
```
benfred, line 90: How about doing this in place? This is more readable and can avoid allocating memory 

```python
output_df[self.column_types["cats"]] += self.offsets
```

benfred (commented):
How has this been tested?

oyilmaz-nvidia:
@benfred I have asked Shashank to test this again. The ultimate solution will be adding an accuracy testing. Created an issue for that (https://github.com/NVIDIA/NVTabular/issues/710)

nvtabular/inference/triton/model_hugectr.py
oyilmaz-nvidia: Sure, we can do that as well. But, I am not sure if cuDF is allocating or not allocation memory space when we do it that way. Something we should test, I guess.

benfred:
> @benfred I have asked Shashank to test this again. The ultimate solution will be adding an accuracy testing. Created an issue for that (#710)

An automated accuracy test would be great (and essential moving forward) - but is there anything blocking you from testing this change manually yourself?  What's Shashank doing that you can't do?  This change looks good to me, but this is the 2nd time we're attempting to fix this bug and I'd like some assurances that it works before merging

oyilmaz-nvidia:
Obviously, I tested myself 😄. I don’t get any runtime error and I wasn’t getting any runtime error before. It’s just following the vcsr format that hugectr team described.

Feel free to go over their document and check yourself as well to see if I missed anything. I did it myself on an example that HugeCTR team provided and it looks good to me now.

To make sure, I just asked extra review from Vinh and Shashank.

Again, the ultimate solution is to add an accuracy test.

________________________________

benfred:
> Obviously, I tested myself 😄. I don’t get any runtime error and I wasn’t getting any runtime error before. It’s just following the vcsr format that hugectr team described.

That sounds like you've only tested that this doesn't thrown any errors, not that this change fixes the bug.  

> To make sure, I just asked extra review from Vinh and Shashank.

How are Vinh and Shashank testing this? Why can't you do the same tests?

oyilmaz-nvidia:
I guess you missed this;

"_I did it myself on an example that HugeCTR team provided and it looks good to me now._"

Don't know what else I can do other than checking the accuracy for a known case.

Please don't ask the same question again :)

vinhngx:
I think checking test accuracy on the Criteo data is a good and necessary step.

jperez999:
rerun tests

oyilmaz-nvidia:
The following code was followed to write the CSR conversion in the integration test;

https://github.com/NVIDIA/HugeCTR/blob/f8e7f94c3e64257f52c2cb148f28c75f56d33b77/tools/criteo_predict/criteo2predict.py#L20

I will ask the hugectr team to review the model.py as well.

@benfred flake8 complains about the 'from mpi4py import MPI' line in the tests/integration/hugectr/test_nvt_hugectr_training.py. We don't use it in the code but when I remove it, hugectr side crashes. I guess, hugectr is using it internally.

Will try to add 1-2 dense features into test data but need to look into HugeCTR API a little bit more to see how we can add that into the model.

benfred:
> @benfred flake8 complains about the 'from mpi4py import MPI' line in the tests/integration/hugectr/test_nvt_hugectr_training.py. We don't use it in the code but when I remove it, hugectr side crashes. I guess, hugectr is using it internally.

You can add ```  # noqa```  comment after the line to skip flake8 on it

nvtabular/inference/triton/model_hugectr.py
```diff
@@ -86,9 +86,7 @@ def execute(self, requests: List[InferenceRequest]) -> List[InferenceResponse]:
                 output_tensors.append(Tensor("DES", np.array([[]], np.float32)))
 
             if "cats" in self.column_types:
-                output_df[self.column_types["cats"]] = (
-                    output_df[self.column_types["cats"]] + self.slot_sizes
-                )
+                output_df[self.column_types["cats"]] += self.offsets
                 cats_np = _convert_cudf2numpy(output_df[self.column_types["cats"]], np.int64)
                 output_tensors.append(
```
yingcanw, line 91: `row_index = np.arange(len_cats_np + 1, dtype=np.int32).reshape(1, len_cats_np + 1)`
The current csr conversion supported by this line is only for the case where each slot contains only one cat feature, right?
In other words, the current script only supports the conversion of one-hot encoded categorical data. The number of cat features is equal to the number of slots (cats_np.shape[1] = slot_sizes.size)

oyilmaz-nvidia: Yes, we only support one-hot encoding right now.

oyilmaz-nvidia:
rerun tests

tests/integration/hugectr/test_nvt_hugectr_training.py
```diff
@@ -0,0 +1,380 @@
...
+                        "embedding_vec_size": 16,
+                        "combiner": 0,
+                    },
+                },
+                {
+                    "name": "reshape1",
+                    "type": "Reshape",
+                    "bottom": "sparse_embedding1",
+                    "top": "reshape1",
+                    "leading_dim": 32,
```
yingcanw, line 310: If you use 3 slots(slot_num=3) to train the model and the _embedding_vec_size_ is 16, then the **leading_dim** should be 3*16=48,

tests/integration/hugectr/test_nvt_hugectr_training.py
```diff
@@ -0,0 +1,380 @@
...
+            source=DATA_DIR + "train/_file_list.txt",
+            eval_source=DATA_DIR + "valid/_file_list.txt",
+            check_type=hugectr.Check_t.Non,
+            label_dim=1,
+            label_name="label",
+            dense_dim=0,
+            dense_name="dense",
+            slot_size_array=slot_sizes,
+            data_reader_sparse_param_array=[
+                hugectr.DataReaderSparseParam(hugectr.DataReaderSparse_t.Distributed, 3, 1, 2)
```
rnyak, line 188: May be this line is the issue: (3, 1, 2)

would it work if you change it to `(3, 1, 3)` or `(4, 1, 3)` ?

oyilmaz-nvidia (commented):
Ok, updated the the leading dimension to 48. What @rnyak suggested worked.

tests/integration/hugectr/test_nvt_hugectr_training.py
oyilmaz-nvidia: (4, 1, 3) worked. Thanks!

tests/integration/hugectr/test_nvt_hugectr_training.py
oyilmaz-nvidia: Updated that. Now it works with 48 leading_dim.

benfred (approved):

PR #767, opened 2021-04-23
Build the model_config proto file in setup.py

benfred:
Rather than hack into the importerror and run protoc to compile the
model_config.proto file when needed, compile this file as part of the
build in setup.py


benfred:
rerun tests

jperez999 (approved):

PR #776, opened 2021-04-26
Adds params config for Criteo Integration Tests

albert17:


tests/integration/test_notebooks.py
```diff
@@ -224,6 +232,21 @@ def _run_notebook(
...
+        )
+        lines = [transform(line) for line in lines]
+        transform = lambda line: line.replace(
+            "device_pool_frac = 0.8", "device_pool_frac = " + str(params[1])
+        )
+        lines = [transform(line) for line in lines]
+        transform = lambda line: line.replace(
+            "part_mem_frac = 0.15", "part_mem_frac = " + str(params[2])
+        )
+        lines = [transform(line) for line in lines]
```
benfred, line 248: I think we don't allow assigning lambdas (flake8). Maybe something like this instead?

```suggestion
        def transform(line):
            line = line.replace("device_limit_frac = 0.7", "device_limit_frac = " + str(params[0]))
            line = line.replace("device_pool_frac = 0.8", "device_pool_frac = " + str(params[1]))
            return line.replace("part_mem_frac = 0.15", "part_mem_frac = " + str(params[2]))

        lines = [transform(line) for line in lines]
```

albert17: Got it! Thanks Ben!

benfred (approved):

PR #780, opened 2021-04-27
Fix hugectr notebooks

benfred:
The recent changes for black-nb and flake8-nb messed up the hugectr notebooks.

Specifically applying black-nb against JSON formatted data added a trailing comma
to most entries, meaning it couldn't be parsed as JSON. Also we need the MPI import
for hugectr training.

Fix - and disable black-nb until it can be disabled on a per-cell level


benfred:
rerun tests

rnyak (approved):
@benfred it works fine but two things needs to be adressed:

1.  I get error from this line `os.rmdir(train_dir)`. it gives  the following error, which does not make sense. may be we can use `shutil.rmtree(train_dir)` instead?
```
OSError: [Errno 39] Directory not empty: '/model/data/train' error.
```
2. we should not forget to update the container image name/tag in this line in the inference nbs:
`docker run -it --gpus=all -p 8000:8000 -p 8001:8001 -p 8002:8002 -v ${PWD}:/model nvcr.io/nvidia/merlin/merlin-inference:0.5`

Issue #500, opened 2020-12-16
[FEA] Abstract PyTorch Data loader to provide Sparse Tensors for list columns


gabrielspmoreira:
**Is your feature request related to a problem? Please describe.**

When working with list columns in parquet files, the PyTorch Data loader returns them in a specific tuple representation, with length 2, described as follows: 
1. The contenated values of lists from all samples
2. The offsets, that allows to know when to break the values into the lists corresponding to each sample

That representation seems useful if the feature is categorical, and if the values will be fed directly to an torch.nn.EmbeddingBag() to lookup for the corresponding embeddings, as show [here](https://github.com/NVIDIA/NVTabular/blob/e2d3c2c65dc2225a1fa050472be16ecdaafbeb94/nvtabular/framework_utils/torch/layers/embeddings.py#L54-L88)

```python
self.embedding_names = [i for i in embedding_table_shapes.keys()]
        self.embedding_layers = torch.nn.ModuleList(
            [
                torch.nn.EmbeddingBag(*embedding_table_shapes[key], mode=mode)
                for key in self.embedding_names
            ]
        )
```

But if the list columns are not categoricals (e.g. floats), or if you don't want to use EmbeddingBag for your categoricals, then you need to reconstruct manually the list columns into a PyTorch sparse tensor.

**Describe the solution you'd like**
It would be better if the NVTabular data loader provides both the offset representation (for EmbeddingBag) and a also sparse tensor representation for general usage. 

I have implemented a Data loader extension that converts the current offset representation of list columns to a sparse or dense tensor, so that it is easier to use in PyTorch pipelines.

```python
from nvtabular.loader.torch import TorchAsyncItr as NVTDataLoader
from nvtabular import Dataset as NVTDataset


class NVTDataLoaderWrapper(NVTDataLoader):
    def __init__(self, *args, **kwargs):
        self.default_seq_features_len = None
        if 'default_seq_features_len' in kwargs:
            self.default_seq_features_len = kwargs.pop('default_seq_features_len')
        else:
            raise ValueError('NVTabular data loader requires the "default_seq_features_len" argument "'+\
                                'to create the sparse tensors for list columns')
        
        super(NVTDataLoaderWrapper, self).__init__(*args, **kwargs)

    def __enter__(self):
        return None

    def __exit__(self, type, value, traceback):
        return None

    def __next__(self):
        cat_features, cont_features, label_features = super(NVTDataLoaderWrapper, self).__next__()
        
[...]

PR #793, opened 2021-05-06
Add Sparse Representation Capability to Dataloaders

jperez999:
This solves for #500.


jperez999:
rerun tests

karlhigley:
The "conflict" to be resolved here was literally whitespace, so I went ahead and resolved it. (Can you tell I'm excited about getting this PR merged? :smile: )

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

examples/getting-started-movielens/03-Training-with-PyTorch.ipynb
```diff
@@ -573,8 +573,15 @@
     "\n",
     "for epoch in range(1):\n",
     "    start = time()\n",
-    "    train_loss, y_pred, y = process_epoch(train_loader, model, train=True, optimizer=optimizer)\n",
-    "    valid_loss, y_pred, y = process_epoch(valid_loader, model, train=False)\n",
+    "    train_loss, y_pred, y = process_epoch(train_loader,\n",
+    "                                          model,\n",
+    "                                          train=True,\n",
+    "                                          transform=dict_transform(train_loader.dataset).transform,\n",
```
benfred, line 579: Can we just make this the default inside process_epoch if not set ?

nvtabular/framework_utils/tensorflow/layers/embedding.py
```diff
@@ -198,7 +185,7 @@ def __init__(self, feature_columns, aggregation="concat", name=None, **kwargs):
         super(DenseFeatures, self).__init__(name=name, **kwargs)
 
     def build(self, input_shapes):
-        assert all(shape[1] == 1 for shape in input_shapes.values())
+        # assert all(shape[1] == 1 for shape in input_shapes.values())
```
benfred, line 188: If we don't need these asserts anymore, we should just delete rather than comment out
```suggestion
```

nvtabular/framework_utils/tensorflow/layers/embedding.py
```diff
@@ -324,7 +311,7 @@ def __init__(self, feature_columns, name=None, **kwargs):
         super(LinearFeatures, self).__init__(name=name, **kwargs)
 
     def build(self, input_shapes):
-        assert all(shape[1] == 1 for shape in input_shapes.values())
+        # assert all(shape[1] == 1 for shape in input_shapes.values())
```
benfred, line 314:
```suggestion
```

nvtabular/framework_utils/torch/utils.py
```diff
@@ -17,6 +17,63 @@
...
+                    cols.append(v)
+            concats.append(torch.cat(cols, axis=1))
+        return (
+            concats[0].to("cuda", dtype=torch.long),
+            concats[1],
+            batch[1].to("cuda", dtype=torch.long),
+        )
+
+
+class dict_transform:
```
benfred, line 40: Why is this lower snake case? 

```suggestion
class DictTransform:
```

nvtabular/framework_utils/torch/utils.py
```diff
@@ -17,6 +17,63 @@
...
+    def transform(self, batch):
+        batch, labels = batch
+        cats = None
+        conts = None
+        # take a part the batch and put together into subsets
+        if self.cats:
+            cats = self.create_stack(batch, self.cats)
+        if self.conts:
+            conts, _ = self.create_stack(batch, self.conts)
+        # labels = self.create_stack(batch, self.labels)
```
benfred, line 55:
```suggestion
```

nvtabular/loader/backend.py
```diff
@@ -504,4 +524,39 @@ def _create_tensors(self, gdf):
...
+            if isinstance(tensor, tuple):
+                tensor, lists = tensor
+            names = [i for i in names if i not in lists]
+
+            # break list tuples into two keys, with postfixes
+            # TODO: better choices for naming?
+            list_columns = list(lists.keys())
+            for column in list_columns:
+                values, nnzs = lists.pop(column)
+                lists[column] = values, nnzs
```
benfred, line 539: Umm - is this necessary? It seems like you're popping of a tuple of (values, nnzs) from column and then inserting back with the same key - can we just delete?

```suggestion
```

nvtabular/loader/tensorflow.py
```diff
@@ -210,6 +211,8 @@ def __init__(
         global_size=None,
         global_rank=None,
         drop_last=False,
+        sparse_list=None,
+        sparse_max=None,
```
benfred, line 215: What do these two parameters do?  Can you add a description to the docstring?

examples/multi-gpu-movielens/tf_trainer.py
```diff
@@ -84,8 +86,9 @@ def seed_fn():
     inputs[col] = tf.keras.Input(name=col, dtype=tf.int32, shape=(1,))
 # Note that we need two input tensors for multi-hot categorical features
 for col in CATEGORICAL_MH_COLUMNS:
-    inputs[col + "__values"] = tf.keras.Input(name=f"{col}__values", dtype=tf.int64, shape=(1,))
-    inputs[col + "__nnzs"] = tf.keras.Input(name=f"{col}__nnzs", dtype=tf.int64, shape=(1,))
+    inputs[col] = tf.keras.Input(name=f"{col}__values", dtype=tf.int64, shape=(1,)), tf.keras.Input(
+        name=f"{col}__nnzs", dtype=tf.int64, shape=(1,)
+    )
```
karlhigley, line 91: I think this syntax makes a tuple, but maybe it would be clearer with explicit parens?

nvtabular/dispatch.py
```diff
@@ -87,14 +87,14 @@ def _series_has_nulls(s):
         return s._column.has_nulls
 
 
-def _is_list_dtype(s):
+def _is_list_dtype(ser):
     """Check if Series contains list elements"""
-    if isinstance(s, pd.Series):
-        if not len(s):  # pylint: disable=len-as-condition
+    # import pdb; pdb.set_trace()
```
karlhigley, line 92: Stray debugging line here

nvtabular/framework_utils/tensorflow/layers/embedding.py
```diff
@@ -99,25 +99,10 @@ def _validate_stack_dimensions(feature_columns):
...
-                "is another dense feature, please rename your features"
-                "as this syntax is reserved for identifying multi-valent"
-                "categorical variables".format(feature_name, feature_name)
-            )
-        if feature_name + "__nnzs" not in inputs:
-            raise ValueError(
-                "Feature {} had __values entry in input dictionary, "
-                "but no __nnzs entry. Make sure all the relevant data "
-                "is being passed".format(feature_name)
-            )
```
karlhigley, line 116: It's really nice to see this failure case disappear. 👍🏻

nvtabular/framework_utils/tensorflow/layers/embedding.py
karlhigley: We might still want to assert some version of this for the tensors inside the tuple?

nvtabular/loader/backend.py
```diff
@@ -180,12 +180,15 @@ def __init__(
         global_size=None,
         global_rank=None,
         drop_last=False,
+        sparse_list=None,
+        sparse_max=None,
```
karlhigley, line 184: We might want to add a docstring to the `DataLoader` class to explain what all these params mean sometime soon.

nvtabular/loader/backend.py
```diff
@@ -480,16 +499,17 @@ def _create_tensors(self, gdf):
             gdf.drop(columns=column_names, inplace=True)
 
             scalars, lists = self._separate_list_columns(gdf_i)
+
             x = None
             if scalars:
+                # should always return dict column_name: values, offsets (optional)
```
karlhigley, line 505: Is there a reasonable way to enforce that it does return this?

nvtabular/loader/backend.py
karlhigley: Ah, maybe this chunk of code was intended to do the rename mentioned in the `TODO` comment above? This would be a reasonable way to do that if the postfixes were applied here.

nvtabular/loader/backend.py
```diff
@@ -504,4 +524,39 @@ def _create_tensors(self, gdf):
...
+                lists.update(zip(names, tensors))
+            elif len(names) == 1:
+                lists[names[0]] = tensor
+            X.update(lists)
+
+        for column_name in X:
+            if column_name in self.sparse_list:
+                if column_name not in self.sparse_max:
+                    raise ValueError(
+                        f"Did not convert {column_name} to sparse missing sparse_max entry"
```
karlhigley, line 553:
```suggestion
                        f"Did not convert {column_name} to sparse due to missing sparse_max entry"
```

nvtabular/loader/backend.py
```diff
@@ -504,4 +524,39 @@ def _create_tensors(self, gdf):
...
+                    )
+                X[column_name] = self._to_sparse_tensor(X[column_name], column_name)
+
+        # TODO: use dict for labels as well?
+        # would require output layers to match naming
```
karlhigley, line 558: I can see arguments both directions on this, but it seems reasonable to leave it as is for now.

nvtabular/loader/tensorflow.py
```diff
@@ -270,6 +275,13 @@ def _get_device_ctx(self, dev):
     def _split_fn(self, tensor, idx, axis=0):
         return tf.split(tensor, idx, axis=axis)
 
+    def _tensor_split(self, tensor, idx, axis=0):
```
karlhigley, line 278: What API is this function matching? 🤔

nvtabular/ops/categorify.py
```diff
@@ -445,7 +445,8 @@ def _get_embedding_order(cat_names):
     cat_names : list of str
         names of the categorical columns
     """
-    return sorted(cat_names)
+    # return sorted(cat_names)
```
karlhigley, line 448:
```suggestion
```

nvtabular/loader/tensorflow.py
jperez999: good catch... I needed it when the abstraction required splitting in the backend module... but that is no longer the case.

nvtabular/loader/tensorflow.py
jperez999: the parameters were updated in the framework specific dl doc strings.

nvtabular/loader/backend.py
jperez999: they are in the framework specific dataloaders, if we want to transition to the backend... we can do that...

jperez999:
rerun tests

nvtabular/loader/tensorflow.py
jperez999: actually this is used it is the api used in the handle_tensor function in the backend code. I was confusing it for something I had in the sparse_tensors logic.

nvtabular/loader/backend.py
jperez999: it was used to break apart the lists into values and nnz dictionary entries.

jperez999:
rerun tests

nvtabular/loader/backend.py
jperez999: Do you mean like an assert? i.e. `assert isinstance(x, dict)` ?

nvtabular/loader/torch.py
```diff
@@ -113,10 +129,48 @@ def _LONG_DTYPE(self):
...
+        return values, offsets, diff_offsets, num_rows
+
+    def _get_max_seq_len(self, diff_offsets):
+        # get_max_seq_len, return int
+        return int(diff_offsets.max())
+
+        # Building the indices to reconstruct the sparse tensors
+
+    def _get_indices(self, offsets, diff_offsets):
+        row_ids = torch.arange(len(offsets) - 1).to("cuda")
```
benfred, line 152: Can we create these directly on the GPU instead of creating on the CPU and then copying over with ```to("cuda")```?

nvtabular/loader/backend.py
benfred: agreed -

nvtabular/loader/tensorflow.py
```diff
@@ -188,6 +189,12 @@ class KerasSequenceLoader(tf.keras.utils.Sequence, DataLoader):
     - reader_kwargs: dict
         extra kwargs to pass when instantiating the underlying
         `nvtabular.Dataset`
+    sparse_list : list(str) or None
+        list with column names of columns that should be represented as sparse tensors
+    sparse_max : dict
+        dictionary of key: column_name + value: integer representing max sequence length for column
+    sparse_dense : bool
+        bool value to activate transforming sparse tensors to dense
```
benfred, line 197: I think this API is a little confusing - like we are specifying which columns to treat as sparse, and then have a flag to convert to dense.

Can we simplify this at all? Like if we're producing dense outputs,  would it be possible to just add the columns to the continuous columns?

benfred: resolved over slack - going to leave as is with some minor name parameter name changes

nvtabular/loader/backend.py
karlhigley: Eh, maybe not worth it.

nvtabular/loader/backend.py
karlhigley: Yeah, maybe something like that? Not sure if we really need one here, just trying to think about how to turn the comment into code.

karlhigley:
As far as I can tell, the comments have now been resolved and it's just a matter of getting pylint to stop timing out on CI, yeah?

jperez999:
rerun tests

jperez999:
rerun tests

benfred:
rerun tests

benfred (approved):

PR #803, opened 2021-05-11
Add a list slicing op

### LINKED ISSUE #734: [FEA] Truncate List columns (sparse tensors) - related to the GroupBy op
This feature is related with #641 "Sequential / Session-based recommendation and time series support - Group by sorting values by timestamp".
After grouping, some sequences (e.g. user sessions or time series) might be very long, and for some ML models sequences with maximum (fixed) length are required. So lists truncation is necessary.
As currently List columns are internally represented as sparse vectors, it is currently not possible to  use a LambdaOp to truncate the list values to a maximum length.

I would like either an option on Groupby op to truncate all aggregated list columns to the same maximum length or an independent TruncateList op that would truncate selected list columns

As a workaround for this problem I am extending the NVT PyTorch dataloader, converting the internal representation of list columns to PyTorch sparse tensors (as shown in #500), converting them to dense tensors (with padding zeros in the right) and then slicing the second dimension of the tensor to the maximum length.
But storing longer sequences than needed in the parquet files is a waste of space and requires workarounds like this in the model size.

benfred:
This adds an operator to slice rows of list columns. This will let us truncate list
column rows to only take the first N or last N items for instance.

Closes #734


nvtabular/ops/list_slice.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @annotate("ListSlice_op", color="darkgreen", domain="nvt_python")
+    def transform(self, columns: ColumnNames, df: DataFrameType) -> DataFrameType:
+        on_cpu = isinstance(df, pd.DataFrame)
+        ret = pd.DataFrame() if on_cpu else cudf.DataFrame()
+        for col in columns:
+            # handle CPU via normal python slicing (not very efficient)
+            if on_cpu:
+                ret[col] = [row[self.start : self.end] for row in df[col]]
+                continue
```
karlhigley, line 66: What would you think about splitting the CPU and GPU paths into separate clauses of an `if/else`? Purely a style thing, but might be a little easier to read.

nvtabular/ops/list_slice.py
```diff
@@ -0,0 +1,139 @@
...
+        end = min(max(0, end), original_row_size)
+
+        row_sizes[rowid + 1] = end - start
+
+
+@numba.cuda.jit
+def _slice_rows(start, offsets, elements, new_offsets, new_elements):
+    """slices rows of a list column. requires the 'new_offsets' to
+    be previously calculated (meaning that we don't need the 'end' slice index
+    since thats baked into the new_offests"""
```
karlhigley, line 127:
```suggestion
    since thats baked into the new_offsets)"""
```

karlhigley (approved):

karlhigley:
Okay, I had to learn about writing CUDA kernels with Numba in order to understand this, but I read through it and it makes sense to me.

nvtabular/ops/list_slice.py
benfred: done!

PR #805, opened 2021-05-11
Dask-graph and host-caching performance optimizations

rjzamora:
While experimenting with performance optimizations in Dask/NVTabular, I discovered a few interesting things:

1. It is no longer faster to retrieve a cudf DataFrame from parquet-formatted host memory than arrow-formatted host memory (unspilling from arrow now seems to be much faster).  Therefore, this PR removes the parquet-based caching strategy from NVTabular.
2. Since NVTabular never calls compute/persist on Dataframe collections, we don't benefit from an automatic Dataframe-based HighLevelGraph optimization pass in Dask.  This means we do not benefit from `Blockwise`-task fusion or from "column-projection" into the IO functions (like cudf.read_parquet).  This PR adds an `_ensure_optimize_dataframe_graph` to explicitly run these optimizations in both `fit` and in `Dataset.to_parquet`
3. While experimenting with (2), I realized that there is actually a silly bug in Dask (my fault) that prevents the column-projection optimization from working in **any** Dask-dataframe code. I submitted [a PR to Dask](https://github.com/dask/dask/pull/7645) to address this.

With both this PR and the fix in dask-7645, I get similar (maybe slightly better?) performance for the dask-Criteo benchmark on a DGX-1:

```
Dask-NVTabular DLRM/Criteo benchmark
--------------------------------------
partition size     | 4261216256
protocol           | tcp
device(s)          | 0,1,2,3,4,5,6,7
rmm-pool-frac      | 0.9
out-files-per-proc | 8
num_io_threads     | 0
shuffle            | PER_PARTITION
cats-on-device     | True
======================================
Runtime[s]         | 154.64741921424866
======================================
```


nvtabular/utils.py
```diff
@@ -121,3 +123,36 @@ def report(chunk, chunksize, total):
         with gzip.open(local_filename, "rb") as input_file:
             with open(local_filename[:-3], "wb") as output_file:
                 shutil.copyfileobj(input_file, output_file)
+
+
+def _ensure_optimize_dataframe_graph(ddf=None, dsk=None, keys=None):
```
karlhigley, line 128: I have a (potentially irrational) vendetta against the word "ensure", but as much as I want to, I really can't argue with this usage. Totally accurate. :+1:

nvtabular/io/dask.py
```diff
@@ -352,7 +353,10 @@ def _ddf_to_dataset(
             task_list.append(key)
         dsk[name] = (lambda x: x, task_list)
 
-    graph = HighLevelGraph.from_collections(name, dsk, dependencies=[ddf])
+    graph = _ensure_optimize_dataframe_graph(
+        dsk=HighLevelGraph.from_collections(name, dsk, dependencies=[ddf]),
```
karlhigley, line 357: What's different about this call to `HighLevelGraph.from_collections()` from the one above in `_simple_shuffle()` that makes the optimization wrapper relevant here and not there? (Based on the name, I'd guess maybe `_simple_shuffle()` wouldn't benefit much from the optimizations?)

karlhigley (approved):

nvtabular/utils.py
rjzamora: :) Happy to rename it to `_optimize_dataframe_graph` if you'd like

nvtabular/io/dask.py
rjzamora: Good question! 

In NVTabular, we only want to run an explicit optimization pass when we are about to `compute` something that is not a Dataframe collection.  For example, we want to make sure the graph is optimized before we compute statistics (which are derived from a collection, but not collections themselves).  We do not need to optimize when we merely construct a new collection, because (1) Dask will automatically run the optimizations if the collection itself is computed, and (2) the user may be performing further transforms on the collection anyway.

karlhigley: Cool, thanks for the explanation!

nvtabular/utils.py
karlhigley: I considered suggesting that, but on reflection, I concluded it would be less accurate/informative. (Mostly left this comment to signal that I'd actually read and tried to think about the code, because a LGTM with no comments is kind of sus.)

rjzamora:
rerun tests

benfred:
rerun tests

benfred (approved):

PR #808, opened 2021-05-12
fix for multihot

jperez999:
multihot fix for embedding sizes


tests/unit/test_torch_dataloader.py
```diff
@@ -366,7 +366,6 @@ def test_mh_support(tmpdir):
     assert idx > 0
 
 
-@pytest.mark.skip(reason="fails with cuda illegal memory access")
```
benfred, line 369: nice - glad you figured out what was going on with this test (though in retrospect we probably should not have skipped this )

benfred (approved):
lgtm!

PR #820, opened 2021-05-18
Add column group select

### LINKED ISSUE #809: [FEA] Select a subset of a ColumnGroup
No

Currently, the best way I know of to select a ColumnGroup generated by your code (say as the result of a LambdaOp) is set-subtractive, I.E., you have a ColumnGroup with generated features A,B,C, and just want B,C, the way to get this is group - ['A']. A better way to do this in some cases would be set-selective, IE, group['B','C'].

An alternative solution would be to add a select op.

**Additional context**

benfred:
Allow selecting columns from column groups using the ```__getitem__``` operator.
Closes #809

### DIFF
--- a/tests/unit/test_column_group.py
+++ b/tests/unit/test_column_group.py
@@ -1,8 +1,28 @@
 import cudf
 import pytest
+from cudf.tests.utils import assert_eq
 
 from nvtabular import ColumnGroup, Dataset, Workflow
 from nvtabular.ops import Categorify, Rename
+
+
+def test_column_group_select(tmpdir):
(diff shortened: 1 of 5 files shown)


tests/unit/test_column_group.py
```diff
@@ -1,10 +1,30 @@
 import cudf
 import pytest
+from cudf.tests.utils import assert_eq
 
 from nvtabular import ColumnGroup, Dataset, Workflow
 from nvtabular.ops import Categorify, Rename
 
 
+def test_column_group_select(tmpdir):
```
karlhigley, line 9: Doesn't look like this test uses `tmpdir`

karlhigley (approved):

tests/unit/test_column_group.py
benfred: nice catch! fixed

PR #824, opened 2021-05-18
Remove multigpu toy example from sphinx docs

benfred:
We are covering multi-gpu operation in the criteo example now, so lets
not also link in the multi-gpu example on a toy dataset


karlhigley:
I think the Horovod multi-GPU training notebooks reference the multi-GPU toy example notebook

benfred:
>I think the Horovod multi-GPU training notebooks reference the multi-GPU toy example notebook

This is just removing from the rendered sphinx docs - not removing from the multi-gpu toy example notebooks altogeher.  The horovod example notebooks also aren't rendered into our sphinx docs just yet either.

rnyak:
@benfred I built the docs, and it looks good to me.

rnyak (approved):

Issue #370, opened 2020-10-21
[FEA] JoinExternal option on nvt.Dataset  in addition to cudf


kkranen:
**Is your feature request related to a problem? Please describe.**
Currently, JoinExternal only supports cudf external tables. This works when the cudf table can be materialized, but in cases where the table you wish to join is large (for example, where the desired table to join is the dataset result of a workflow spread across many parquet files), the cudf table cannot be materialized directly and thus cannot be joined.

**Describe the solution you'd like**
Allow JoinExternal to accept nvt.Dataset as input.

**Describe alternatives you've considered**
Allow passing of multiple cudf dataframes into JoinExternal to be treated as a single dataframe 

**Additional context**
N/A

EvenOldridge:
Rick, can you take a look at this?  From what I remember we implemented join before we were using dask_cudf, and I think there's some missing plumbing to make this work more easily.  Ideally we should be able to plug data from multiple sources in using the native join from dask_cudf.

rjzamora:
Since `JoinExternal` is an `Operator`, it is required that the `transform` logic operates at the cudf/pandas level (not at the Dask-collection level).  Therefore, to join to an external `Dataset`/collection we will need to (within `transform`): (1) promote the target partition to a single-partition DataFrame-collection, (2) merge the promoted collection with the external Dataset/collection, and then (3) return a synchronously-computed result.  This might work, but would effectively mean that we are constrained to a broadcast-based join.

For more-complex join operations, we could always introduce a `Dataset.join_external` method.  However, I think it goes against the design of NVTabular to require inter-prtition shuffling within a transform operator.

PR #843, opened 2021-05-26
Expand JoinExternal to allow df_ext to be Dataset and Collection-based 

rjzamora:
This PR expands the available input types for the `df_ext` argument in `JoinExternal` to include `Dataset` and Dask-collection objects. These improvements do **not** completely address #370 , because that issue requests the ability to join a many-partition collection/Dataset.  Since `JoinExternal` is a vanilla `Operator`, we cannot perform a large shuffle-based join (the multi-partition join support in this PR requires all of `df_ext` to be broadcasted to every partition).

For a large (shuffle-based) merge, the user will need to use `dask.dataframe`/`dask_cudf`.  I don't think a more complex version of `JoinExternal` is practical.  We can certainly expose the dask.dataframe merge API within NVT, but it would need to be at the `Dataset` level.

**Side note on future work**:  The "multi-partition" join used in this PR could also be used within `Cateforify` to reference a multi-partition category list at transform time.  This is one possible way to avoid cudf's string-column size limit, because we could persist the category statistics as a list of parquet files (or as a collection).


nvtabular/dispatch.py
```diff
@@ -168,3 +190,76 @@ def _to_arrow(x):
...
+            # If input is a Dask collection, covert to dask_cudf
+            if isinstance(x, dask_cudf.DataFrame):
+                return x
+            return x.map_partitions(cudf.from_pandas)
+        elif isinstance(x, pa.Table):
+            return cudf.DataFrame.from_arrow(x)
+        else:
+            if isinstance(x, cudf.DataFrame):
+                return x
+            return cudf.DataFrame.from_pandas(x)
```
benfred, line 254: Would it be possible to use the existing nvtabular dataset code here?  Like we have code to convert from parquet/csv/dask/pandas/cudf/ here https://github.com/NVIDIA/NVTabular/blob/94d2960c95a6b9bef71483c80d80bf212429af32/nvtabular/io/dataset.py#L228-L269 . I'm wondering if we can just use the Dataset to convert to a known format in the JoinExternal op, and avoid having to have an extra conversion step here

rjzamora: Good point - The logic should probably be shared, and live in dispatch.py for both of these cases (so we can eventually isolate explicit references to cudf/dask_cudf).

rjzamora: I moved the shared Dataset logic into `dispatch.py` - let me know if you think this makes sense.  I am hoping to eventually consolidate all cudf-specific logic into one location (like dispatch).

benfred: nice! thanks

nvtabular/dispatch.py
```diff
@@ -13,22 +13,50 @@
...
 from cudf.core.column import as_column, build_column
 from cudf.utils.dtypes import is_list_dtype
-from dask.dataframe.utils import hash_object_dispatch
+
+try:
+    # Dask >= 2021.5.1
+    from dask.dataframe.core import hash_object_dispatch
+except ImportError:
+    # Dask < 2021.5.1
+    from dask.dataframe.utils import hash_object_dispatch
```
benfred, line 36: thanks for this! We've had this raised as an issue =)

nvtabular/ops/join_external.py
```diff
@@ -92,58 +102,91 @@ def __init__(
...
-        gdf.drop(columns=[tmp], inplace=True)
-        new_gdf.reset_index(drop=True, inplace=True)
-        return new_gdf
+        df[tmp] = _arange(len(df), like_df=df, dtype="int32")
+        new_df = self._merge(df, self._ext)
+        new_df = new_df.sort_values(tmp)
+        new_df.drop(columns=[tmp], inplace=True)
+        df.drop(columns=[tmp], inplace=True)
+        new_df.reset_index(drop=True, inplace=True)
+        return new_df
```
benfred, line 189: I like how you also added CPU support here at the same time =)

benfred (approved):
thanks for this - looks great!

rjzamora:
Thanks for the review @benfred !

PR #845, opened 2021-05-26
Expose useful Dask-Dataframe methods in Dataset class

rjzamora:
This PR has four general goals:

1. Expose common/useful Dask-Dataframe methods in `Dataset`.  This includes `compute`, `persist`, `head`, `tail`, `repartition`, and `merge`.  The idea is to make it easier for users to do Dask-like things without using `to_ddf` (e.g. `datset.head()` instead of `datset.to_ddf().head()`, and `datset.repartition(npartitions=8)` instead of `Dataset(datset.to_ddf().repartition(npartitions=8))` (note that `repartition` also supports the `partition_size` argument, where something like "1GB" can be used to specify the desired partition size).
2. Fix problematic `output_files` behavior in `to_parquet` when the user asks for a larger number of output files than the total number of partitions.  This PR adds an implicit `repartition` step to make sure NVTabular will actually produce the correct number of output files.
3. Make it possible for the user to merge two `Dataset` objects (using `Datset.merge`)
4. Add an optional `npartitions` argument to the `Dataset` initializer so that an input pandas/cudf DataFrame can be partitioned.


benfred (approved):

benfred:
Thanks for this!

PR #847, opened 2021-05-27
Fixing NumPy environment with CuPy environment

beingaryan:


examples/scaling-criteo/01-Download-Convert.ipynb
```diff
@@ -225,9 +225,9 @@
     "# the values will be hexadecimal strings that should\n",
     "# be converted to int32\n",
     "dtypes = {}\n",
-    "dtypes[\"label\"] = np.int32\n",
+    "dtypes[\"label\"] = cp.int32\n",
```
benfred, line 228: ```cp.int32``` is just an alias for ```np.int32```, so this change doesn't have any effect:
https://github.com/cupy/cupy/blob/13304cc25e1c6bd5810855236a883f2b398f93e5/cupy/__init__.py#L136

PR #848, opened 2021-05-27
[REVIEW] NVTabular+TF training and inference integration test

oyilmaz-nvidia:
Even though we run notebooks as a part of our integration tests, we do not compare the output of the Triton inference with the NVTabular transform() and TF predict() using the same data. This PR addresses the need of an integration test that does that. 

We'll be able to test C++ backend with these tests as well.

We might wanna add or run this test from the C++ backend repo.


oyilmaz-nvidia:
@albert17 Moved hugectr integration tests one folder up. FYI.

tests/integration/test_nvt_tf_training.py
```diff
@@ -0,0 +1,452 @@
...
+    workflow.output_dtypes["movieId"] = "int32"
+
+    MODEL_NAME_ENSEMBLE = "test_movielens_tf"
+    # model path to save the models
+    MODEL_PATH = os.path.join(MODEL_BASE_DIR, "models")
+
+    export_tensorflow_ensemble(model, workflow, MODEL_NAME_ENSEMBLE, MODEL_PATH, ["rating"])
+
+
+def test_nvt_tf_rossmann_training(tmpdir):
```
benfred, line 231: How is this different than the train notebook?  We're already running the TF training script here 

https://github.com/NVIDIA/NVTabular/blob/9e14acc4eeb103bdd9590fb2a28eb5a182151be2/tests/integration/test_notebooks.py#L144-L146

as part of the integration tests.

Is this code just cut and paste from the TF notebook - with the only difference being that we export the model to triton at the end?  If so maybe we should just add the export_tensorflow_ensemble call to the end of the rossmann training notebook like we do for movielens?

tests/integration/test_nvt_tf_training.py
```diff
@@ -0,0 +1,452 @@
...
+
+import nvtabular as nvt
+from nvtabular import ops
+from nvtabular.inference.triton import export_tensorflow_ensemble
+from nvtabular.utils import download_file
+
+TEST_N_ROWS = 64
+
+
+def test_nvt_tf_movielens_training():
```
benfred, line 41: Is this also any different from the movielens notebooks?

tests/integration/test_nvt_tf_inference.py
```diff
@@ -0,0 +1,104 @@
...
+import tritonclient.grpc as grpcclient
+from tritonclient.utils import np_to_triton_dtype
+
+import nvtabular as nvt
+
+
+# Update TEST_N_ROWS param in test_nvt_tf_trainin.py to test larger sizes
+@pytest.mark.parametrize("n_rows", [64, 45, 32, 14, 7, 1])
+@pytest.mark.parametrize("err_tol", [0.00001])
+def test_nvt_tf_movielens_inference(n_rows, err_tol):
```
benfred, line 33: This seems to require that tritonserver is already started with the model, which complicates how this will be run.

We do have some code in the unittests that starts the tritonserver process and waits for it to start up before running the test: https://github.com/NVIDIA/NVTabular/blob/9e14acc4eeb103bdd9590fb2a28eb5a182151be2/tests/unit/test_triton_inference.py#L23-L45 . 

At a minimum I think we need a shell script (potentially in ci/ folder) that shows how to run these tests. The inference integration tests are currently configured only in jenkins which not everyone has access to. @albert17 fyi.

benfred (commented):
Thanks for getting these tests together!

One of the important pieces of our integration tests is benchmarking on real data. We can verify correctness for the most part in unittests - but the cool thing about our integration tests is tracking performance over time and getting the numbers into ASV. Can you work with Alberto on this?

Also can you run black on your code? The CI should be passing before asking for a review =)

tests/integration/test_nvt_tf_training.py
oyilmaz-nvidia: _Can you work with Alberto on this?_

Sure! That's a pretty good idea!

_Is this code just cut and paste from the TF notebook_ 

Most of them, Yes. I added extra code that runs NVT transform and TF predict to test the whole Triton Pipeline. This test will also help us to test the C++ backend.

If you think we add that test part in the notebooks, that works for me but I didn't think that'd be something useful for the users.

tests/integration/test_nvt_tf_training.py
oyilmaz-nvidia: Similarly, this test has the extra code that runs NVT transform and TF predict.

tests/integration/test_nvt_tf_inference.py
oyilmaz-nvidia: @albert17 Let's work together on whatever is needed.

tests/integration/test_nvt_tf_training.py
oyilmaz-nvidia: _Also can you run black on your code?_

I actually did it. I am not sure why CI is giving this error. Maybe, I should update the black on my conda env.

benfred: > I actually did it. I am not sure why CI is giving this error. Maybe, I should update the black on my conda env

We have a pre-commit hook configured with the right version of black on it https://pre-commit.com/ if you enable this it'll run automatically on every commit you make

benfred: > If you think we can add that test part in the notebooks, that works for me but I didn't think that'd be something useful for the users.

I think we can keep the training in the notebooks (to reduce duplicate code) but can add the inference tests themselves separate from the notebooks

tests/integration/test_nvt_tf_inference.py
albert17: @benfred @oyilmaz-nvidia What I do right now for the existing integration tests is to launch the triton server before running the integration tests.

Fort unit tests we could use the same approach modifying the CI script.

Yes, for integration I need to add the scripts under CI folder. I keep forgetting. Let me try to put a PR today. Sorry

oyilmaz-nvidia: @albert17 I also need some help on how to get the running time measurements to the CI.

And, once this gets merged, we'll need to run this on CI.

tests/integration/test_nvt_tf_training.py
benfred: How about we put the nvt transform/ tf predict code to run in the other test script? That way we can eliminate most of the boilerplate/duplicate code here and just run the notebook.  The workflow and tensorflow model are both saved so this should be pretty easy

oyilmaz-nvidia: @benfred  I was planning to do something similar when I first started to this but one of my concern on this approach was that sometimes the notebook based integration tests fail due to some reasons (i.e. file name change, etc). And, this is one of the reasons I had this duplicate code. With this way, I know that it only fails if there is any change in the APIs. Also, I can make any changes easily if I have to (like adding extra randomly generated columns for testing a particular feature. We did this for the HugeCTR integration test using movielens example). 

I still think we should keep it as it is even though most of the them is duplicate code right now. I really like these integration tests are independent from the notebooks for easy changeability whenever we need it for testing purposes.

Also, we might wanna switch the original movielens and rossmann data with randomly generated data in the future as well to test certain features whenever we want to.

What do you think?

tests/integration/test_nvt_tf_inference.py
albert17: @benfred @oyilmaz-nvidia sorry for the delay. https://github.com/NVIDIA/NVTabular/pull/863 I will ping you for the measurements

albert17: @oyilmaz-nvidia Actually I will reply here for measurements. 

We need to use a benachmark in the similar way that criteo test has one, for example: https://github.com/NVIDIA/NVTabular/blob/main/tests/integration/test_notebooks.py#L69

For each benchmark there is a Class, for example for Criteo FastAI: https://github.com/NVIDIA/NVTabular/blob/main/tests/integration/criteo_parsers.py#L4

So we need to create one.

tests/integration/test_nvt_tf_training.py
oyilmaz-nvidia: @benfred I am trying to put only the nvt transform/ tf predict code in a separate test but I am hitting an error when I try to load the trained tf model to run predict. When I googled the error, I found out that other developers also hit the same error with TF version 2.3.

Here is a link about the error;

https://github.com/tensorflow/tensorflow/issues/43478

What do you suggest we should do?

tests/integration/test_nvt_tf_inference.py
oyilmaz-nvidia: @albert17 Updated the tests and now we have two main tests (test_nvt_tf_rossmann_inference_triton and test_nvt_tf_rossmann_inference_triton_mt) that take time measurements. I called the send_benchmark function but getting error. Actually, I don't know how to call the tests that send benchmark results for CI. Can you please help me with this?

oyilmaz-nvidia:
@benfred I updated the tests based on your feedback. FYI.

tests/integration/test_nvt_tf_inference.py
```diff
@@ -0,0 +1,339 @@
...
+        os.path.expanduser(MODEL_DIR), "movielens", TRITON_SERVER_PATH, TRITON_DEVICE_ID
+    ) as client:
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            for n_row in n_rows:
+                futures.append(executor.submit(_run_movielens_query, client, n_row))
+
+    for future in concurrent.futures.as_completed(futures):
+        diff, run_time = future.result()
+        assert (diff < err_tol).all()
+        # send_results(asv_db, bench_info, str(run_time))
```
albert17, line 72: `send_results `may be failing because os lack of aws credentials

albert17 (changes requested):

albert17 (approved):

oyilmaz-nvidia:
rerun tests

benfred:
rerun tests

albert17:
rerun tests

PR #861, opened 2021-06-04
Provide better graph labels for LambaOp

### LINKED ISSUE #860: [FEA] Add a way to name LambdaOps
It'd be nice to have a way to name LambdaOps so that they show up with that name in the workflow graph instead of just the non-descriptive moniker 'LambdaOp'

benfred:
The workflow graphs all had 'LambdaOp' before, which was pretty confusing if you
had multiple lambdas in the workflow. This change adds a 'label' to the operator
api which is used to label nodes in the graph.

LambdaOp will now allow users to specify their own label, or alternatively will
try to figure out a decent representation from the function itself using the
inspect module.

Closes #860

### DIFF
--- a/nvtabular/column_group.py
+++ b/nvtabular/column_group.py
@@ -205,7 +205,7 @@ class ColumnGroup:
     @property
     def label(self):
         if self.op:
-            return str(self.op.__class__.__name__)
+            return self.op.label
         elif self.kind:
             return self.kind
         elif not self.parents:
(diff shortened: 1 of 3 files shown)


benfred:
Some examples of how this looks:

* Recsys 2020 example (shows both lambda and named function ('splitmedia'):
Screen Shot 2021-06-04 at 3 21 38 PM
* Movielens example:
Screen Shot 2021-06-04 at 3 23 14 PM
* Outbrain example:
Screen Shot 2021-06-04 at 3 24 04 PM

jperez999 (approved):

PR #863, opened 2021-06-05
Adds integration scripts

albert17:
Adds initial integration scripts


ci/test_inference.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+cd /nvtabular/
+git pull origin main
+
+tritonserver --model-repository /model/models --backend-config=tensorflow,version=2 --model-control-mode=explicit &
+sleep 1m
+
+pytest tests/integration/test_notebooks.py::test_tf_inference_multihot_examples
```
benfred, line 9: Can we also run the hugectr inference test here too?

ci/test_inference.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+cd /nvtabular/
+git pull origin main
+
+tritonserver --model-repository /model/models --backend-config=tensorflow,version=2 --model-control-mode=explicit &
+sleep 1m
```
benfred, line 7: We have support here for running the inference server - and polling until its started (rather than just sleeping a fixed amount of time).
https://github.com/NVIDIA/NVTabular/blob/84ad9776e82c9be84259bd6d11ad54769bbce210/tests/unit/test_triton_inference.py#L23-L47

I'm wondering if we should move this function up one level (to tests/conftest.py) and then use in both the integration tests and unittest, instead of starting in this shell script.

albert17:
I will update after all the new integration tests are merged for 0.6

ci/test_inference.sh
albert17: This does not apply anymore. Servers are being launched within the tests now.

ci/test_inference.sh
albert17: Added and completely changed structure

albert17:
@oyilmaz-nvidia @benfred let me know if you think I am missing anything here

ci/test_integration.sh
```diff
@@ -0,0 +1,34 @@
...
+  pytest tests/integration/test_nvt_tf_inference.py::test_nvt_tf_rossmann_inference_triton
+  pytest tests/integration/test_nvt_tf_inference.py::test_nvt_tf_rossmann_inference_triton_mt
+  pytest tests/integration/test_nvt_tf_inference.py::test_nvt_tf_movielens_inference_triton
+  pytest tests/integration/test_nvt_tf_inference.py::test_nvt_tf_movielens_inference_triton_mt
+  pytest tests/integration/test_nvt_hugectr_inference.py::test_nvt_hugectr_inference
+else
+  echo "INVALID Container name"
+  exit 1
+fi
+
```
oyilmaz-nvidia, line 34: So, for nvt + tf inference integration tests, you first run the line 13 and 14, then 20 and 21, and lastly 25, 26, 27, and 28, right? If that's the case,

albert17: Correct. 

The idea is to run first the integration tests for merlin-training, merlin-tensorflow-training, and merlin-pytorch-training; and when it is done, run for merlin-inference

albert17:
rerun tests

albert17:
rerun tests

oyilmaz-nvidia (approved):

PR #865, opened 2021-06-07
Fix versions of Dask and Tensorflow in conda env for CUDA 11.0

karlhigley:
This also sorts the list of dependencies to make the file  easier to read.


benfred (approved):
lgtm - should we also update the other conda environment files though (cuda 11.1/11.2 etc).

karlhigley:
@benfred I'll check the others. I wanted to get a PR in quickly in case there was an advantage to using NVT's main branch instead of a hand-edited env file.

PR #868, opened 2021-06-09
Add a benchmarking script for nvtabular triton models

benfred:
This adds a simple script to benchmark nvtabular triton models, using
triton's perf_analyzer utility.

Since perf_analyzer requires a fairly long commandline with nvtabular
(we need to specify shape for each column), and we need to pass example
data - this wraps up the execution to automatically set the appropiate
commandline and transform a csv/parquet file with data into the JSON
format perf analyzer expects.

This saves out the results from perf_analyzer for various batch sizes to
a dataframe, and contains some basic matplotlib code to help visualize

### DIFF
--- a/nvtabular/inference/triton/benchmarking_tools.py
+++ b/nvtabular/inference/triton/benchmarking_tools.py
@@ -1,0 +1,138 @@
+""" utilities for running the triton perf_analyzer on a nvtabulat model """
+import json
+import os
+import subprocess
+import tempfile
+
+import cudf
+import pandas as pd
+
+import nvtabular as nvt
+
+
+def run_perf_analyzer(model_path, input_data_path, num_rows=10, model_version=1):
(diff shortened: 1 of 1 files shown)


benfred:
The output of this benchmarking script is kinda interesting to compare the cpp_backend and python_backend.

With the movielens example, we don't see much difference between the cpp_backend and python_backend:

Screen Shot 2021-06-08 at 10 35 50 PM

With the rossmann example, at smaller batch sizes the cpp backend and python backend are basically identical - but the cpp_backend starts to break away as results get higher.

Screen Shot 2021-06-08 at 10 38 03 PM

While the cpp_backend is doing better here with rossmann - the times are still too slow to be acceptable. Looking at the inference POC I put together last week (zoomed into the first 1k rows), shows that we are starting to get to our < 5ms target for latency here.

Screen Shot 2021-06-08 at 10 40 00 PM

Issue #869, opened 2021-06-09
[BUG] Model' object has no attribute 'mh_cat_layer' error with only multi-hot columns when training PyT model


rnyak:
**Describe the bug**
I am getting `AttributeError: 'Model' object has no attribute 'mh_cat_layer'`when I train my PyTorch model with only multi-hot columns.  

**Steps/Code to reproduce bug**
I used example in the [getting-started-with-movielens Pyt training nb](https://github.com/NVIDIA/NVTabular/blob/main/examples/getting-started-movielens/03-Training-with-PyTorch.ipynb).  I removed `CATEGORICAL_COLUMNS = ["movieId", "userId"]` single hot columns and trained model only with multi-hot columns `CATEGORICAL_MH_COLUMNS = ["genres"]` .

See the full error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<timed exec> in <module>

/nvtabular/nvtabular/framework_utils/torch/utils.py in process_epoch(dataloader, model, train, optimizer, loss_func, transform, amp, device)
    123             if amp:
    124                 with torch.cuda.amp.autocast():
--> 125                     y_pred = model(x_cat, x_cont)
    126                     y_pred_list.append(y_pred.detach())
    127                     loss = loss_func(y_pred, y)

/opt/conda/lib/python3.8/site-packages/torch/nn/modules/module.py in _call_impl(self, *input, **kwargs)
    887             result = self._slow_forward(*input, **kwargs)
    888         else:
--> 889             result = self.forward(*input, **kwargs)
    890         for hook in itertools.chain(
    891                 _global_forward_hooks.values(),

/nvtabular/nvtabular/framework_utils/torch/models.py in forward(self, x_cat, x_cont)
     86             x_cat, mh_cat = x_cat
     87         if mh_cat:
---> 88             mh_cat = self.mh_cat_layer(mh_cat)
     89             concat_list.append(mh_cat)
     90         x_cat = self.initial_cat_layer(x_cat)

/opt/conda/lib/python3.8/site-packages/torch/nn/modules/module.py in __getattr__(self, name)
    945             if name in modules:
    946                 return modules[name]
--> 947         raise AttributeError("'{}' object has no attribute '{}'".format(
    948             type(self).__name__, name))
    949 

AttributeError: 'Model' object has no attribute 'mh_cat_layer'

```
**Expected behavior**

`Models.py` should be designed to work only with `multi-hot` columns as well. 

[...]

PR #870, opened 2021-06-09
Make single-hot categoricals optional in the base PyTorch model

karlhigley:
This is related to #869, and takes a few steps toward making single-hot categoricals optional there. An additional step we might want to take would be to either make the embedding table shapes argument a dictionary or separate it into two different arguments for single and multi-hot categoricals to avoid those kinds of issues.


benfred:
This seems to fail with ````AttributeError: 'Model' object has no attribute 'mh_cat_layer’```` in ```tests/unit/test_notebooks.py::test_rossman_example```

karlhigley:
Superseded by #873

PR #874, opened 2021-06-10
Add cpu support for Bucketize

rjzamora:
Related to #534 

Adds support for cpu-backed data in `Bucketize` (and relevant test coverage).

### DIFF
--- a/nvtabular/ops/bucketize.py
+++ b/nvtabular/ops/bucketize.py
@@ -12,8 +12,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import cudf
+import numpy as np
 from nvtx import annotate
+
+from nvtabular.dispatch import DataFrameType, _array
 
 from .operator import ColumnNames, Operator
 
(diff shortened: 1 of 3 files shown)


nvtabular/ops/bucketize.py
```diff
@@ -54,15 +56,15 @@ def __init__(self, boundaries):
...
-            for boundary in b:
-                val += (gdf[col] >= boundary).astype("int")
-            new_gdf[col] = val
-        return new_gdf
+            new_df[col] = np.digitize(
+                df[col].values,
+                _array(b, like_df=df),
+                right=False,
+            )
+        return new_df
```
rjzamora, line 68: The deleted comment says that `cupy.digitize` was not supported in 7.8.  However, it looks like [`digitize` was added](https://github.com/cupy/cupy/pull/2758) before that release.  Therefore, I changed the logic to use `np.digitize` (which dispatches to cupy when the data is a cupy array).  I still need to figure out if we need a version check here.

benfred (approved):

nvtabular/ops/bucketize.py
rjzamora: Okay - I added a `LooseVersion` check that cupy is >=8.0.0 before using digitize (I confirmed that this is the earliest version with support)

PR #877, opened 2021-06-10
Add cpu support for JoinGroupby

rjzamora:
Related to #534

Adds support for cpu-backed data in `JoinGroupby` (and relevant test coverage).

### DIFF
--- a/nvtabular/dispatch.py
+++ b/nvtabular/dispatch.py
@@ -161,14 +161,34 @@ def _read_dispatch(df: DataFrameType = None, cpu=None, collection=False, fmt="pa
     return getattr(_mod, _attr)
 
 
-def _parquet_writer_dispatch(df: DataFrameType):
+def _parquet_writer_dispatch(df: DataFrameType, path=None, **kwargs):
     """Return the necessary ParquetWriter class to write
     data of a specified type.
+
+    If `path` is specified, an initialized `ParquetWriter`
+    object will be returned.  To do this, the pyarrow schema
+    will be inferred from df, and kwargs will be used for the
+    ParquetWriter-initialization call.
     """
+    _args = []
     if isinstance(df, pd.DataFrame):
-        return pq.ParquetWriter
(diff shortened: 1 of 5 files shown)


nvtabular/ops/categorify.py
```diff
@@ -1064,7 +1069,7 @@ def _read_groupby_stat_df(path, name, cat_cache, read_pq_func):
         with get_worker_cache("stats") as cache:
             if cache:
                 return fetch_table_data(cache, path, cache=cat_cache, reader=read_pq_func)
-    return read_pq_func(path, index=False)
+    return read_pq_func(path)  # , index=False)
```
rjzamora, line 1072:
```suggestion
    return read_pq_func(path)
```

nvtabular/dispatch.py
```diff
@@ -177,14 +177,34 @@ def _read_dispatch(df: DataFrameType = None, cpu=None, collection=False, fmt="pa
...
+    if isinstance(df, pd.DataFrame):
+        _cls = pq.ParquetWriter
+        if path:
+            _args.append(pa.Table.from_pandas(df, preserve_index=False).schema)
+    else:
+        _cls = cudf.io.parquet.ParquetWriter
+
+    if not path:
+        return _cls
+    return _cls(path, *_args, **kwargs)
```
benfred, line 199: I'm wondering if we can get rid of the _write_table_dispatch call entirely, and return an object where write_table works consistently between cpu/gpu versions.

We could either have an adaptor class that handles the conversion from pandas to pyarrow, or we could even just monkey patch the write_table call for the cpu version like this:
```suggestion
    ret = _cls(path, *_args, **kwargs)
    if isinstance(df, pd.DataFrame):
        ret.write_table = lambda df: _cls.write_table(ret, pa.Table.from_pandas(df, preserve_index=False))
    return ret
```

rjzamora: Sounds good to me :)

benfred (approved):

PR #878, opened 2021-06-10
Remove conda env files and install instructions

karlhigley:
After several days of troubleshooting, I haven't managed to get conda to create an environment that:
* has a version of Tensorflow with GPU support
* avoids numpy version conflicts between TF and cupy/cudf
* makes the NVT tests pass

At this point, I have concluded that supporting the creation of pre-baked NVT environments via conda
is not feasible. This PR removes the env files, Github action, and conda install instructions.


README.md
```diff
@@ -50,27 +50,6 @@ To install NVTabular, ensure that you meet the following prerequisites:
...
 **NOTE**: NVTabular will only run on Linux. Other operating systems are not currently supported.
 
-#### Installing NVTabular Using Conda
-
-NVTabular can be installed with Anaconda from the ```nvidia``` channel by running the following command:
-
-```
-conda install -c nvidia -c rapidsai -c numba -c conda-forge nvtabular python=3.7 cudatoolkit=10.2
-```
-
```
benfred, line 60: We shouldn't delete this part - we're still distributing nvtabular via conda.

benfred (commented):
Does everything work if we remove the deep learning frameworks (tensorflow/pytorch/fastai etc) from the conda environment files? 

I think people are still getting some value from these environment files (like in terms of getting NVT installed) - so I'm wondering if we can just provide the base and then let people layer on either TF or PT as needed afterwards

karlhigley:
The problem I'm having is that I can't layer on TF afterward. There's no combination of conda and pip I've been able to figure out that results in NVT and TF playing nicely together in a conda env.

If there's value in the env files related to installing NVT that isn't already covered by NVT's dependencies, seems like we should capture it there?

karlhigley:
Okay, it's not quite that bad. You can do it with manual steps.

benfred (approved):

karlhigley:
Closing in favor of #885.

PR #899, opened 2021-06-24
Prototype of a serving pipeline with candidate retrieval, filtering, scoring, and ordering

karlhigley:


karlhigley:
This POC has now served its purpose, and since it wasn't intended to be a user facing example or releasable code, it's now time to close this PR.

PR #900, opened 2021-06-24
Removes cudf dependency

albert17:
Removes cudf dependency #534

### DIFF
--- a/nvtabular/dispatch.py
+++ b/nvtabular/dispatch.py
@@ -19,7 +19,6 @@ from typing import Callable, Union
 
 import cupy as cp
 import dask.dataframe as dd
-import dask_cudf
 import numpy as np
 import pandas as pd
 import pyarrow as pa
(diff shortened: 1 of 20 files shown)


albert17:
@benfred Let me know what you think, if we want to keep going into this direction. Many work-arounds  and specific cases needed.

benfred:
> @benfred Let me know what you think, if we want to keep going into this direction. Many work-arounds and specific cases needed.

Lets just get a single unittest working - that verifies we can run a workflow without cudf . I don't think we need to update all the unittests to work without cudf for this release.  (like create a 'test_cpu_workflow.py' file that tests a basic workflow fit/transform cycle - and then we can run just this one unittest on github actions to verify our cpu support.)

marcromeyn:
I feel like knowing what libraries are available (like cudf/fask-cudf/etc.) could be useful in user-code as well. So maybe it would be good to create a few methods for this that we can expose? As inspiration we could look at how Huggingface transformers does this for the various ML-modeling libraries: https://github.com/huggingface/transformers/blob/master/src/transformers/file_utils.py#L287-L311.

albert17:
Got all the feedback, thanks @benfred @marcromeyn . I am going to create a new PR, I added too many changes here based on updating the unit tests.

PR #905, opened 2021-06-25
Allow ColumnGroup to be passed to get_embedding_sizes

### LINKED ISSUE #732: [FEA] Return embedding table shape by feeding colum group to get_embedding_sizes() function

In the current case, to get the embedding table shape, we need to do as follows:
```
cat_features = ['col'] >> Cateforify()
workflow = nvt.Workflow(cat_features) 
dataset = nvt.Dataset(gdf)
workflow.fit(dataset)
EMBEDDING_TABLE_SHAPES = nvt.ops.get_embedding_sizes(workflow)
```

but if we do have many other features from other ops, and not want to add `cat_features`  in the final workflow, we cannot get  embedding table shape from `nvt.ops.get_embedding_sizes(workflow)` because it would return {}.  

To avoid that can we change `def get_embedding_sizes(workflow):` to `def get_embedding_sizes(ColumnGroups):`
 or something like that so that we can run `nvt.ops.get_embedding_sizes(cat_features)` instead.

https://github.com/NVIDIA/NVTabular/blob/c7b7e05564ea675f00cad6e784cc360f8bb4ee85/nvtabular/ops/categorify.py#L449

benfred:
closes #732

### DIFF
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -444,7 +444,7 @@ def get_embedding_sizes(source, output_dtypes=None):
     output_dtypes : dict, optional
         Optional dictionary of column_name:dtype. If passing a workflow object dtypes
         will be read from the workflow. This is used to figure out which columns
-        are categorical, which are split out by this function. If passed a column_group
+        are multihot-categorical, which are split out by this function. If passed a column_group
         and this parameter isn't set, you won't have multihot columns returned separately
     """
     # TODO: do we need to distinguish multihot columns here?  (if so why? )
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -433,10 +433,33 @@ def _get_embedding_order(cat_names):
     return cat_names
 
 
-def get_embedding_sizes(workflow):
-    """Returns a dictionary of best embedding sizes from the workflow"""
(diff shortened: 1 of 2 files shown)


nvtabular/ops/categorify.py
```diff
@@ -433,23 +433,47 @@ def _get_embedding_order(cat_names):
...
+        Optional dictionary of column_name:dtype. If passing a workflow object dtypes
+        will be read from the workflow. This is used to figure out which columns
+        are categorical, which are split out by this function. If passed a column_group
+        and this parameter isn't set, you won't have multihot columns returned separately
+    """
     # TODO: do we need to distinguish multihot columns here?  (if so why? )
-    queue = [workflow.column_group]
+
+    # have to lazy import Workflow to avoid circular import errors
+    from nvtabular.workflow import Workflow
```
karlhigley, line 453: Can we avoid the circular import by checking `isinstance(source, ColumnGroup)` below?

karlhigley (approved):

nvtabular/ops/categorify.py
benfred: Looks like we'll still have circular import errors =( https://github.com/NVIDIA/NVTabular/blob/main/nvtabular/column_group.py#L20

PR #906, opened 2021-06-25
Add options dataclass to Categorify

benfred:
When fitting statistics in Categorify, we were passing around 12 different options
as parameters across multiple methods. This lead to long call signatures, as well as
making it hard to add new parameters. Also the parameter names changes between methods
sometimes (max_size -> max_emb_size, num_buckets -> nbuckets, agg_cols -> cont_cols)
making it hard to trace through and figure out whats going.

Refactor by moving all these parameters to a dataclass , and pass this class around
instead.


nvtabular/ops/categorify.py
```diff
@@ -520,26 +524,40 @@ def _make_name(*args, sep="_"):
     return sep.join(args)
 
 
+@dataclass
+class _FitOptions:
```
marcromeyn, line 528: Should we add some docs here?

benfred: I've added some brief descriptions of each parameters in the last commit .

karlhigley (approved):

benfred:
rerun tests

PR #912, opened 2021-06-28
Fix multihots for when not passed a dtype to Categorify

benfred:
We are seeing issues with Categorify when not passed a dtype with multihot
columns. For instance the movielens triton TF example fails with

 ```
 E0628 19:45:54.418425 1 model_repository_manager.cc:1430] Invalid argument: in ensemble movielens, ensemble tensor genres__values_nvt: inconsistent data type: TYPE_FP64 is inferred from model movielens_nvt while TYPE_INT64 is inferred from model movielens_tf
 ```

The problem seems to be that calling astype with a None value causes cudf to
change the dtype to int64. Fix.

### DIFF
--- a/nvtabular/dispatch.py
+++ b/nvtabular/dispatch.py
@@ -249,8 +249,10 @@ def _encode_list_column(original, encoded, dtype=None):
         return pd.Series(new_data)
     else:
         # CuDF version
+        encoded = as_column(encoded)
+        if dtype:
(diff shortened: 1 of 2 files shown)


benfred:
rerun tests

benfred:
rerun tests

karlhigley (approved):

PR #915, opened 2021-06-29
Use a single triton model.py for tf/torch/hugectr

benfred:
Rather than have 3 separate triton models for each framework, combine them all
and just use a single model.py. This adds a 'model_output' parameter to the
triton config file, and uses that to output tensors in the format that tf/hugectr
expects.

This also unifies the representation for pytorch to be the same as for tensorflow,
since we've changed the dataloader format recently - concatenated the conts/cats
tensors for pytorch is no longer correct.


benfred:
rerun tests

jperez999 (approved):

PR #918, opened 2021-07-06
Add 'supports' and 'inference_initialize' for operators.

benfred:
This is the first in a series of PR's for providing optimized inference.

This adds both a 'supports' flag for operators and a 'inference_initialize' method.

The supports flag shows what dataformats the operator can accept during transforms.
We've found for inference that a dictionary of column names to numpy/cupy arrays
can be orders of magnitude faster at inference time. As an example, the normalize operator
works unchanged on these dictionaries, and times for me go from 8ms to around 140us on the
Rossmann dataset using dictionaries of numpy arrays instead of cudf or pandas dataframes.

The inference_initialize change provides a mechanism for operators to provide an optimized
version for inference, as well as to preload any necessary data. This is currently only
serving to preload categories in the categorify op - but will be used to provide optimize
versions of operators in a future PR.

### DIFF
--- a/nvtabular/inference/triton/model.py
+++ b/nvtabular/inference/triton/model.py
@@ -48,6 +48,19 @@ from nvtabular.inference.triton import _convert_tensor, get_column_types
 class TritonPythonModel:
     """Generic TritonPythonModel for nvtabular workflows"""
 
+    def _initialize_ops(self, column_group, initialized):
+        if column_group.op:
+            inference_op = column_group.op.inference_initialize(
+                column_group.columns, self.model_config
+            )
+            if inference_op:
+                column_group.op = inference_op
+
+        for parent in column_group.parents:
+            if parent not in initialized:
+                initialized.add(parent)
+                self._initialize_ops(parent, initialized)
+
     def initialize(self, args):
         workflow_path = os.path.join(
             args["model_repository"], str(args["model_version"]), "workflow"
@@ -55,6 +68,10 @@ class TritonPythonModel:
         self.workflow = nvtabular.Workflow.load(workflow_path)
         self.model_config = json.loads(args["model_config"])
         self.output_model = self.model_config["parameters"]["output_model"]["string_value"]
+
+        # recurse over all column groups, initializing operators for inference pipeline
+        initialized_column_groups = set()
+        self._initialize_ops(self.workflow.column_group, initialized_column_groups)
 
         self.input_dtypes = {
             col: dtype
--- a/nvtabular/inference/triton/model.py
+++ b/nvtabular/inference/triton/model.py
@@ -48,7 +48,10 @@ from nvtabular.inference.triton import _convert_tensor, get_column_types
 class TritonPythonModel:
     """Generic TritonPythonModel for nvtabular workflows"""
 
-    def _initialize_ops(self, column_group, initialized):
+    def _initialize_ops(self, column_group, visited=None):
+        if visited is None:
+            visited = set()
+
         if column_group.op:
             inference_op = column_group.op.inference_initialize(
                 column_group.columns, self.model_config
@@ -57,9 +60,9 @@ class TritonPythonModel:
                 column_group.op = inference_op
 
         for parent in column_group.parents:
-            if parent not in initialized:
-                initialized.add(parent)
-                self._initialize_ops(parent, initialized)
+            if parent not in visited:
+                visited.add(parent)
+                self._initialize_ops(parent, visited)
 
     def initialize(self, args):
         workflow_path = os.path.join(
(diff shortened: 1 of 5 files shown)


nvtabular/inference/triton/model.py
```diff
@@ -48,6 +48,19 @@
...
+        if column_group.op:
+            inference_op = column_group.op.inference_initialize(
+                column_group.columns, self.model_config
+            )
+            if inference_op:
+                column_group.op = inference_op
+
+        for parent in column_group.parents:
+            if parent not in initialized:
+                initialized.add(parent)
```
karlhigley, line 61: The name of the `initialized` parameter is a little confusing, because the `column_group`'s op hasn't actually been initialized until after its added to the `initialized` set. Maybe something like `visited` would be clearer here?

karlhigley: I suppose the problem with giving it a default value is that the default is only evaluated once resulting in a single mutable set being shared across invocations. Could the default be `None` and then the body of the method would create an empty set to use if so? That would make it a little easier to use from the perspective of the calling code.

nvtabular/inference/triton/model.py
```diff
@@ -56,6 +69,10 @@ def initialize(self, args):
         self.model_config = json.loads(args["model_config"])
         self.output_model = self.model_config["parameters"]["output_model"]["string_value"]
 
+        # recurse over all column groups, initializing operators for inference pipeline
+        initialized_column_groups = set()
+        self._initialize_ops(self.workflow.column_group, initialized_column_groups)
```
karlhigley, line 74: The potential default value change mentioned above would make this:
```suggestion
        self._initialize_ops(self.workflow.column_group)
```

nvtabular/ops/categorify.py
```diff
@@ -421,6 +422,14 @@ def get_embedding_sizes(self, columns):
             self.categories, columns, self.num_buckets, self.freq_threshold, self.max_size
         )
 
+    def inference_initialize(self, columns: ColumnNames, model_config: dict) -> Optional[Operator]:
+        # on the first transform call we load up categories from disk, which can
+        # take multiple seconds. preload this data by running an empty dataframe through
```
karlhigley, line 427: This feels like a long way around to load the data during initialization. Is there something about loading the categories from disk that we wouldn't want to do during the normal `__init__` method for non-inference use cases?

nvtabular/ops/operator.py
```diff
@@ -26,6 +27,19 @@
...
 
+class Supports(Flag):
+    """ Indicates what type of data representation this operator supports. """
+
+    # cudf dataframe
+    CPU_DATAFRAME = auto()
+    # pandas dataframe
+    GPU_DATAFRAME = auto()
+    # dict of column name to numpy array
+    CPU_ARRAY = auto()
```
karlhigley, line 38: The names of these array constants seem a little confusing, because the format is actually a `dict`. Maybe `CPU_DICT_ARRAY` or something like that?

nvtabular/ops/operator.py
```diff
@@ -83,3 +97,13 @@ def __rrshift__(self, other) -> ColumnGroup:
...
         return self.__class__.__name__
+
+    @property
+    def supports(self) -> Supports:
+        """ Returns what kind of data representation this operator supports """
+        return Supports.CPU_DATAFRAME | Supports.GPU_DATAFRAME
+
+    def inference_initialize(self, columns: ColumnNames, model_config: dict) -> Optional[Operator]:
+        """Configures this operator for use in inference. May return a different operator to use
+        instead of the one configured during training"""
```
karlhigley, line 108:
```suggestion
        instead of the one configured for use during training"""
```

tests/unit/test_ops.py
```diff
@@ -293,6 +293,14 @@ def test_normalize(tmpdir, df, dataset, gpu_memory_frac, engine, op_columns):
         ]
         assert np.all((df[col] - new_gdf[col]).abs().values <= 1e-2)
 
+    # our normalize op also works on dicts of cupy/numpy tensors. make sure this works like we'd
```
karlhigley, line 296: I can't throw any stones here because I've done it too (recently even), but adding additional assertions inside existing tests seems like the kind of thing it would be easy to miss as we add more operators that support this format. I wonder if this might (eventually?) be worth either parametrizing or capturing in a separate set of tests.

karlhigley (commented):
This looks okay to me in the sense that I don't see anything I'd find concerning on a technical level. Left some comments about naming and API design, and about the use of `inference_initialize` in `Categorify`. This usage seems a little weird to me, but since this is the first in a series of PRs, I'm assuming there are additional use cases for `inference_initialize` yet to come.

karlhigley (approved):
Approving based on a Slack conversation indicating that the specific operator changes in this PR are mostly for demonstration and will be changed further in future PRs. Comments re: naming etc are still relevant to consider, but nothing that would stop us from merging.

nvtabular/ops/categorify.py
benfred: agreed - this method will be swapped out in a future PR

nvtabular/inference/triton/model.py
benfred: done

PR #924, opened 2021-07-07
Adding LocalCUDACluster to avoid OOM

bschifferer:
Providing a troubleshoot section to avoid OOM errors by using LocalCUDACluster


docs/source/resources/troubleshooting.md
```diff
@@ -55,3 +59,21 @@ def _memory_usage(df):
     size += df.index.memory_usage(deep=True)
     return size
 ```
+
+### 2. Initializing a Dask CUDA Cluster
+
+Even if you execute the NVTabular workflow with only a single GPU, using a dask CUDA cluster can optimize your memory usage. If there is no CUDA cluster initialized, NVTabular executes the workflow on workers of a dask cluster. Dask CUDA cluster provides different memory spilling strategies for GPUs and in our experiment, we could resolve OOM errors by initializing a dask CUDA cluster, first. It is easy to provide NVTabular with a dask CUDA cluster by using `LocalCUDACluster`.
```
benfred, line 65: > If there is no CUDA cluster initialized, NVTabular executes the workflow on workers of a dask cluster

I don't think this is quite accurate - if we don't pass a dask client we run the workflow on the default 'synchronous' scheduler inside the current process (instead of on a dask cluster).

docs/source/resources/troubleshooting.md
```diff
@@ -55,3 +59,21 @@ def _memory_usage(df):
...
+
+import nvtabular as nvt
+
+cluster = LocalCUDACluster()
+client = Client(cluster)
+
+features = ['col'] >> nvt.ops.Normalize()
+
+workflow = nvt.Workflow(features, client=client)
+```
```
benfred, line 79: we should add something about cleaning up the client (```client.shutdown()```) after running the workflow

bschifferer: Is it `client.shutdown()` or `client.close()`?

I normally use client.close()

docs/source/resources/troubleshooting.md
bschifferer: Thank you. I am sorry, I do not know it that depth :) Does it work to update the sentence with

```
If there is no CUDA cluster initialized, NVTabular's default behaviour is to execute the workflow with a synchronous scheduler inside the current process.
```

docs/source/resources/troubleshooting.md
rjzamora: After all work is completed, you would do something like:
```python
client.shutdown()
client.close()
```

docs/source/resources/troubleshooting.md
```diff
@@ -55,3 +59,21 @@ def _memory_usage(df):
...
+
+Even if you execute the NVTabular workflow with only a single GPU, using a dask CUDA cluster can optimize your memory usage. If there is no CUDA cluster initialized, NVTabular executes the workflow on workers of a dask cluster. Dask CUDA cluster provides different memory spilling strategies for GPUs and in our experiment, we could resolve OOM errors by initializing a dask CUDA cluster, first. It is easy to provide NVTabular with a dask CUDA cluster by using `LocalCUDACluster`.
+
+```python
+from dask_cuda import LocalCUDACluster
+from dask.distributed import Client
+
+import nvtabular as nvt
+
+cluster = LocalCUDACluster()
```
rjzamora, line 73:
```suggestion
cluster = LocalCUDACluster(
    n_workers=1,  # Number of GPU workers
    device_memory_limit=“24GB”,  # GPU->CPU spill threshold (~75% of GPU memory)
    local_directory=“/nvme/scratch/”,   # Fast directory for disk spilling
)
```

docs/source/resources/troubleshooting.md
rjzamora, line 65: Possible alternative (no worries if you want to pick and choose some/none of it):

```suggestion
Even if you only have a single GPU to work with, it is best practice to use a distributed Dask-CUDA cluster to execute memory-intensive NVTabular workflows. If there is no distributed `client` object passed to an NVTabular `Workflow`, it will fall back on Dask’s single-threaded “synchronous” scheduler at computation time. The primary advantage of using a Dask-CUDA cluster is that the Dask-CUDA workers enable GPU-aware memory spilling.  In our experience, many OOM errors can be resolved by initializing a dask-CUDA cluster with an appropriate `device_memory_limit` setting, and by passing a corresponding client to NVTabular.  It is easy to deploy a single-machine dask-CUDA cluster using `LocalCUDACluster`.
```

rjzamora (commented):
Thanks for putting this together @bschifferer - I added some minor suggestions.

docs/source/resources/troubleshooting.md
bschifferer: Updated with Rick's suggestion

docs/source/resources/troubleshooting.md
bschifferer: Added both shutdown and close

docs/source/resources/troubleshooting.md
bschifferer: Added sugestion

docs/source/resources/troubleshooting.md
bschifferer: Added suggestion

docs/source/resources/troubleshooting.md
```diff
@@ -55,3 +59,28 @@ def _memory_usage(df):
...
+from dask_cuda import LocalCUDACluster
+from dask.distributed import Client
+
+import nvtabular as nvt
+
+cluster = LocalCUDACluster(
+    n_workers=1,                        # Number of GPU workers
+    device_memory_limit="24GB",         # GPU->CPU spill threshold (~75% of GPU memory)
+    local_directory="/nvme/scratch/",   # Fast directory for disk spilling
+)
```
rjzamora, line 77:
```suggestion
cluster = LocalCUDACluster(
    n_workers=1,                        # Number of GPU workers
    device_memory_limit="24GB",         # GPU->CPU spill threshold (~75% of GPU memory)
    rmm_pool_size="28GB",                   # Memory pool size on each worker
    local_directory="/nvme/scratch/",   # Fast directory for disk spilling
)
```

rjzamora (approved):

rjzamora:
We should probably add a section about the `output_files` option in `to_parquet` (since it forces depth-first graph execution for one input partition at a time; dramatically reducing memory pressure), but I can add it in a follow-up PR.

benfred (approved):

PR #925, opened 2021-07-07
fix notebook criteo

jperez999:
fixes criteo OOM error


examples/scaling-criteo/02-ETL-with-NVTabular.ipynb
```diff
@@ -309,7 +300,7 @@
     "\n",
     "# Deploy a Single-Machine Multi-GPU Cluster\n",
     "protocol = \"tcp\"  # \"tcp\" or \"ucx\"\n",
-    "NUM_GPUS = [0, 1, 2, 3, 4, 5, 6, 7]\n",
+    "NUM_GPUS = [0] #[0, 1, 2, 3, 4, 5, 6, 7]\n",
```
benfred, line 303: Did you just run this on a single gpu? Should we revert this line?

 Looks like the fit times are about 2x previous - and the transform time is about 4.5x previous below

benfred (commented):
Nice work finding out whats causing the OOM errors!

I'd like to figure out *why* this num_files_per_proc parameter has this effect though.  I don't understand why this parameter would cause an OOM. Especially because it also seems like this is related to the normalize op (?)

rjzamora:
>I don't understand why this parameter would cause an OOM.

I just looked into this and it seems it has everything to do with the difference between [_add_data_scatter](https://github.com/NVIDIA/NVTabular/blob/14111c8b8b759b77580e89de567e39ecfb19761d/nvtabular/io/writer.py#L153) (used for multi-file output), and [_add_single_file](https://github.com/NVIDIA/NVTabular/blob/14111c8b8b759b77580e89de567e39ecfb19761d/nvtabular/io/writer.py#L216) (used otherwise).  When you change `num_files_per_proc=5` to `num_files_per_proc=1`, you effectively switch from the former to the latter.

So, why does `_add_single_file` use more memory than `_add_data_scatter`?   **The answer is that `_add_data_scatter` does not really shuffe the order of the rows when `shuffle=True`,  but `_add_single_file` does**.  As far as I can tell, we can reduce the memory requirements of [`_shuffle_df`](https://github.com/NVIDIA/NVTabular/blob/14111c8b8b759b77580e89de567e39ecfb19761d/nvtabular/io/shuffle.py#L52) (used by `_add_single_file`) by leveraging the `sample` API.  However, a proper shuffle still requires device memory ~2.6x the partition size (in contrast to the ~1.8x required by `_add_data_scatter`).

**Take-away**:  We can reduce the memory usage a bit for `num_files_per_proc=1`, but it may actually be considered a "bug" that `num_files_per_proc=5` memory usage is as low as it currently is.  For the multi-file case we are randomly shuffling the data between output files, but we are **not** reordering the data within those files.  For example, a column with data like `[0, 1, 2, 3, 4, 5]` might be split between two files as `[1, 2, 4]` and `[0, 3, 5]` (note that the data order is not shuffled within the file).

In general, it seems like the "proper" fix here is to:
- Use a smaller partition size in the Criteo example.  Alternatively, we may want to consider using the `output_files=` API for the `to_parquet` operation.  This approach should isolate the subgraph for each output file in such a way that *may* be more memory efficient at run time.
- "Fix" `_add_data_scatter` to properly shuffle the rows (although I am not completely sure intra-file shuffling is always necessary...)
- Update `_shuffle_df` to use `df.sample` for more-efficient partition shuffling

examples/scaling-criteo/02-ETL-with-NVTabular.ipynb
```diff
@@ -551,14 +491,15 @@
...
-      "Wall time: 1min 58s\n"
+      "CPU times: user 7.59 s, sys: 2.1 s, total: 9.7 s\n",
+      "Wall time: 9min 28s\n"
      ]
     }
    ],
    "source": [
     "%%time\n",
     "workflow.transform(train_dataset).to_parquet(\n",
+    "    output_files=5,\n",
```
rjzamora, line 502: We might want to explicitly set the number of output files to be the same as the number of workers.  Since the `output_files` approach is forcing a depth-first execution of each output-file subgraph, an 8-GPU system will not be fully utilized if this is set to `5`.

examples/scaling-criteo/02-ETL-with-NVTabular.ipynb
```diff
@@ -144,8 +144,8 @@
    "outputs": [],
    "source": [
     "# define some information about where to get our data\n",
-    "BASE_DIR = os.environ.get(\"BASE_DIR\", \"/raid/data/criteo\")\n",
-    "INPUT_DATA_DIR = os.environ.get(\"INPUT_DATA_DIR\", BASE_DIR + \"/converted/criteo\")\n",
+    "BASE_DIR = os.environ.get(\"BASE_DIR\", \"/raid/criteo/tests\") #\"/raid/data/criteo\"\n",
+    "INPUT_DATA_DIR = os.environ.get(\"INPUT_DATA_DIR\", BASE_DIR + \"/crit_int_pq\") #\"/converted/criteo\"\n",
```
rjzamora, line 148: Should this change be reverted?

examples/scaling-criteo/02-ETL-with-NVTabular.ipynb
```diff
@@ -180,7 +180,10 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "converted  test_dask\n"
+      "Miniconda3-latest-Linux-x86_64.sh  crit_test  model_size.json\n",
+      "benf\t\t\t\t   ctr\t      model_size_full.json\n",
+      "crit_fl_pq\t\t\t   datagen    test_dask\n",
+      "crit_int_pq\t\t\t   jp_movie   tf\n"
```
rjzamora, line 186: I assume we don't want all this output?

jperez999:
rerun tests

examples/scaling-criteo/02-ETL-with-NVTabular.ipynb
```diff
@@ -351,49 +317,6 @@
...
-    "    rmm.reinitialize(\n",
-    "        # RMM may require the pool size to be a multiple of 256.\n",
-    "        pool_allocator=True,\n",
-    "        initial_pool_size=(device_pool_size // 256) * 256,  # Use default size\n",
-    "    )\n",
-    "\n",
-    "\n",
-    "client.run(_rmm_pool)"
-   ]
-  },
```
benfred, line 396: Did you mean to delete this section ?

rjzamora: He addeed the `rmm_pool_size` argument to `LocalCUDACluster`, which makes this unncessary.  I'm not sure if we still need a sentence explaining that the rmm pool is important.  Perhaps we do, since we are still not using a "default" setting.

benfred (approved):

jperez999:
rerun tests

PR #926, opened 2021-07-07
Use C++ operators for categorify/fill during inference

benfred:


oyilmaz-nvidia (commented):
I am not sure about the benefit of writing the cpu transform functions in C++ and calling them from python. Did you have a chance to measure the running time of the same functions written in python? Maybe, we can just compare one of the functions. Writing all of them might take significant amount of time.

I can definitely see the benefit of writing these functions in C++ and calling them directly from C++ backend. Considering python is written with CPython, what we have here seems like what python has already been doing (calling C++ functions behind the scenes).

benfred:
> I am not sure about the benefit of writing the cpu transform functions in C++ and calling them from python. Did you have a chance to measure the running time of the same functions written in python?

Yeah - this PR is pretty essential in order to get the run times of the Criteo and Rossmann examples down.  Here is a graph comparing the times on the rossmann dataset to the times on the main branch (both running using the C++ backend):
operators

Its not the language that the operators are written in so much - its that our existing operators aren't optimized for running in small batch sizes . I tried calling this out in the design doc https://docs.google.com/document/d/1YbJH0OiSRNWrnsNe1RIo7-yrBQDSVajZZgyy5f1U7Oc/edit . This PR gets us to almost everything listed in the 'proposed changes section', aside from the C++ workflow code calling these operators directly which slipped this release.

benfred:
rerun tests

oyilmaz-nvidia:
@benfred I guess I wasn't clear enough. I was just curious about running time comparison of the cpu python transform functions (we don't have this, I guess) versus cpu C++ transform functions that are called from another python functions. I am just saying that If we were gonna call these cpu C++ code from python, we could have implemented directly in python. I was originally thinking to call these cpu C++ transform functions directly from C++ backend (without going to python at all).

Anyways, we definitely need this PR to get the required running times and we don't have time to test anything else.

oyilmaz-nvidia (approved):
Approving since we have only a day left for the release.

PR #935, opened 2021-07-12
Allowing to pass in a vocab in Categorify

### LINKED ISSUE #763: [BUG] make_feature_column_workflow doesn't work with categorical columns
Since the API overhaul in 0.4 - the make_feature_column_workflow function doesn't work with categorical data.

The problem is that its trying to directly set the category statistics on the workflow https://github.com/NVIDIA/NVTabular/blob/8b834310680154c102237b14cd36c165aa5c4cff/nvtabular/framework_utils/tensorflow/feature_column_utils.py#L292-L301

We should update to using a fit categorify op - and add a test to catch this in the future

marcromeyn:
This also fixes #763.


benfred:
rerun tests

.gitignore
```diff
@@ -71,3 +71,5 @@ ipython_config.py
 .dmypy.json
 dmypy.json
 
+# PyCharm
+.idea
```
karlhigley, line 75:
```suggestion
.idea

```

nvtabular/ops/categorify.py
```diff
@@ -285,6 +287,9 @@ def __init__(
                 "with this num_buckets setting!"
             )
 
+        self.vocabs = self.process_vocabs(vocabs if vocabs is not None else {})
```
karlhigley, line 290: Would simplify the calling code a bit if the check for `None` and associated default happened at the top of the `process_vocabs` method

nvtabular/ops/categorify.py
```diff
@@ -320,31 +325,69 @@ def fit(self, columns: ColumnNames, ddf: dd.DataFrame):
...
+            os.makedirs(base_path, exist_ok=True)
+            for col in list(vocabs.columns):
+                col_df = vocabs[[col]]
+                if col_df[col].iloc[0] is not None:
+                    if isinstance(col_df, pd.DataFrame):
+                        col_df = pd.DataFrame(
+                            {col: pd.concat([pd.Series([None]), col_df[col]]).reset_index()[0]}
+                        )
+                    else:
+                        import cudf
```
karlhigley, line 357: @benfred Is this consistent with the way y'all are handling imports in the CPU-only changes?

nvtabular/ops/categorify.py
```diff
@@ -320,31 +325,69 @@ def fit(self, columns: ColumnNames, ddf: dd.DataFrame):
...
+        else:
+            error = """Unrecognized vocab type,
+            please provide either a dictionary with paths to a parquet files
+            or a DataFrame that contains the vocabulary per column.
+            """
+            raise ValueError(error)
+
+        return categories
+
+    def _create_fit_options_from_columns(self, columns) -> "FitOptions":
```
karlhigley, line 377: I like the extract refactor here 👍🏻

nvtabular/ops/categorify.py
benfred: I think we can hide some of this complexity here. We have a 'concat' function in nvtabular.dispatch that does much the same here.

https://github.com/NVIDIA/NVTabular/blob/9f25d42932f107528882c91ccae77906b9f9502f/nvtabular/dispatch.py#L307-L311
 
I think we just need to add something to do the same as _make_df (https://github.com/NVIDIA/NVTabular/blob/9f25d42932f107528882c91ccae77906b9f9502f/nvtabular/dispatch.py#L314-L321) to make the ```pd.Series[None]``` or ```cudf.Series([None])``` for the initial value.

benfred (commented):
this looks good! 

There are a couple of minor things I'd like to see before merging:

1) I think we should probably move the pandas/cudf code into nvtabular/dispatch.py
2) We should add an entry to the Categorify docstring  for the  'vocabs' parameter
3) I don't think this will work with encode_type='joint' (like when using this op to compute feature crosses). I'm not too concerned about this - but maybe we can throw an exception in the constructor if both vocabs and encode_type='joint' are both set?

nvtabular/ops/categorify.py
```diff
@@ -238,8 +228,10 @@ def __init__(
         # Only support two kinds of multi-column encoding
         if encode_type not in ("joint", "combo"):
             raise ValueError(f"encode_type={encode_type} not supported.")
+        if encode_type == "joint" and vocabs:
```
benfred, line 231: Maybe something like 
```suggestion
        if encode_type == "joint" and vocabs is not None:
```

instead?

We're seeing unittest failures on this line with errors like ```E       ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all()```

tests/unit/test_tf_feature_columns.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+
+tf = pytest.importorskip("tensorflow")
+nvtf = pytest.importorskip("nvtabular.framework_utils.tensorflow")
+
+
+def test_feature_column_utils():
```
benfred, line 7: nice! I like having a test here - the feature column utils was a notable gap in our code coverage

benfred (approved):

nvtabular/ops/categorify.py
```diff
@@ -238,8 +228,10 @@ def __init__(
         # Only support two kinds of multi-column encoding
         if encode_type not in ("joint", "combo"):
             raise ValueError(f"encode_type={encode_type} not supported.")
+        if encode_type == "joint" and vocabs is not None:
+            raise ValueError("Passing in vocabs is not supported with a joint encoding.")
```
benfred, line 232: I messed up in my comment there - the problem is with 'combo' encodings not joint (like combo is the feature crosses here that we don't support - joint is all good). 

```suggestion
        if encode_type == "combo" and vocabs is not None:
            raise ValueError("Passing in vocabs is not supported with a combo encoding.")
```

benfred:
rerun tests

marcromeyn:
@benfred The build is failing now with `AttributeError: module 'versioneer' has no attribute 'get_cmdclass'` which is unrelated to the changes in this PR. Any idea how to resolve this?

benfred:
> @benfred The build is failing now with AttributeError: module 'versioneer' has no attribute 'get_cmdclass' which is unrelated to the changes in this PR. Any idea how to resolve this?

I a different PR I had to change the cpu CI to work around this (https://github.com/NVIDIA/NVTabular/pull/926/files#diff-f48959cf62357a95aff9d53b6b9cdbbf3aa81317d24eb321c296a7d9fd6866b7R43).  I've pushed the same change here to see if it helps

benfred:
rerun tests

benfred:
rerun tests

benfred:
@marcromeyn The unittests are failing here with what looks like an off by one error ```E           assert [[0], [0, 3], [1, 2], [2]] == [[1], [1, 4], [2, 3], [3]] ``` .  The tests sorta assume that the index '0' is reserved for out of vocabulary/ unknown items - is this being handled?

marcromeyn:
rerun tests

nvtabular/framework_utils/tensorflow/feature_column_utils.py
```diff
@@ -227,7 +226,7 @@ def _get_parents(column):
         features += features_replaced_buckets
 
     if len(categorifies) > 0:
-        features += categorifies.keys() >> Categorify()
+        features += categorifies.keys() >> Categorify(vocabs=pd.DataFrame(categorifies))
```
diggerk, line 229: This line fails if features have vocabularies of varying size. In particular, the sample notebook "using-feature-columns.ipynb" fails on this line with the following error. Am I missing anything, or this is a bug here?

I used docker image "merlin-tensorflow-training" that had nvtabular 0.6.1 installed.

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_783/4041583971.py in <module>
----> 1 online_workflow, feature_columns = make_feature_column_workflow(feature_columns, "AdoptionSpeed")

/nvtabular/nvtabular/framework_utils/tensorflow/feature_column_utils.py in make_feature_column_workflow(feature_columns, label_name, category_dir)
    227 
    228     if len(categorifies) > 0:
--> 229         features += categorifies.keys() >> Categorify(vocabs=pd.DataFrame(categorifies))
    230 
    231     if len(hashes) > 0:

/usr/local/lib/python3.8/dist-packages/pandas/core/frame.py in __init__(self, data, index, columns, dtype, copy)
    466 
    467         elif isinstance(data, dict):
--> 468             mgr = init_dict(data, index, columns, dtype=dtype)
    469         elif isinstance(data, ma.MaskedArray):
    470             import numpy.ma.mrecords as mrecords

/usr/local/lib/python3.8/dist-packages/pandas/core/internals/construction.py in init_dict(data, index, columns, dtype)
    281             arr if not is_datetime64tz_dtype(arr) else arr.copy() for arr in arrays
    282         ]
--> 283     return arrays_to_mgr(arrays, data_names, index, columns, dtype=dtype)
    284 
    285 

/usr/local/lib/python3.8/dist-packages/pandas/core/internals/construction.py in arrays_to_mgr(arrays, arr_names, index, columns, dtype, verify_integrity)
     76         # figure out the index, if necessary
     77         if index is None:
---> 78             index = extract_index(arrays)
     79         else:
     80             index = ensure_index(index)

/usr/local/lib/python3.8/dist-packages/pandas/core/internals/construction.py in extract_index(data)
    395             lengths = list(set(raw_lengths))
    396             if len(lengths) > 1:
--> 397                 raise ValueError("arrays must all be same length")
    398 
    399             if have_dicts:

ValueError: arrays must all be same length
```

karlhigley: Thoughts on how to handle this @marcromeyn?

benfred: Thanks for reporting! That looks like a bug to me - I opened https://github.com/NVIDIA/NVTabular/issues/1062 to track

PR #936, opened 2021-07-12
Update index.rst

lgardenhire:


benfred:
I've merged commit into https://github.com/NVIDIA/NVTabular/pull/934 (neither PR will build without the other - so need to have both changes in a single PR to pass the build checks).

PR #937, opened 2021-07-12
Update shuffle algorithm

rjzamora:
This PR makes the changes to `_shuffle_df` and `_add_data_scatter` that were discussed/motivated in [this comment](https://github.com/NVIDIA/NVTabular/pull/925#issuecomment-876678067) of #925

Outcome:

- `_shuffle_df` is slightly more efficient
- `_add_data_scatter` now results in re-ordered data when `shuffle` is is not False

### DIFF
--- a/nvtabular/io/shuffle.py
+++ b/nvtabular/io/shuffle.py
@@ -15,8 +15,11 @@
 #
 import enum
 import warnings
+from distutils.version import LooseVersion
 
 import pandas as pd
+
+_IGNORE_INDEX_SUPPORTED = pd.__version__ >= LooseVersion("1.3.0")
 
 
 class Shuffle(enum.Enum):
@@ -52,6 +55,12 @@ def _shuffle_df(df, size=None, keep_index=False):
     ordered rows"""
     size = size or len(df)
     if isinstance(df, pd.DataFrame):
-        return df.sample(n=size, ignore_index=not keep_index)
+        if _IGNORE_INDEX_SUPPORTED:
+            return df.sample(n=size, ignore_index=not keep_index)
+        else:
+            # Pandas<1.3.0
+            if keep_index:
+                return df.sample(n=size)
+            return df.sample(n=size).reset_index(drop=True)
     else:
         return df.sample(n=size, keep_index=keep_index)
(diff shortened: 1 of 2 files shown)


karlhigley (approved):
LGTM

(It's still weird to me that the canonical way to shuffle is to sample all the rows, but 🤷🏻 )

benfred (approved):

nvtabular/io/shuffle.py
```diff
@@ -53,21 +47,11 @@ def _check_shuffle_arg(shuffle):
...
     """Shuffles a DataFrame, returning a new dataframe with randomly
     ordered rows"""
     size = size or len(df)
-    # NOTE: We can use np.arange for both gpu and cpu-backed
-    # dataframes once NEP-35 is fully accepted (`like` argument).
-    # This should be available in numpy>=1.20
     if isinstance(df, pd.DataFrame):
-        arr = np.arange(size)
-        np.random.shuffle(arr)
+        return df.sample(n=size, ignore_index=not keep_index)
```
benfred, line 55: we're seeing a unittest failure here
```
E           TypeError: sample() got an unexpected keyword argument 'ignore_index'
```

I see this parameter on the pandas 1.3.0 docs https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sample.html - but I don't see it on my pandas 1.2.4 install. Do we need to bump the version of pandas on our containers for this ?

rjzamora: You're right, it was added in 1.3 - Grr
Bumping the pandas version would resolve the issue, but we could also add a `tr`y block or a version check to handle older pandas versions.  Any preference?

benfred: cudf claims not to support pandas 1.3 so I'm not sure if we can bump the version after all =(
https://github.com/rapidsai/cudf/blob/7823a181bfbffcfb63375d8f38506d0a20696a74/conda/recipes/cudf/meta.yaml#L41 

Maybe check the pandas version to see if we can enable ```distutils.version.LooseVersion(pd.__version__) > "1.3"``` ?

PR #938, opened 2021-07-12
CPU Support Doc

albert17:
CPU Support Doc

### DIFF
--- a/docs/source/core_features.md
+++ b/docs/source/core_features.md
@@ -80,3 +80,4 @@ NVTabular support CPU using [pandas](https://pandas.pydata.org/), [pyarrow](http
 ```
 dataset = Dataset(path, cpu=True)
 ```
+Setting this option on the Dataset will cause all processing on that dataset to happen on the CPU. This includes feature engineering and preprocessing as well as training models on TensorFlow or PyTorch using NVTabular's dataloaders.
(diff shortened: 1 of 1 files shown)


docs/source/core_features.md
```diff
@@ -72,3 +73,10 @@ NVTabular makes it possible to shuffle during dataset creation. This creates a u
 ## Cloud Integration ##
 
 NVTabular offers cloud integration with Amazon Web Services (AWS) and Google Cloud Platform (GCP), giving you the ability to build, train, and deploy models on the cloud using datasets. For additional information, see [Amazon Web Services](./resources/cloud_integration.md#amazon-web-services) and [Google Cloud Platform](./resources/cloud_integration.md#google-cloud-platform).
+
+## CPU Support ##
+
+NVTabular support CPU using [pandas](https://pandas.pydata.org/), [pyarrow](https://arrow.apache.org/docs/python/), and [dask dataframe](https://examples.dask.org/dataframe.html). To enable CPU, the Dataset must be initialized with the `cpu` parameter as follows.
+```
+dataset = Dataset(path, cpu=True)
+```
```
benfred, line 82: Looks good! Maybe we should add a note that setting this option will let you do both feature engineering and training on the CPU? Something like :
```suggestion
```

Setting this option on the Dataset will cause all processing on that dataset to happen on the CPU. This includes feature engineering and preprocessing as well as training models on TensorFlow or PyTorch using NVTabular's dataloaders. 
```

albert17: Perfect. I will add it

albert17:
Added

benfred (approved):

PR #942, opened 2021-07-13
Update core_features.md

lgardenhire:


docs/source/core_features.md
```diff
@@ -76,8 +76,9 @@ NVTabular offers cloud integration with Amazon Web Services (AWS) and Google Clo
 
 ## CPU Support ##
 
-NVTabular support CPU using [pandas](https://pandas.pydata.org/), [pyarrow](https://arrow.apache.org/docs/python/), and [dask dataframe](https://examples.dask.org/dataframe.html). To enable CPU, the Dataset must be initialized with the `cpu` parameter as follows.
+NVTabular supports CPU using [pandas](https://pandas.pydata.org/), [pyarrow](https://arrow.apache.org/docs/python/), and [dask dataframe](https://examples.dask.org/dataframe.html). To enable CPU, the dataset must be initialized with the `cpu` parameter as follows:
```
benfred, line 79: Nice catch on the plural / punctuation!

However, Dataset in this case is referring to the Dataset class - and should be capitalized :

```suggestion
NVTabular supports CPU using [pandas](https://pandas.pydata.org/), [pyarrow](https://arrow.apache.org/docs/python/), and [dask dataframe](https://examples.dask.org/dataframe.html). To enable CPU, the Dataset must be initialized with the `cpu` parameter as follows:
```

lgardenhire: @benfred - In that case, we need to be specific and say "Dataset class" for context.

benfred (approved):

PR #956, opened 2021-07-15
Improve shuffling behavior for output_files approach in to_parquet

rjzamora:
This PR is motivated by an off-line discussion with @benfred - We noticed that:

1. **The default behavior of `Dataset.to_parquet` should likely be the `output_files`-based approach (rather than `out_files_per_proc=1`)**.   This is because the `output_files` algorithm was  designed in a way that turns out to be much more efficient with memory.  There **can** certainly be performance and dashboard/viz disadvantages to this approach, but the improved memory stability is a pretty-significant advantage.
2.  **If we do use `output_files` by default, we will want to enable the same multi-file pseudo-shuffling procedure as the one used for `out_files_per_proc`.**

This PR is intended to address point (2).  Here, we allow a task/process to shuffle an input partition between files by repurposing the `out_files_per_proc` argument whenever an integer or file list is specified for `output_files`.  In these cases, we interpret `out_files_per_proc` to mean the number of files we should write within the same task (i.e. process) at run time.  Therefore, if the user specifies `output_files=32` and `out_files_per_proc=8`, the `to_parquet` algorithm will generate four separate subgraphs, allowing the input partitions for each subgraph to be shuffled between 8 different files.


benfred:
rerun tests

nvtabular/io/dataset.py
```diff
@@ -688,24 +698,40 @@ def to_parquet(
...
+            # Use out_files_per_proc to calculate how
+            # many output files should be written within the
+            # same subgraph.  Note that we must a
+            files_per_task = out_files_per_proc or 1
             required_npartitions = ddf.npartitions
             if isinstance(output_files, int):
                 required_npartitions = output_files
-            elif isinstance(output_files, list):
+                files_per_task = min(files_per_task, output_files)
+            elif isinstance(output_files, int):
```
benfred, line 709: Is this right? Looks like this  should still be checking for a list (like we're checking isinstance int on line 706)
```suggestion
            elif isinstance(output_files, list):
```

benfred (commented):
Looks mostly good!

nvtabular/io/dataset.py
rjzamora: Oops - good catch :)

benfred (approved):

PR #957, opened 2021-07-15
example: fix "device" kwarg in dataloader_bench.py

chongxiaoc:
Also update comments for torch loader.


chongxiaoc:
After 2nd check, it is not a correct fix, `device` in `Dataloader` device  idx instead of number of all GPUs.

chongxiaoc:
@karlhigley PTAL

karlhigley (approved):

benfred (approved):

PR #963, opened 2021-07-16
Categorical label reorder and export

jperez999:
This PR tackles two issues:
Export count to file
Order categorical labels in decreasing frequency order


nvtabular/ops/categorify.py
```diff
@@ -30,6 +30,7 @@
 from dask.delayed import Delayed
 from dask.highlevelgraph import HighLevelGraph
 from fsspec.core import get_fs_token_paths
+from nvtx import annotate
```
benfred, line 33: We shouldn't be importing nvtx.annotate directly anymore - we should be using the version inside nvtabular.dispatch (contains a fallback for case where nvtx isn't installed, like CPU only systems)

benfred (commented):
I'm not convinced the 'gb.sort_values' in mid_level_groupby is sufficient here - like I believe we're calling again on the column nam in ```_write_uniques``` that well undo this. Can you add a unittest to make sure the output is still sorted by frequency?

Also we should probably only do the sorting by frequency when the ```search_sorted``` parameter isn't set : otherwise we won't be returning correct results in this case (or alternatively, we could remove the search_sorted option).

nvtabular/ops/categorify.py
jperez999: yeah this is a draft still.

nvtabular/ops/categorify.py
```diff
@@ -824,13 +824,22 @@ def _write_uniques(dfs, base_path, col_group, options):
                 if nlargest < len(df):
                     df = df.nlargest(n=nlargest, columns=name_count)
             if not _series_has_nulls(df[col]):
+                if name_count in df:
+                    df = df.sort_values(name_count, ascending=False, ignore_index=True)
```
benfred, line 828: I think we might be sorting the values a couple times here unnecessarily: Is the sort_values on line 804 still necessary with this?  Also is this sort_values still necessary if we have a max_size parameter passed?

jperez999: Absolutely! The sort on 804 is done by the column itself, not the count. This sort brings the NA values first... it allows us to guarantee handling logic when a series does have nulls (we know its at the front and its only the first value). The second sort is by count. But we still need to maintain unknown as index 0.

benfred: ok. For just ordering the NA value first we could probably do something a little cheaper than a full sort (like O(n) instead of O(n * log(N)) - but this works for now .

benfred (approved):

PR #971, opened 2021-07-19
Use output_files by default in to_parquet

rjzamora:
This is a follow-up to #956, with the intention of improving default `to_parquet` memory stability.

This PR changes the default algorithm used by `to_parquet` from a "worker"-centered methodology to a "subgraph"-centered methodology.


nvtabular/utils.py
```diff
@@ -156,3 +157,15 @@ def _ensure_optimize_dataframe_graph(ddf=None, dsk=None, keys=None):
...
+def set_dask_client(client):
+    # Check for distributed client.
+    # Use `get_client` if a  client is not
+    # explicitly specified (and `client!=False`)
+    if client is None:
+        try:
+            return get_client()
+        except ValueError:
+            return None
+    return client
```
rjzamora, line 171: I am making the assumption here that we should start allowing NVTabular to automatically detect the presence of a `client` (without requiring the user to specify it explicitly).  If the user has deployed a cluster and defined a client, it is up to them to specify `client=False` if they want single-threaded execution.

benfred: Does this assumption play well with the unittests? We have a dask client fixture that will exist for the lifetime of the module: https://github.com/NVIDIA/NVTabular/blob/460c287428dda507a7a23f50c2a6184281c38cd7/tests/conftest.py#L87-L93

And it seems like our tests will use it now even if use_client is false: https://github.com/NVIDIA/NVTabular/blob/460c287428dda507a7a23f50c2a6184281c38cd7/tests/unit/test_workflow.py#L47-L49

edit: I see you've updated one of the tests (test_io) - but think we also need to update test_workflow and test_dask_nvt

rjzamora: I am moving the client changes to a distinct PR  - for several reasons, including this :)

rjzamora:
rerun tests

benfred:
Looks like there are some failures to build the docs here:

> /var/jenkins_home/workspace/nvtabular_tests/nvtabular/docs/./../nvtabular/io/dataset.py:docstring of nvtabular.io.dataset.Dataset.to_parquet:39:Unexpected indentation.

rjzamora:
@benfred - Okay. Hopefully this PR is a bit less controversial now :)

benfred (approved):

PR #972, opened 2021-07-19
Fix triton concatenating dataframes

benfred:
There was a bug in the triton model.py file where dataframes weren't getting
concatenated appropriately.  This didn't happen with the dict representation
causing this to get missed during manual testing.

Fix and add a unittest that would catch this.


albert17 (approved):

benfred:
rerun tests

PR #974, opened 2021-07-20
Automatically detect Dask client in Workflow and Dataset

rjzamora:
NVTabular currently requires a distributed Client object to be specified explicitly when a new `Workflow` and/or `Dataset` object are constructed.  This does not really align with conventional Dask-collection behavior, where the client will be automatically detected when compute/persist is called.  This PR adds the necessary logic to autodetect a Dask cleint within the `Workflow` and `Dataset` constructors.

Some open questions:

- What do we want the `client` argument options and defaults to be? This PR currently allows the user to specify an explicit client, or a `bool` to reflect whether or not a client should be automatically detected.
- Should this auto-detection logic even live in the constructor, or should we be detecting the client when compute is internally called within NVTabular (allowing users to spin up a cluster after the workflow/Dataset are defined)?

### DIFF
--- a/nvtabular/io/dataset.py
+++ b/nvtabular/io/dataset.py
@@ -31,6 +31,7 @@ from fsspec.utils import stringify_path
 
 from nvtabular.dispatch import _convert_data, _hex_to_int, _is_dataframe_object
 from nvtabular.io.shuffle import _check_shuffle_arg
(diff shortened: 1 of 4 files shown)


rjzamora:
Closing in favor of #985 - We can revisit this idea in a fresh PR for the next release cycle.

PR #978, opened 2021-07-21
HugeCTR notebook updates

bschifferer:
Update Notebooks with latest HugeCTR Python API


examples/getting-started-movielens/inference-HugeCTR/Triton-Inference-with-HugeCTR.ipynb
```diff
@@ -51,6 +51,48 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Overwriting /model/models/ps.json\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%writefile '/model/models/ps.json'\n",
+    "# fmt: off\n",
```
benfred, line 82: Do we need this in there? We've removed black-nb from our pre-commit hooks because of issues like this, and I don't think that black-nb should be blocking anything (like the only reference to black-nb I see is in our requirements-dev.txt https://github.com/NVIDIA/NVTabular/search?q=black-nb and I think we can probably remove that).

Also: does this `# fmt: off` flag get written to json file? It seems to work with the hugectr json parser, but isn't strictly valid json

examples/scaling-criteo/04-Triton-Inference-with-HugeCTR.ipynb
```diff
@@ -106,233 +85,13 @@
...
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "Overwriting ./criteo_hugectr/1/criteo.json\n"
-     ]
-    }
-   ],
-   "source": [
-    "%%writefile './criteo_hugectr/1/criteo.json'\n",
```
benfred, line 152: I love that we can get rid of these json files now =)

examples/scaling-criteo/04-Triton-Inference-with-HugeCTR.ipynb
```diff
@@ -430,65 +189,110 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Overwriting /model/ps.json\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%writefile '/model/ps.json'\n",
+    "# fmt: off\n",
```
benfred, line 284: same question as w/ movielens

benfred (approved):
some minor questions - but looks good! 

I tested the criteo notebooks (both training and inference) successfully on my system fwiw,

PR #982, opened 2021-07-22
fix getting TF_VISIBLE_DEVICE from the environment

rongou:
`tf.config.list_physical_devices` returns a list which can only be indexed by an integer.


benfred:
rerun tests

benfred (approved):
thanks!

PR #990, opened 2021-07-26
Column tagging API 

marcromeyn:


nvtabular/column_group.py
```diff
@@ -44,28 +54,41 @@ def __init__(self, columns):
         if isinstance(columns, str):
             columns = [columns]
 
+        if not tags:
+            tags = []
+        if isinstance(tags, DefaultTags):
+            tags = tags.value
```
karlhigley, line 60: Think it would be worth creating a `Taggable` mix-in to handle shared code between `Column` and `ColumnGroup`?

nvtabular/column_group.py
```diff
@@ -179,27 +207,160 @@ def __getitem__(self, columns):
...
+            filtered = [c for c in self.columns if filter_fn(c.name)]
+        else:
+            filtered = [c for c in self.columns if filter_fn(c)]
+
+        return self[filtered]
+
+    def filter_by_namespace(self, namespace):
+        return self.filter_columns(lambda c: c.startswith(namespace))
+
+    def get_tagged(self, tags, output_list=False, tags_to_filter=None):
```
karlhigley, line 231: When would you want this method to return a `list` vs a `ColumnGroup`? What would be the impact of standardizing the output type to one or the other?

nvtabular/column_group.py
```diff
@@ -179,27 +207,160 @@ def __getitem__(self, columns):
...
+    def tags_by_column(self):
+        outputs = {}
+
+        for col in self.flattened_columns:
+            outputs[col.name] = col.tags
+
+        return outputs
+
+    def targets_columns(self):
+        return self.get_tagged(Tag.TARGETS, output_list=True)
```
karlhigley, line 267: If we can sort out a way to have `get_tagged` return a consistent type, could we live without these additional public methods? It seems like they're mostly syntactic sugar to avoid needing to pass the flag, and it would be hard to add methods like this for custom tags.

nvtabular/ops/base.py
```diff
@@ -45,7 +45,7 @@ class Operator:
     Base class for all operator classes.
```
karlhigley, line 45: Since the file has been renamed to `base.py`, I wonder if we should also rename this class to `BaseOperator` to match. 🤔

nvtabular/ops/base.py
```diff
@@ -62,13 +62,12 @@ def transform(self, columns: ColumnNames, df: DataFrameType) -> DataFrameType:
...
-    def output_column_names(self, columns: ColumnNames) -> ColumnNames:
-        """Given a set of columns names returns the names of the transformed columns this
-        operator will produce
+    def output_columns(self, columns: Columns) -> Columns:
+        """Given a set of columns names returns transformed columns this operator will produce
 
         Parameters
         -----------
-        columns: list of str, or list of list of str
+        columns: Columns
```
karlhigley, line 70: This seems like it would be clearer what's happening (to me anyway) if it was `List[Column]`

nvtabular/ops/base.py
```diff
@@ -103,7 +102,43 @@ def supports(self) -> Supports:
...
         return Supports.CPU_DATAFRAME | Supports.GPU_DATAFRAME
 
-    def inference_initialize(self, columns: ColumnNames, model_config: dict) -> Optional[Operator]:
+    def inference_initialize(self, columns: Columns, model_config: dict) -> Optional[Operator]:
         """Configures this operator for use in inference. May return a different operator to use
         instead of the one configured for use during training"""
         return None
+
+
+class AddMetadata(Operator):
```
karlhigley, line 111: Where/how does this operator get used?

nvtabular/ops/categorify.py
```diff
@@ -438,11 +443,20 @@ def transform(self, columns: ColumnNames, df: DataFrameType) -> DataFrameType:
 
         return new_df
 
-    def output_column_names(self, columns: ColumnNames) -> ColumnNames:
+    # def output_column_names(self, columns: ColumnNames) -> ColumnNames:
+    #     if self.encode_type == "combo":
+    #         cat_names, _ = _get_multicolumn_names(columns, columns, self.name_sep)
+    #         return cat_names
+    #     return list(flatten(columns, container=tuple))
```
karlhigley, line 450:
```suggestion
```

nvtabular/tag.py
```diff
@@ -0,0 +1,58 @@
...
+
+class DefaultTags(Enum):
+    # Feature types
+    CATEGORICAL = ["categorical"]
+    CONTINUOUS = ["continuous"]
+    LIST = ["list"]
+    IMAGE = ["image"]
+    TEXT = ["text"]
+    TEXT_TOKENIZED = ["text_tokenized"]
+    TIME = ["time"]
```
karlhigley, line 12: For the moment, I think we only support `CATEGORICAL`, `CONTINUOUS`, and `LIST`, so we might want to leave these other tags for later?

requirements.txt
```diff
@@ -1,4 +1,5 @@
 PyYAML>=5.3
+dm-tree
```
karlhigley, line 2: As noted above, wondering if we can live without introducing a dependency into the core classes of this library. It looks really handy, but that couples our core abstractions to external code.

tests/unit/test_ops.py
```diff
@@ -1152,32 +1153,32 @@ def test_list_slice(cpu):
     df = DataFrame({"y": [[0, 1, 2, 2, 767], [1, 2, 2, 3], [1, 223, 4]]})
 
     op = ops.ListSlice(0, 2)
-    transformed = op.transform(["y"], df)
+    transformed = op.transform(Columns.from_names("y"), df)
```
karlhigley, line 1156: If we're going to have a `Columns` class, maybe `from_names` could be the `__call__` method? Seems like that would make this part a bit more ergonomic.

karlhigley (commented):
We probably have some philosophical discussions ahead of us to sort out how we want to approach the API design, but looks pretty good on the whole. 👍🏻

nvtabular/column.py
```diff
@@ -0,0 +1,75 @@
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional, Text
+
+import tree
```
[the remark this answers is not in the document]
marcromeyn: You're totally right. I mentioned this to @benfred yesterday, at this point I am not even sure whether this `Columns` class makes sense. I just created it to replace `ColumnNames` and add some utility methods to make it easy to use and easy to change the `fit` & `transform` method of each op. I wanted to cut some corners to have the option of merging a PR this week and then iterate on some parts of the design later on, this because quite a few things like the schema stuff would depend on some sort of column tagging. I would also be fine with the approach of hashing out the full design of this first, but at the very least having cut these corners allows us to make a decision which approach to take.

You could argue that we already have a `ColumnGroup` class that conceptually is kinda the same as `Columns`, so it might be better to merge the 2 classes. `ColumnGroup` is an interesting class since it also represents a node in the transform-DAG. So an option could be to break out parts of that class in a base-class and potentially put that in a merlin-utils library that can be used in various places. That would make it much easier to use libraries like Transformers4Rec without NVTabular.

.pylintrc
```diff
@@ -58,7 +58,8 @@ disable=fixme,
     inconsistent-return-statements,
     duplicate-string-formatting-argument,
     len-as-condition,
-    cyclic-import
+    cyclic-import,
+    too-many-public-methods
```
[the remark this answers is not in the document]
marcromeyn: That would work for sure! I would like keep the Enum as well but we could make retrieving it by string the default since it's the most concise way.

nvtabular/column.py
```diff
@@ -0,0 +1,75 @@
...
+            return self
+
+        tags = list(set(list(self.tags) + list(tags))) if add else tags
+
+        return Column(self.name, tags=tags, properties=self.properties)
+
+    def with_name(self, name) -> "Column":
+        return Column(name, tags=self.tags, properties=self.properties)
+
+    def with_properties(self, add=True, **properties) -> "Column":
```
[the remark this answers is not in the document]
marcromeyn: I feel like for production use-cases the properties could be super useful. We could think of some potential use-cases for it for the library itself as well.

nvtabular/column.py
```diff
@@ -0,0 +1,75 @@
...
+        if not properties:
+            return self
+        properties = {**self.properties, **properties} if add else properties
+
+        return Column(self.name, tags=self.tags, properties=properties)
+
+
+class Columns(list):
```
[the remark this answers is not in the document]
marcromeyn: Yes, for sure! See the bottom part of my comment here: https://github.com/NVIDIA/NVTabular/pull/990#discussion_r677201059/

nvtabular/ops/base.py
marcromeyn: I haven't ported this yet, but an example can be found [here](https://github.com/NVIDIA/NVTabular/blob/romeyn/input-api/nvtabular/dataset/movielens.py#L35). This allows for manual tagging of columns. I should add some tests showcasing this as well.

tests/unit/test_ops.py
marcromeyn: Oh, I like that! Lemme add that

nvtabular/column.py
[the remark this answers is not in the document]
karlhigley: Yeah, that's a good point re: `ColumnGroup` filling multiple conceptual roles as both a collection of columns and a workflow graph node. 🤔

.pylintrc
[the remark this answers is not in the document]
karlhigley: Yeah, I think the Enum should definitely stay. 👍🏻

nvtabular/column.py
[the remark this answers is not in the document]
karlhigley: Yeah, I'm not opposed to adding properties in general, just wondering if there's anything currently relying on it. If not, it might be worth postponing to a future PR to keep the scope for this release as small as we can.

nvtabular/ops/base.py
karlhigley: Ahh, if this is used in the Datasets API, it might fit better in that PR.

nvtabular/ops/base.py
```diff
@@ -100,10 +99,46 @@ def label(self) -> str:
...
```
jperez999, line 111: This should not be in the base class file... Its a stand alone operator. should be in metadata.py in ops or something.

nvtabular/ops/base.py
```diff
@@ -100,10 +99,46 @@ def label(self) -> str:
...
+        if is_binary_target:
+            tags.extend(DefaultTags.TARGETS_BINARY.value)
+
+        self.tags = list(set(tags))
+
+    def transform(self, columns: Columns, df: DataFrameType) -> DataFrameType:
+        return df
+
+    def output_columns(self, columns: Columns) -> Columns:
+        return columns.map(lambda col: col.with_tags(self.tags).with_properties(**self.properties))
```
jperez999, line 144: If we were to merge those _with calls... we would only need to return the columnGroup once... If I am reading this correctly we will first create a columngroup with tags then another columngroup with those tags and the properties...

nvtabular/ops/categorify.py
```diff
@@ -360,13 +363,13 @@ def process_vocabs(self, vocabs):
...
-            self.tree_width,
-            self.on_host,
+            col_groups=columns,
+            agg_cols=[],
+            agg_list=[],
+            out_path=self.out_path,
+            freq_limit=self.freq_threshold,
+            tree_width=self.tree_width,
+            on_host=self.on_host,
             concat_groups=self.encode_type == "joint",
```
jperez999, line 373: Whats the reasoning behind the kwargs shift?

nvtabular/ops/categorify.py
```diff
@@ -378,10 +381,10 @@ def set_storage_path(self, new_path, copy=False):
         self.out_path = new_path
 
     @annotate("Categorify_transform", color="darkgreen", domain="nvt_python")
-    def transform(self, columns: ColumnNames, df: DataFrameType) -> DataFrameType:
+    def transform(self, columns: Columns, df: DataFrameType) -> DataFrameType:
         new_df = df.copy(deep=False)
         if isinstance(self.freq_threshold, dict):
-            assert all(x in self.freq_threshold for x in columns)
+            assert all(x in self.freq_threshold for x in columns.names())
```
jperez999, line 387: Do we need to flatten this here? What happens if you have a tuple in this... will it check inside the tuple? is it supposed to?

nvtabular/column.py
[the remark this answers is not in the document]
benfred: Is one use case the dimensionality of embeddings? (like the tags would be 'categorical' but the properties would be 'max=embedding_dimensionality?' I could see this being useful for the hugectr flat embedding table if so

.pylintrc
```diff
@@ -58,7 +58,8 @@ disable=fixme,
     inconsistent-return-statements,
     duplicate-string-formatting-argument,
     len-as-condition,
-    cyclic-import
+    cyclic-import,
```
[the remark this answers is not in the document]
marcromeyn: I agree but this was already in there, the change I made is to add this line: https://github.com/NVIDIA/NVTabular/blob/b2ffc729956fceb147954de5207c5a85e66d7d15/.pylintrc#L62

nvtabular/column.py
```diff
@@ -0,0 +1,75 @@
...
+    @classmethod
+    def from_names(cls, *names):
+        return Columns([Column(name) for name in names])
+
+    def map(self, fn) -> "Columns":
```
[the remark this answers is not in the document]
marcromeyn: See my comment here on why I added it for now: https://github.com/NVIDIA/NVTabular/pull/990#discussion_r677201059. I do think there is value in these utility functions, we could just reimplement them ourselves without an external dependency.

nvtabular/ops/categorify.py
marcromeyn: I was wondering the same! There are some other parts of the codebase where this might be an issue as well.

nvtabular/ops/categorify.py
marcromeyn: It's a bit more robust against changes of the class itself. In these case the values we use for these arguments are pretty explanatory but there some other parts of the codebase where that's not the case. So to keep things consistent I changed it here as well.

nvtabular/column.py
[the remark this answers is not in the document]
marcromeyn: I agree! We could also reason from the other way, since we want to allow to create a `ColumnGroup` from a schema we should make sure that we cover all fields there. As you can see here they allow to set the `annotation` field that next to tags also includes a comment and extra_metadata ([source](https://github.com/tensorflow/metadata/blob/master/tensorflow_metadata/proto/v0/schema.proto#L214-L224)).

nvtabular/ops/base.py
marcromeyn: I would argue that tagging columns manually is a core part of the API proposed in this PR. I do agree with @jperez999 that we should break this out into a separate file.

nvtabular/tag.py
marcromeyn: We have some code in Transformers4Rec that uses `TEXT` & `TEXT_TOKENIZED`. I don't see an issue of having some often used tags here that don't have direct usage in our library code from the start.

nvtabular/column_group.py
karlhigley: See 248-249

nvtabular/column.py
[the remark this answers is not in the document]
karlhigley: I'm cool with all of that, but will still probably opt to postpone adding properties until we're actually implementing a feature that requires them.

nvtabular/column.py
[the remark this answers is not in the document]
karlhigley: You know, the more I think about it, the less I think that's the right path forward. `Columns` here is actually an expansion of `ColumnNames` that shifts from representing columns by their string names to representing them as an object with metadata (including name.) `Columns` and `ColumnGroup` aren't the same thing for the same reason that `ColumnNames` and `ColumnGroup` weren't the same thing.

nvtabular/column.py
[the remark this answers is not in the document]
karlhigley: If I understand correctly, these utility functions (at least where `tree` is involved) are handling iteration through nested lists of column names for operations that involve feature crosses (e.g. column similarity.) I sense a few code smells here though, which are:
* Sub-classing a built-in type (`list`)
* Creating an aggregate psuedo-collection type with a plural name (`Columns`)
* Integrating an external dependency into a core abstraction (`tree` in our column abstractions)

Having played around with the code a bit and attempted some (ill-fated) refactors, I think I see how/why the code ended up in this shape though. It seems like the original type alias `ColumnNames = List[Union[str, List[str]]]` and the circular dependency between `Operator` and `ColumnGroup` are what force the code this direction.

nvtabular/ops/base.py
karlhigley: This point is an outgrowth of much earlier decisions in the code base (e.g. `ColumnNames`) so it may or may not be worth addressing here, but I'm still wondering if we can massage the code in a direction where plain un-nested lists would suffice. In order to do that, we'd have to figure out a way to make feature crosses across two lists of columns work with the `Operator` API somehow.

nvtabular/ops/base.py
karlhigley: Hmm, okay. It's a little hard to tell from the content of the PR itself without examples/tests, but I can get on board with manual tagging as important part of this API. Agreed with y'all that we should break it out then.

nvtabular/ops/base.py
karlhigley: Yeah, I think you're right. Since we're not using the properties yet, I'd be happy just to remove `with_properties` for now and kick the can down the road a little bit.

nvtabular/ops/base.py
karlhigley: I think it's just a style thing? I don't have a strong preference either way.

nvtabular/column.py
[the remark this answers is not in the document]
karlhigley: I think the version of `ColumnGroup` in Transformers4Rec is roughly the core of a column-grouping class that isn't also a graph node. I'm not entirely sure how to sort out `Column` vs `Columns` vs `ColumnGroup` vs `ColumnGroupNode`, but I'll keep pondering it.

benfred (commented):
I like the ideas in this PR - and think this will be super useful for a bunch of features.

I think the implementation could be improved though. I have some high level thoughts here : 
* I think we can merge the Columns class and the ColumnGroup class
* Instead of adding methods like ‘categorical_columns’ to the column_group, I think we can just use a generic select function.  We're going to allow user defined tags , so we can't have an exhaustive list of functions anyways - and it won't be all that much  more verbose
* The ‘tags’  representation here could be improved:
   * DefaultTags is multi-valued and probably shouldn’t be. The ’tags’ object on the ‘Column’ class is a collection - we shouldn’t have values of this collection also being a collection. 
   * Can we merge the Tag and DefaultTags class? Using the DefaultTags enum doesn’t buy us anything since we’re just duplicating the values in the Tag class - and we allow user defined tags here. 
   * The ‘tags’ member on the Column class should probably be a set rather than a list: order doesn’t matter .


Nitpicky details:

 * We need docstrings for all new major functionality
  * Using a dataclass for the Column class, which  also has multiple methods seems wrong to me.  I’d rather just use a regular class here
  * the AddMetadata class should probably just take tags/parameters arguments instead of boolean flags indicating what to add. This will be more flexible and about the same level of verbosity
  * There are a couple of changes here that are unrelated to the core of the PR:
	* Using named parameters in FitOptions
	* Moving operator base classes from operator.py to base.py

karlhigley:
>     * I think we can merge the Columns class and the ColumnGroup class

I thought this too at first, but `Columns` (soon to be `ColumnSchemas`) is an expanded version of `ColumnNames`, not a conceptual parallel to `ColumnGroup` (which also has the responsibility of being a graph node in addition to collecting information about a set of columns.) I don't think we want to be passing around graph nodes into `Operators` to identify sets of columns, which would create the very weird situation where a `ColumnGroup` (node/set of columns) has an `Operator` that takes a `ColumnGroup` (set of columns/node.)

Recall also that you mentioned that you probably used `ColumnNames` instead of `ColumnGroup` in the `Operator` public methods due to the circular dependency between `Operator` and `ColumnGroup`. If we merged `Columns` and `ColumnGroup`, we'd end up exacerbating the circular dependency.

As much as I'd like to avoid having to make deeper changes to the library, this particular aspect of the code is deceptively difficult to wrangle. I'm hoping that the combination of renaming `Column` to `ColumnSchema` and `Columns` to `ColumnSchemas` will make it clearer what's going on and help us think through this. Pretty sure we're going to need to introduce a new concept somewhere in here to make it work though (possibly by splitting or clarifying the `ColumnGroup` class.)

>     * Instead of adding methods like ‘categorical_columns’ to the column_group, I think we can just use a generic select function.  We're going to allow user defined tags , so we can't have an exhaustive list of functions anyways - and it won't be all that much  more verbose

See also my [PR on Transformers4Rec](https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/77) which slims down the `ColumnGroup` API and addresses the query method conundrum.

nvtabular/column_group.py
```diff
@@ -179,27 +207,160 @@ def __getitem__(self, columns):
...
+    def text_tokenized_column_group(self):
+        return self.get_tagged(Tag.TEXT_TOKENIZED, output_list=False)
+
+    def embedding_sizes(self):
+        return self._embedding_sizes_from_op()
+
+    def cardinalities(self):
+        return {k: v[0] for k, v in self._embedding_sizes_from_op().items()}
+
+    def _embedding_sizes_from_op(self):
```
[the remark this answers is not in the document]
karlhigley: Oh, good catch! I'm not sure where these should go, but this does seem like kind of a weird place for them.

karlhigley:
(I made a few schema-related renames and removed some extraneous changes in the preceding commits.)

PR #991, opened 2021-07-26
Data-schema & statistics

marcromeyn:
Depends on #990

### DIFF
--- a/nvtabular/column_group.py
+++ b/nvtabular/column_group.py
@@ -20,7 +20,7 @@ from typing import Dict, List, Optional, Text, Union
 import joblib
 from dask.core import flatten
 
-from nvtabular.column import Column
+from nvtabular.column import Column, Columns
 from nvtabular.ops import LambdaOp, Operator
 from nvtabular.tag import DefaultTags, Tag
 
@@ -65,7 +65,7 @@ class ColumnGroup:
         # if any of the values we're passed are a columngroup
         # we have to ourselves as a childnode in the graph.
         if any(isinstance(col, ColumnGroup) for col in columns):
-            self.columns = []
+            self.columns: Columns = Columns()
             self.kind = "[...]"
             for col in columns:
                 if not isinstance(col, ColumnGroup):
@@ -81,11 +81,12 @@ class ColumnGroup:
                 self.columns.append(tuple(col.columns))
 
         else:
-            self.columns = [_convert_col(col, tags=tags, properties=properties) for col in columns]
+            columns = [_convert_col(col, tags=tags, properties=properties) for col in columns]
+            self.columns: Columns = Columns(columns)
 
     @property
     def column_names(self):
-        return [str(c) for c in self.columns]
+        return self.columns.names()
 
     def __call__(self, operator, **kwargs):
         """Transforms this ColumnGroup by applying an Operator
@@ -107,10 +108,7 @@ class ColumnGroup:
 
         if not isinstance(operator, Operator):
             raise ValueError(f"Expected operator or callable, got {operator.__class__}")
-        if getattr(operator, "output_columns", None):
-            child_columns = operator.output_columns(self.columns)
-        else:
-            child_columns = operator.output_column_names(self.column_names)
+        child_columns = operator.output_columns(self.columns)
 
         child = ColumnGroup(child_columns)
         child.parents = [self]
@@ -152,7 +150,8 @@ class ColumnGroup:
             other = ColumnGroup(other)
 
         # check if there are any columns with the same name in both column groups
-        overlap = set(self.column_names).intersection(other.column_names)
+        overlap = self.column_names.intersection(other.column_names)
+
         if overlap:
             raise ValueError(f"duplicate column names found: {overlap}")
 
@@ -252,7 +251,7 @@ class ColumnGroup:
         child = ColumnGroup(columns, tags=tags)
         child.parents = [self]
         self.children.append(child)
-        child.kind = f"tagged={child._tags_repr} " + self._cols_repr
+        # child.kind = f"tagged={child._tags_repr} " + self._cols_repr
 
         return child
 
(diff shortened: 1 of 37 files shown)


nvtabular/ops/data_schema.py
```diff
@@ -0,0 +1,173 @@
...
+        super().__init__()
+        self.schema_path = os.path.join(output_path, self.SCHEMA_FILE_NAME) if output_path else None
+        self.col_names = []
+        self.col_types = []
+        self.col_dtypes = []
+        self.tags_by_column = tags_by_column
+        self.schema = None
+
+    @classmethod
+    def calculate_on_dataset(cls, dataset, tags_by_column, output_path=None, client=None):
```
karlhigley, line 44: Since there's a convenience method that calls this on the `Workflow` class anyway, wonder if this method could be in-lined over there to avoid the additional imports here.

requirements.txt
```diff
@@ -5,4 +5,5 @@ distributed==2021.4.1
 pandas>=1.0,<1.3.0dev0
 numba>=0.53.1
 tdqm
+tensorflow-metadata
```
karlhigley, line 8: This is another place where I wonder if it's worth adding the dependency, or if we should implement something similar ourselves.

nvtabular/ops/data_schema.py
```diff
@@ -0,0 +1,173 @@
...
+            col_group += ColumnGroup(column, tags=tags_by_column.get(column))
+
+        workflow = Workflow(col_group >> stats, client=client)
+        workflow.fit(dataset)
+
+        return stats.schema
+
+    @classmethod
+    def load(cls, directory):
+        from tensorflow_metadata.proto.v0 import schema_pb2
```
karlhigley, line 62: Does this schema work across frameworks (TF/Torch)?

nvtabular/ops/data_schema.py
```diff
@@ -0,0 +1,173 @@
...
+            return None
+
+        schema = schema_pb2.Schema()
+        with open(schema_file, "rb") as f:
+            schema.ParseFromString(f.read())
+
+        return schema
+
+    def transform(self, columns: Columns, df: DataFrameType) -> DataFrameType:
+        return df
```
karlhigley, line 78: This method seems extraneous unless maybe the super-class requires it to implemented?

nvtabular/ops/data_schema.py
```diff
@@ -0,0 +1,173 @@
...
+                dask_stats[col]["min"] = domain.min()
+                dask_stats[col]["max"] = domain.max()
+
+        return dask_stats
+
+    def fit_finalize(self, stats):
+        return self.prepare_schema(stats)
+
+    def prepare_schema(self, stats):
+        from tensorflow_metadata.proto.v0 import schema_pb2
```
karlhigley, line 113: If we are going to depend on this library across frameworks, it seems like we can just import it at the top of the file.

nvtabular/ops/data_stats.py
```diff
@@ -13,95 +13,290 @@
...
+from .data_schema import DataSchema
 from .stat_operator import StatOperator
 
+LOG = logging.getLogger("nvtabular")
 
-class DataStats(StatOperator):
-    def __init__(self):
-        super().__init__()
+
+class DataStats(DataSchema):
```
karlhigley, line 34: This seems like an unusual inheritance chain, at least from the class names. I normally wouldn't expect stats to be a schema.

nvtabular/ops/data_stats.py
```diff
@@ -13,95 +13,290 @@
...
+        self.tags_by_column = tags_by_column
+
+        return self.prepare_schema(self.stats)
 
     transform.__doc__ = Operator.transform.__doc__
     fit.__doc__ = StatOperator.fit.__doc__
     fit_finalize.__doc__ = StatOperator.fit_finalize.__doc__
+
+
+class DatasetCollectionStatistics:
```
karlhigley, line 263: Since this for rendering, I wonder if we should split it out into a separate sub-package (e.g. `notebooks` or `rendering_utils`)

nvtabular/workflow.py
```diff
@@ -80,6 +82,11 @@ def __init__(self, column_group: ColumnGroup, client: Optional["distributed.Clie
         self.input_dtypes = None
         self.output_dtypes = None
 
+    def generate_schema(self, dataset, output_path=None):
```
karlhigley, line 85: If this is where we want to use the method, could we inline it here?

nvtabular/ops/data_schema.py
marcromeyn: Yeah, good point! I am not completely happy with the design of it as of now. Because going through the Workflow feels like a little bit of an overkill that also requires some boilerplate. Since what you typically want for both `DataSchema` & `DataStats` is to apply it on a `Dataset` where you use all columns in a `ColumnGroup`. Thinking out loud here, one other way we could do that is to overload the `>>` operator on the `Dataset` for that?

requirements.txt
marcromeyn: So for consistency I think it would be best to make use of the protobuf schemas defined in `tensorflow-metadata`, this would also open up the possibility to offer deep integration with TFX in a separate library. We could, however, choose to copy over the relevant protobuf schemas to our repo. The reason I didn't do that now is because I am not sure if that's possible with the license they use. The `tensorflow-metadata` dependency is just a couple of protobuf definitions so it is super lightweight, so I think it could be fine.

nvtabular/ops/data_schema.py
marcromeyn: Yeah, for sure! The schema definition is super generic (source can be found [here](https://github.com/tensorflow/metadata/blob/master/tensorflow_metadata/proto/v0/schema.proto). In Transformers4Rec we already are using this for PyTorch as well.

nvtabular/ops/data_schema.py
marcromeyn: Good catch!

nvtabular/ops/data_stats.py
```diff
@@ -13,95 +13,290 @@
...
-            # if string, replace string for their lengths for the rest of the computations
             if dtype == "object":
-                ddf[col] = ddf[col].map_partitions(lambda x: x.str.len(), meta=("x", int))
-            # Add list support when cudf supports it:
-            # https://github.com/rapidsai/cudf/issues/7157
-            # elif col_type == "cat_mh":
-            #    ddf[col] = ddf[col].map_partitions(lambda x: x.list.len())
+                dask_stats[col]["nunique"] = domain.nunique()
+                str_len = domain.map_partitions(lambda x: x.str.len(), meta=("x", int))
+                dask_stats[col]["avg_str_length"] = str_len.mean()
```
jperez999, line 115: Can you get the min and max string lengths also... this is helpful for transferring to the data generator.

nvtabular/ops/data_stats.py
```diff
@@ -13,95 +13,290 @@
...
+                dask_stats[col]["mean"] = domain.mean()
+                dask_stats[col]["std"] = domain.std()
 
-            # Get Percentage of NaNs for all
-            dask_stats[col]["per_nan"] = 100 * (1 - ddf[col].count() / len(ddf[col]))
+                dask_stats[col]["num_zeroes"] = (domain == 0).sum()
 
-        return dask_stats, _custom_moments(ddf[columns.names().flatten()])
+                h, bins = da.histogram(
+                    domain.to_dask_array(), 10, range=[domain.min(), domain.max()]
```
jperez999, line 128: how did we come to the select 10 bins? Is this best practice?

jperez999 (commented):
I am not a fan of adding the tensorflow package into our core operator code. But if its the best way to handle the protobuf transition?

nvtabular/column_group.py
```diff
@@ -84,6 +87,34 @@ def __init__(
             columns = [_convert_col(col, tags=tags, properties=properties) for col in columns]
             self.columns: Columns = Columns(columns)
 
+    @staticmethod
```
karlhigley, line 90: Reframing `ColumnGroup` as primarily a workflow graph node rather than a collection of columns makes me question whether the schema belongs here. It seems like this class already has more responsibilities than it should. I think we were talking yesterday about having `ColumnSchema` and `DatasetSchema` classes, so this might be the kind of thing we'd put in a `DatasetSchema` class.

PR #993, opened 2021-07-27
Fix Normalize.transform result when stdev is zero

rjzamora:
The `Normalize.transform` method is currently failing to return columns with a standard-deviation statistic equal to zero.  This is obviously a problem when a follow-on operator depends on that column.  This PR effectively sets all values in a "problematic column" to zero.


benfred (approved):

rjzamora:
rerun tests

rjzamora:
Not sure what might have cause [the first CI failure](https://github.com/NVIDIA/NVTabular/pull/993#issuecomment-887179814) - I cannot reproduce it locally, so I do not expect that it is related to this PR.

PR #994, opened 2021-07-27
Introducing DatasetCollection

marcromeyn:
This makes it easier to work with datasets with splits.

### DIFF
--- a/nvtabular/io/dataset_collection.py
+++ b/nvtabular/io/dataset_collection.py
@@ -1,0 +1,102 @@
+import os
+from typing import Any, Optional
+
+import pandas as pd
+
+from nvtabular.io.dataset import LOG, Dataset
+from nvtabular.utils import Namespace
+
+
+class DatasetCollection(Namespace):
+    def __init__(self, **kwargs: Any) -> None:
+        splits = {}
+        for key, val in kwargs.items():
+            if not val:
+                LOG.warning("%s is emtpy, remove it from the collection", key)
+            else:
+                splits[key] = val
+        assert all(isinstance(dataset, (Dataset, DatasetCollection)) for dataset in kwargs.values())
+        super().__init__(**kwargs)
+
+    @classmethod
+    def from_splits(cls, train, eval=None, test=None):  # pylint: disable=W0622
+        splits = dict(train=train)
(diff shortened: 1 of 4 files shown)


nvtabular/io/dataset_collection.py
```diff
@@ -0,0 +1,102 @@
+import os
+from typing import Any, Optional
+
+import pandas as pd
+
+from nvtabular.io.dataset import LOG, Dataset
+from nvtabular.utils import Namespace
+
+
+class DatasetCollection(Namespace):
```
karlhigley, line 10: `DatasetCollection` looks like it's specific to splits of a dataset, which makes the name seem more generic than the implementation. If splits are currently the only use case for this, maybe this class should be called something like `DatasetSplits`?

nvtabular/utils.py
```diff
@@ -168,3 +168,26 @@ def global_dask_client(client):
         except ValueError:
             return None
     return None
+
+
+class Namespace(SimpleNamespace):
```
karlhigley, line 173: If I understand correctly, the `SimpleNamespace` class provides syntactic sugar that allows you to create (for example) an object that acts like a namespace to organize member variables so you can do things like `object_from_custom_class.namespace.property` without having to use the dict indexing syntax like `object_from_custom_class.namespace["property"]`.

I think the advice from the Python docs is to use `namedtuple` if what you need is a structured record type though, which is what `DatasetCollection` appears to be. Is there a particular benefit of one over the other here? Are they essentially interchangeable?

nvtabular/workflow.py
```diff
@@ -195,6 +196,21 @@ def fit_transform(self, dataset: Dataset) -> Dataset:
         self.fit(dataset)
         return self.transform(dataset)
 
+    def fit_transform_collection(
```
karlhigley, line 199: What do you think about allowing a `DatasetCollection` to be passed into the normal `fit_transform` method? Is it the additional `to_fit` arg that makes that awkward?

karlhigley (commented):
Generally makes sense, but wonder if either there's a bit of premature abstraction here or there are more use cases for `DatasetCollection` than I realize.

PR #995, opened 2021-07-27
Creating operator block

marcromeyn:
This is the equivalent of a `torch.Module`/`tf.Layer` in a NVTabular workflow. This enables grouping various ops together in a single data-structure.

### DIFF
--- a/nvtabular/ops/__init__.py
+++ b/nvtabular/ops/__init__.py
@@ -16,6 +16,7 @@
 
 # alias submodules here to avoid breaking everything with moving to submodules
 # flake8: noqa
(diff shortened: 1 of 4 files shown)


karlhigley:
I see that `LogNormalize` uses `OperatorBlock` and some kind of sub-graph abstraction definitely makes sense to me. Aside from `LogNormalize` are there other current use cases for this? Wondering because the discussion about `ColumnGroup` functioning as both a group of columns and as a transform graph node makes me think we might want to tackle splitting `ColumnGroup` and creating a sub-graph/block abstraction together in order to end up with a cleaner set of graph abstractions.

PR #1001, opened 2021-07-28
putting import skip in correct locale

jperez999:
fixes import skip logic in torch layers test file.

### DIFF
--- a/tests/unit/test_torch_layers.py
+++ b/tests/unit/test_torch_layers.py
@@ -17,9 +17,9 @@ import pytest
 
 # If pytorch isn't installed skip these tests. Note that the
 # torch_dataloader import needs to happen after this line
-# needed for torch import for ConcatenatedEmbeddings
 torch = pytest.importorskip("torch")
 
+# Must come after `pytest.importorskip("torch")`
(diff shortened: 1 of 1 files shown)


tests/unit/test_torch_layers.py
```diff
@@ -15,12 +15,14 @@
...
 import pytest
 
-from nvtabular.framework_utils.torch.layers.embeddings import ConcatenatedEmbeddings
-
 # If pytorch isn't installed skip these tests. Note that the
 # torch_dataloader import needs to happen after this line
+# needed for torch import for ConcatenatedEmbeddings
 torch = pytest.importorskip("torch")
 
+from nvtabular.framework_utils.torch.layers.embeddings import ConcatenatedEmbeddings
```
karlhigley, line 23:
```suggestion
torch = pytest.importorskip("torch")

# Must come after `pytest.importorskip("torch")`
from nvtabular.framework_utils.torch.layers.embeddings import ConcatenatedEmbeddings
```

karlhigley (approved):
Looks okay to me. I'd suggest putting the comment about the ordering closer to the line it applies to though.

jperez999:
rerun tests

PR #1002, opened 2021-07-28
Fix the bugs in model.py for pytorch

oyilmaz-nvidia:
This PR fixes the bugs in model.py for pytorch.


tests/unit/test_triton_inference.py
```diff
@@ -211,15 +262,32 @@ def test_generate_triton_model(tmpdir, engine, df):
...
 
     # save workflow to triton / verify we see some expected output
+    if output_model == "pytorch":
+        model_info = {
+            "name-cat": {"columns": ["name-cat"], "dtype": "int64"},
+            "name-string": {"columns": ["name-string"], "dtype": "int64"},
+            "id": {"columns": ["id"], "dtype": "float32"},
+            "x": {"columns": ["x"], "dtype": "float32"},
+            "y": {"columns": ["y"], "dtype": "float32"},
+        }
```
benfred, line 272: I think this is broken on CPU only systems - maybe we should just skip the test if we're not running on a system ?

oyilmaz-nvidia: How can I detect if the system is CPU only? I am looking at the other tests and couldn't figure out how I can do that.

benfred: you can detect by using the ```nvtabular.dispatch.HAS_GPU``` flag. 

I pushed a commit that fixes the underlying problem though (bug in _is_list_dtype) - so we should be able to run this code on CPU only systems now.

benfred (approved):

PR #1015, opened 2021-08-03
ColumnSelector init setup and nested subgroup level checks

jperez999:
This PR adds new init and nested loop detection functionality with testing.

### DIFF
--- a/nvtabular/column_selector.py
+++ b/nvtabular/column_selector.py
@@ -92,6 +92,4 @@ class ColumnSelector:
     def __eq__(self, other):
         if not isinstance(other, ColumnSelector):
             return False
-        if other._names == self._names and other.subgroups == self.subgroups:
-            return True
-        return False
+        return other._names == self._names and other.subgroups == self.subgroups:
--- a/nvtabular/column_selector.py
+++ b/nvtabular/column_selector.py
@@ -20,7 +20,7 @@ import nvtabular
 
 class ColumnSelector:
     def __init__(self, names: List[str] = None, subgroups: List["ColumnSelector"] = None):
-        self._names = names if names is not None else [] 
+        self._names = names if names is not None else []
         self.subgroups = subgroups if subgroups else []
         if isinstance(self._names, nvtabular.ColumnGroup):
             raise TypeError("ColumnSelectors can not contain ColumnGroups")
--- a/nvtabular/column_selector.py
+++ b/nvtabular/column_selector.py
@@ -19,7 +19,7 @@ import nvtabular
 
 
 class ColumnSelector:
-    def __init__(self, names: List[str], subgroups=None):
+    def __init__(self, names: List[str], subgroups: List["ColumnSelector"] = None):
         self._names = names
         self.subgroups = subgroups if subgroups else []
         self.__post_init__()
--- a/nvtabular/column_selector.py
+++ b/nvtabular/column_selector.py
@@ -19,34 +19,9 @@ import nvtabular
 
 
 class ColumnSelector:
-    def __init__(self, names: List[str], subgroups: List["ColumnSelector"] = None):
-        self._names = names
+    def __init__(self, names: List[str] = None, subgroups: List["ColumnSelector"] = None):
+        self._names = names if names is not None else [] 
         self.subgroups = subgroups if subgroups else []
-        self.__post_init__()
-
-    @property
-    def names(self):
-        names = []
-        names += self._names
-        for subgroup in self.subgroups:
-            names += subgroup.names
-        return names
-
-    @property
-    def grouped_names(self):
-        names = []
-        names += self._names
-        for subgroup in self.subgroups:
-            names.append(tuple(subgroup.names))
-        return names
-
-    def __nested_check(self, nests=0):
-        if nests > 2:
-            raise AttributeError("Too many nested subgroups")
-        for col_sel0 in self.subgroups:
-            col_sel0.__nested_check(nests=nests + 1)
-
-    def __post_init__(self):
         if isinstance(self._names, nvtabular.ColumnGroup):
             raise TypeError("ColumnSelectors can not contain ColumnGroups")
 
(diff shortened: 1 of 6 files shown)


nvtabular/column_selector.py
```diff
@@ -13,16 +13,16 @@
...
 
 
-@dataclass
 class ColumnSelector:
-    _names: List[str]
-    subgroups: List["ColumnSelector"] = field(default_factory=list)
+    def __init__(self, names: List[str], subgroups: List["ColumnSelector"] = None):
+        self._names = names
+        self.subgroups = subgroups if subgroups else []
+        self.__post_init__()
```
karlhigley, line 25: I think we can inline the `__post_init__` method here in the constructor. `dataclass` calls that hook method, but it probably doesn't make sense as a separate method any more.

tests/unit/test_column_selector.py
```diff
@@ -37,6 +37,20 @@ def test_constructor_works_with_single_subgroups_and_lists():
     assert selector2.subgroups[0] == ColumnSelector(["a", "b", "c"])
 
 
+def test_constructor_too_many_level():
+
+    selector0 = ColumnSelector([], subgroups=ColumnSelector(["a", "b", "c"]))
+    selector1 = ColumnSelector(["f", "w", "g"], subgroups=selector0)
+    # next should break
+
+    test_pass = False
```
karlhigley, line 46: To test that a method raises an error, you can do `with pytest.raises(AttributeError) as exc_info:` as a context manager. The test will fail if the code inside the context block doesn't raise the error, and you can make assertions against the error message (e.g. to check the error message) after the block.

nvtabular/column_selector.py
```diff
@@ -84,3 +91,9 @@ def __radd__(self, other):
 
     def __rshift__(self, other):
         return nvtabular.ColumnGroup(self) >> other
+
+    def __eq__(self, other):
+        if isinstance(other, ColumnSelector):
```
karlhigley, line 96: The implementation looks right here. It might be clearer to first check if `other` is _not_ a `ColumnSelector` and immediately return `False` if so, because then the `if` statements wouldn't need to be nested and the return values would be close together with the conditions that return them.

tests/unit/test_column_selector.py
karlhigley: (Here's [an example](https://github.com/NVIDIA/NVTabular/pull/1014/files#diff-8f19358af314f63dd2835237d2bcbf90c9065980912f10c2f77ca98d990ba746R40-R51) from the `ColumnSelector` PR. I have to look up how to do this every time I want to do it.)

nvtabular/column_selector.py
```diff
@@ -40,6 +40,12 @@ def grouped_names(self):
             names.append(tuple(subgroup.names))
         return names
 
+    def __nested_check(self, nests=0):
```
karlhigley, line 43: I think the convention for a private method is just one underscore:
```suggestion
    def _nested_check(self, nests=0):
```

nvtabular/column_selector.py
```diff
@@ -13,16 +13,16 @@
...
 
 import nvtabular
 
 
-@dataclass
 class ColumnSelector:
-    _names: List[str]
-    subgroups: List["ColumnSelector"] = field(default_factory=list)
+    def __init__(self, names: List[str], subgroups: List["ColumnSelector"] = None):
+        self._names = names
```
karlhigley, line 23: It doesn't need to be addressed here, but I've been wondering if we should also make `names` an optional parameter, so that creating a `ColumnSelector` with only subgroups is easier. Right now, you have to do `ColumnSelector([],subgroups=[a_subgroup])`, which seems unnecessarily verbose.

nvtabular/column_selector.py
```diff
@@ -40,6 +40,12 @@ def grouped_names(self):
             names.append(tuple(subgroup.names))
         return names
 
+    def __nested_check(self, nests=0):
+        if nests > 2:
+            raise AttributeError("Too many nested subgroups")
+        for col_sel0 in self.subgroups:
+            col_sel0.__nested_check(nests=nests + 1)
```
karlhigley, line 47: I'm not entirely sure, but there might be an off by one error here? We start with `nests=0`, so the first nested layer has `nests=1` and that's the maximum nesting we allow (I think.)

karlhigley (commented):
Looks good! Left some thoughts and suggestions.

nvtabular/column_selector.py
jperez999: interesting... I can add that in now. No biggie and I will add a test also.  That way we support both methods.

nvtabular/column_selector.py
jperez999: oh man my bad. I thought we were running 2 levels deep... I didn't count start level.

karlhigley: No worries! It's so common and easy to do in this field. We have jokes about it for a reason. 😃

nvtabular/column_selector.py
```diff
@@ -93,7 +90,8 @@ def __rshift__(self, other):
         return nvtabular.ColumnGroup(self) >> other
 
     def __eq__(self, other):
-        if isinstance(other, ColumnSelector):
-            if other._names == self._names and other.subgroups == self.subgroups:
-                return True
+        if not isinstance(other, ColumnSelector):
+            return False
+        if other._names == self._names and other.subgroups == self.subgroups:
```
karlhigley, line 95: I think here you can return the value of the expression `other._names == self._names and other.subgroups == self.subgroups` directly, since this returns `True` when the expression is `True` and `False` when the expression is `False`.

tests/unit/test_column_selector.py
```diff
@@ -27,6 +27,12 @@ def test_constructor_works_with_single_strings_and_lists():
     assert selector2._names == ["a", "b", "c"]
 
 
+def test_constructor_no_names():
```
karlhigley, line 30: 👍🏻 Thanks!

nvtabular/ops/logop.py
```diff
@@ -35,6 +35,6 @@ class LogOp(Operator):
 
     @annotate("LogOp_op", color="darkgreen", domain="nvt_python")
     def transform(self, col_selector: ColumnSelector, df: DataFrameType) -> DataFrameType:
-        return _natural_log(df[col_selector.names].astype(np.float32) + 1)
+        return _natural_log(df[col_selector._names].astype(np.float32) + 1)
```
karlhigley, line 38: What's going on with this line? Should the `Log` op not apply to columns in subgroups? 🤔

karlhigley (approved):

PR #1017, opened 2021-08-04
Improve multi-gpu testing

albert17:
- Skips multi-gpu unit tests in single GPU systems
- Allows to pass GPU devices as parameter to integration tests


tests/unit/test_torch_dataloader.py
```diff
@@ -574,6 +575,9 @@ def rmspe_func(y_pred, y):
 
 
 @pytest.mark.skipif(importlib.util.find_spec("horovod") is None, reason="needs horovod")
+@pytest.mark.skipif(
+    cupy.cuda.runtime.getDeviceCount() <= 1, reason="This unittest requires multiple gpu's to run"
```
jperez999, line 579: This is legit!

tests/integration/test_notebooks.py
```diff
@@ -147,21 +151,35 @@ def test_movielens(asv_db, bench_info, tmpdir):
...
+        )
+        import hugectr
+
+        print(hugectr.__version__)
+        # _run_notebook(tmpdir, notebook, data_path, input_path, gpu_id=devices, clean_up=False)
+        return
+    except ImportError:
+        print("HugeCTR not installed, skipping " + notebook)
+
+    # Run Tensorflow or PyTorch
```
jperez999, line 167: doing inference before preprocessing and training?

tests/integration/test_notebooks.py
```diff
@@ -88,18 +88,18 @@ def test_criteo(asv_db, bench_info, tmpdir):
...
-def test_rossman(asv_db, bench_info, tmpdir):
+def test_rossman(asv_db, bench_info, tmpdir, devices):
     data_path = os.path.join(DATA_DIR, "rossman/data")
     input_path = os.path.join(DATA_DIR, "rossman/input")
     output_path = os.path.join(DATA_DIR, "rossman/output")
 
     # Run Donwload & Convert for all
     notebook = os.path.join(dirname(TEST_PATH), ROSSMAN_DIR, "01-Download-Convert.ipynb")
-    out = _run_notebook(tmpdir, notebook, data_path, input_path, gpu_id="0", clean_up=False)
+    out = _run_notebook(tmpdir, notebook, data_path, input_path, gpu_id=devices, clean_up=False)
```
jperez999, line 98: Does this work correctly... doesnt the converted rossman data end up in input path?  Then does step 2, the nvtabular preprocessing, run on the original data (i.e. datapath) instead of using the converted data in the input_path as input?

tests/integration/test_notebooks.py
albert17: That notebook actually has Download. preproc, and ETL. It does not follow the TF and PyTorch format; However, it is being updated this release; so I will remove it from this PR (That is one of the reasons it was commented) and I will addut following the standard format when it is ready

tests/integration/test_notebooks.py
albert17: It works correctly because the notebooks is using the output_dir and not the input_dir to get the date
```
DATA_DIR = os.environ.get(
    "OUTPUT_DATA_DIR", os.path.expanduser("~/nvt-examples/data/")
)
```

I didn't want to modify the notebook just in case someone is relying in that import and gets affected.

jperez999 (approved):

PR #1018, opened 2021-08-04
Rename `ColumnGroup` to `WorkflowNode` and move to sub-package

karlhigley:
Now that the column tracking functionality that used to be in `ColumnGroup` has moved to `ColumnSelector`, the class functions primarily as a node in the `Workflow` graph. This commit changes the class name to `WorkflowNode` to reflect that shift in emphasis.


rjzamora:
@karlhigley - Thanks for getting this ready so quickly!  I will reserve time tomorrow morning to give a careful review.  Before then, I would like to suggest that we allow  people to continue using the `ColumnGroup` name  for tne next release, but raise a deprecation warning.  It seems like we can assume that the user wants `ColumnSelector([...])` when they use `ColumnGroup([...])`?

Feel free to push back on this if my understanding is incorrect, or if you already covered this somewhere.

karlhigley:
@rjzamora We're actually working through a backlog of open PRs that depend on each other, so they're gonna keep coming one right after the other once I adjust them for feedback on the previous PRs. 😅 (The next one in this series is #1020.)

I did add [an alias](https://github.com/NVIDIA/NVTabular/pull/1018/files#diff-74446f303fc21b288745569a3696843a64f071e8865ee3e0aeefebfde0a022b8R32-R34) that makes `ColumnGroup` just another name for `ColumnSelector`, but I haven't figured out how to handle the deprecation warning. I guess `ColumnGroup` could be a sub-class of `ColumnSelector` whose `__init__` method raises the deprecation warning and then delegates everything else?

nvtabular/__init__.py
```diff
@@ -24,13 +24,17 @@
...
 
-ColumnGroup = column_group.ColumnGroup
+WorkflowNode = workflow.WorkflowNode
 ColumnSelector = column_selector.ColumnSelector
 Workflow = workflow.Workflow
 Dataset = io.dataset.Dataset
 
+# Provides an alias of ColumnSelector so that old usages of ColumnGroup to
+# select columns at the beginning of an operator chain don't break
+ColumnGroup = ColumnSelector
```
benfred, line 34: mentioned this elsewhere - but how about 
```suggestion
def ColumnGroup(columns):
   warnings.warn("ColumnGroup is deprecated, use ColumnSelector instead", DeprecationWarning)
   return ColumnSelector(columns)
```

benfred (approved):

PR #1020, opened 2021-08-04
Create `ColumnSchema` and `DatasetSchema` for tagging API

karlhigley:
Depends on #1018. Based on code from #990.


nvtabular/columns/schema.py
```diff
@@ -0,0 +1,106 @@
...
+    def __eq__(self, other):
+        if not isinstance(other, DatasetSchema):
+            return False
+
+        if len(self.column_schemas) != len(other.column_schemas):
+            return False
+
+        return all(column in other.column_schemas for column in self.column_schemas)
+
+    def __add__(self, other):
```
benfred, line 97: should we add a ```__sub__``` method too?

karlhigley: Maybe? The shape of this code is largely determined by the code that's coming next, and we haven't run into a use case for it yet.

benfred: fair enough - can always add later =)

benfred (approved):
lgtm!

I'm still not sure about ```DatasetSchema``` or ```Schema``` - but I'm good with either really.

karlhigley:
@benfred I'm still not sure about `DatasetSchema` either. Someone smart whose opinion I trust unironically suggested `ColumnGroupSchema` while we were cooking dinner the other day, but I think that would just be confusing with the `ColumnGroup->ColumnSelector` alias. There's also `ColumnSchemaGroup`, which is accurate but still pretty wordy. I've been wondering if `DataSchema` is too generic?

karlhigley:
(I'm gonna merge it as is for lack of a clearly better alternative, but name changes are still on the table.)

benfred:
> (I'm gonna merge it as is for lack of a clearly better alternative, but name changes are still on the table.)

Sounds good! I think DatasetSchema is fine , and like you said we can always update later if something better comes to us

PR #1023, opened 2021-08-06
Fix conda recipe

benfred:
We have some native dependencies - and need pybind11 and protoc to build the package.
Add these to the conda recipe. Also no longer mark our conda packages as 'noarch'.

The install_requires changes in setup.py also weren't playing well with conda,
disable during the conda build.


jperez999 (approved):

benfred:
rerun tests

PR #1028, opened 2021-08-10
Add sample_dtypes method to Dataset

rjzamora:
The current `Dataset.fit` logic uses a "dangerous" `to_ddf().head(1)` call to sample the "real" pandas/cudf data to infer the data types.  It makes sense that we cannot trust the `dtypes` attribute of the underlying Dask collection (for cudf-based collections), but there is no guarentee that `head(n)` will return the first `n` rows of the dataset **unless** there are >=n rows in the first Dask partition.  To avoid this problem, this PR adds a simple `Dataset.sample_dtypes` method that will iterate through the Dask partitions until a non-empty partition is found.  Although this may be slow for a large Dask collection with many empty partitions, the common case should have the same performance (and we are clearly fixing a bug).


rjzamora:
cc @oyilmaz-nvidia

karlhigley:
This looks handy for the upcoming schema work. 👍🏻

karlhigley (approved):

oyilmaz-nvidia (approved):
Thanks Rick! Just tested it and it worked.

PR #1033, opened 2021-08-11
Updates notebooks to be platform agnostic

albert17:


examples/advanced-ops-outbrain/01-Download-Convert.ipynb
```diff
@@ -52,8 +52,11 @@
     "import os\n",
     "\n",
     "import cupy\n",
-    "import cudf\n",
-    "import rmm"
+    "import rmm\n",
```
jperez999, line 55: we want to require rmm even though we dont have cudf? What uses rmm other than dask?

examples/advanced-ops-outbrain/01-Download-Convert.ipynb
```diff
@@ -116,23 +119,21 @@
...
    "outputs": [],
    "source": [
-    "from cudf import read_csv\n",
-    "\n",
     "# use managed memory for device memory allocation\n",
     "rmm.reinitialize(managed_memory=True)\n",
     "\n",
     "# Merge all the CSV files together\n",
-    "documents_meta = read_csv(DATA_BUCKET_FOLDER + \"documents_meta.csv\", na_values=[\"\\\\N\", \"\"])\n",
+    "documents_meta = df_lib.read_csv(DATA_BUCKET_FOLDER + \"documents_meta.csv\", na_values=[\"\\\\N\", \"\"])\n",
```
jperez999, line 126: instead of calling df_lib everywhere why not import read_csv from df_lib after its set?

examples/advanced-ops-outbrain/02-ETL-with-NVTabular.ipynb
```diff
@@ -152,9 +155,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "documents_categories_cudf = cudf.read_csv(DATA_BUCKET_FOLDER + \"documents_categories.csv\")\n",
-    "documents_topics_cudf = cudf.read_csv(DATA_BUCKET_FOLDER + \"documents_topics.csv\")\n",
-    "documents_entities_cudf = cudf.read_csv(DATA_BUCKET_FOLDER + \"documents_entities.csv\")\n",
+    "documents_categories_cudf = df_lib.read_csv(DATA_BUCKET_FOLDER + \"documents_categories.csv\")\n",
```
jperez999, line 158: Same as previous comment

jperez999 (changes requested):
Lets try to abstract out the df_lib package. I think we should use from df_lib import read_... to get out the necessary functionality in each notebook... This way we dont have to ref df_lib everywhere. Does that make sense?

examples/advanced-ops-outbrain/01-Download-Convert.ipynb
albert17: That is a good point. We are using dask even in CPU Version with no cudf installed. I assumed we would need rmm becuase of that, but maybe it is not needed. @rjzamora may be able to confirm.

examples/advanced-ops-outbrain/01-Download-Convert.ipynb
albert17: I agree it would be nice. I tried doing that, but I kept getting this error
```
>>> from nvtabular.dispatch import get_lib
>>> df_lib = get_lib()
>>> from df_lib import read_csv
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'df_lib'
```

It seems it tried to use the variable name as a library itself, instead of whatever library it is pointing. I spent some time looking online, but I was not able to find how to do it successfully. Do you have some idea?

jperez999: Can you create a second alias and move that around? i.e 
```
df_lib = get_lib()
read_csv = df_lib.read_csv
```

albert17: This actually works. Thanks Julio

examples/advanced-ops-outbrain/02-ETL-with-NVTabular.ipynb
albert17: Done

examples/advanced-ops-outbrain/01-Download-Convert.ipynb
albert17: @rjzamora could you confirm when possible please.

rjzamora: My understanding is that RMM (i.e. "RAPIDS Memory Manager") is only useful/available for cuda-enabled systems (at least for now).  RMM can be used with both cudf and cupy. Therefore, we shouldn't tie an RMM dependency to Dask, but to these other GPU-focused libraries.

jperez999: If rmm is tied to GPU only and then it should not be installed on systems that do not have GPUs. So it needs to be abstracted out IMO. @albert17

albert17:
rerun tests

benfred (approved):

PR #1034, opened 2021-08-11
Update and rerender movielens PyTorch training nb

rnyak:
The training notebook needed to be updated and rerun since, now, we have dictionary representation of  the tensors in each batch from the dataloader.


benfred (approved):

benfred:
rerun tests

PR #1041, opened 2021-08-17
Rework `WorkflowNode` column methods to be easier to think about

karlhigley:
I've been having trouble understanding what's going on with various parts of the code that call `input_column_names` and `flattened_columns`, because it's unclear whether those are the input or the output columns. This PR refactors those methods out and replaces them with `input_columns` and `output_columns` to make some of the other graph/workflow code we're working on easier to think through.

`input_column_names` is obviously for the input, but `flattened_columns` is ambiguous. It gets trickier to think about when `WorkflowNode.selector` is added to the picture, because that's also currently a list of column names, which ought to represent the input columns for the current node but actually represents the input columns for the next node (i.e. the current node's output.)

We're trying to refactor our way out of this conundrum by changing some things about how the graph works:

* `WorkflowNode.selector` should represent the current node's input columns
* Then we wouldn't need separate "input nodes" at the beginning of the graph that have a selector but no op
* In order to make that work with `+` nodes, we need to be able to use the selector to grab additional columns from the root dataframe of the workflow
* In order to make that work, we need unambiguous input/output column methods on `WorkflowNode` to use in `Workflow._transform_ddf` (which this PR provides.)

### DIFF
--- a/nvtabular/inference/triton/__init__.py
+++ b/nvtabular/inference/triton/__init__.py
@@ -397,7 +397,7 @@ def _generate_nvtabular_config(
     if output_model == "hugectr":
         config.instance_group.append(model_config.ModelInstanceGroup(kind=2))
 
-        for column in workflow.output_node.input_column_names:
+        for column in workflow.output_node.input_columns.names:
             dtype = workflow.input_dtypes[column]
             config.input.append(
                 model_config.ModelInput(name=column, data_type=_convert_dtype(dtype), dims=[-1])
--- a/nvtabular/inference/triton/model.py
+++ b/nvtabular/inference/triton/model.py
@@ -41,7 +41,6 @@ from triton_python_backend_utils import (
 )
 
 import nvtabular
-from nvtabular import ColumnSelector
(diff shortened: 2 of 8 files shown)


jperez999 (approved):

karlhigley:
`input_column_names` is obviously for the input, but `flattened_columns` is ambiguous. It gets trickier to think about when `WorkflowNode.selector` is added to the picture, because that's also currently a list of column names, which ought to represent the input columns for the current node but currently represents the input columns for the _next_ node (i.e. the current node's output.)

We're trying to refactor our way out of this conundrum by changing some things about how the graph works:
* `WorkflowNode.selector` should represent the current node's input columns
* Then we wouldn't need separate "input nodes" at the beginning of the graph that have a selector but no op
* In order to make that work with `+` nodes, we need to be able to use the selector to grab additional columns from the root dataframe of the workflow
* In order to make that work, we need unambiguous input/output column methods on `WorkflowNode` to use in `Workflow._transform_ddf` (which this PR provides.)

PR #1048, opened 2021-08-18
Refactor workflow graph to avoid needing input nodes without ops

karlhigley:
This PR changes the Workflow graph structure to avoid needing input nodes that only contain a list of columns to fetch but don't have operators with transforms to apply. That change will help us postpone binding column selectors to dataset schemas until the `Workflow.fit()` step, which both enables abstract selectors like tags and allows the same workflow to be applied to datasets with somewhat different schemas.

This involves a number of changes to the workflow graph structure:

* The `ColumnSelector` on a `WorkflowNode` now describes the desired inputs to the node instead of the inputs for the next node.
* The `["a","b","c"] >> Operator` syntax now only creates a single `WorkflowNode` and sets its selector, instead of creating an input node with a selector and a following node with the operator.
* Fetching additional columns from the root dataframe that weren't provided by the parent nodes' outputs is now possible directly from a node's input column selector
* As a consequence, addition/concat nodes (`+`) where raw column name strings are added to `WorkflowNodes` no longer create separate input nodes for the additional columns
* Dependency columns are now also fetched from the root dataframe when they aren't produced by parent node transforms


nvtabular/workflow/workflow.py
```diff
@@ -311,6 +340,9 @@ def _transform_ddf(ddf, workflow_nodes, meta=None):
     if isinstance(workflow_nodes, WorkflowNode):
         workflow_nodes = [workflow_nodes]
 
+    if not workflow_nodes:
+        return ddf
```
karlhigley, line 344: @rjzamora Also want to call out this change, because I'm not sure what the performance implications are here. Since we've refactored out the concept of an input node (that only has a list of columns to select), there are now cases where `_transform_ddf` gets called with an empty list of parent nodes. This check avoids the rest of the function struggling to handle that case and throwing errors.

However, it also means we bypass this chunk of code, which never executes anymore:
```
# Check if we are only selecting columns (no transforms).
# If so, we should perform column selection at the ddf level.
# Otherwise, Dask will not push the column selection into the
# IO function.
if all((c.op is None and not c.parents) for c in workflow_nodes):
    return ddf[_get_unique(columns)]
```
We're not sure if we need to explicitly do something to push the column selection down into the IO function now that we don't have input nodes, or if we can delete that chunk of code and call it good. Thoughts?

nvtabular/workflow/workflow.py
```diff
@@ -148,6 +149,26 @@ def fit(self, dataset: Dataset):
...
+                    )
+                    addl_input_cols = set(workflow_node.input_columns.names) - set(
+                        parent_output_cols.names
+                    )
+
+                    if addl_input_cols:
+                        # transformed_ddf = _concat_columns(
+                        #   [transformed_ddf, ddf[list(addl_input_cols)]]
+                        # )
+                        transformed_ddf = dask.dataframe.multi.concat(
```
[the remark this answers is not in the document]
rjzamora: I do think this `multi.concat` call is a bit dangerous.  You are effectively assuming that you can concatenate a list of columns from the original ddf onto a transformed version of that ddf.  Given that we do not track `divisions` in NVTabular, I am surprised that this command works without setting `ignore_unknown_divisions=True`.  However, even if the command does "work", we are assuming that the number of rows cannot change within any NVTabular transformation.  I suspect this is not a promise we can/should make.

I think the only reliable solution would be to select the columns before `_transform_ddf` is called.  Is there an efficient way that we can construct the union of `addl_input_cols` and the transformation columns before `_transform_ddf` is called?

nvtabular/workflow/workflow.py
rjzamora: Good question.  I'm pretty sure that we will need to apply this same column projection in a new way now that the column selection is represented differently.  For example, we may want to add an optional `input_columns=` argument to `_transform_ddf`, and always start with the `ddf[input_columns]` projection (when input_columns is known) in that function.  In `fit`, above, we would pass in the union of `addl_input_cols` and `workflow_node.selector`.

We definitely want to apply column projections on the origin ddf whenever possible, otherwise Dask will be unable to push the projection into `read_parquet`.

Note that I may be misunderstanding the selector/node machinery here, so feel free to correct me.

rjzamora (commented):
Sorry it took so long for me to get to this today @karlhigley !

I had a chance to give a shallow review and address your explicit questions.  I will be happy to follow up in more detail tomorrow.

nvtabular/workflow/workflow.py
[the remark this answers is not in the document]
jperez999: if we do the concatenation before the `_transform_ddf` and lets say the transform is going to add rows, would that also break because we would still use selector to pull out target columns and then only transform those columns. This would cause problems on the reassembly, wouldn't it? Or are we now presuming that the ddf that goes into transform_ddf will always work on the entire input ddf?

nvtabular/workflow/workflow.py
karlhigley: That sounds right to me. We'll give it a try.

nvtabular/workflow/workflow.py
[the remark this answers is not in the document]
karlhigley: I'm not immediately sure if there's an efficient way to compute the columns ahead, but we'll see what we can do to refactor in that direction

karlhigley:
@rjzamora No worries! Those are helpful suggestions and we're going to have to do some rework to account for them anyway, so a deeper review can wait until we have a chance to do that.

nvtabular/workflow/workflow.py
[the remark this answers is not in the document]
rjzamora: >if we do the concatenation before the _transform_ddf and lets say the transform is going to add rows, would that also break because we would still use selector to pull out target columns and then only transform those columns.

Right. The transformations would not be allowed to change the row-count of individual columns.  They can depend on individual columns, but would need to transform the DataFrame as a whole.  However, this has always been an implicit requirement.

nvtabular/workflow/workflow.py
karlhigley: @rjzamora I think this is along the lines of what you were thinking? We end up pushing the additional column fetch down into `_transform_partition`, where it can happen in the same way as the other column concats.

karlhigley: As a result, the check to see if we're only selecting columns without transforms can move up to the top of this method so that we don't skip over it when there are no parents to pull output columns from.

benfred (commented):
I didn’t see anything wrong with the implementation, but the workflow code is starting to get complicated enough that I’m having some problems following along with it now.  This extra complexity seems to be mainly in handling dependencies: before this change dependencies were just another parent node that the workflow didn’t need to be aware of, but now we need to handle dependencies separately in the workflow.  Also, having dependencies be either a ColumnSelector or a WorkflowNode doesn’t seem quite right to me.

I don’t see a way to remove these complexities without maintaining input nodes in the graph though =(. 

Is the main benefit of this change to make the upcoming ‘fit_schema’ change easier?  Like is the the problem that this PR aims to solve that input nodes don’t have an operator associated with them, so it’s not clear how to calculate the output schema from them?

karlhigley:
@benfred Yeah, all of these PRs are trying to set up the next one. This one paves the path for `fit_schema()`, which will cascade schemas through the graph, applying selectors where applicable. I don't love everything about this either, but satisfying the constraints imposed by the existing API at the same time as making it possible to select columns by tags is really heckin' hard. I've been refactoring for a month, and this is the best direction I can come up with that makes this impossible task possible.

The issue with input nodes is less that it's hard to compute the schemas and more that they don't fit the same flow as other nodes or fit well with the new concepts that have been introduced. They're essentially the old version of `ColumnSelector`, which makes them redundant now, and doing the same thing two ways quickly gets confusing and difficult to work with. They also don't fit with the node invariants we're pushing toward (e.g. every node has an op) in service of the neat API stuff (e.g. subgraph chaining with `>>`) that y'all had me come up with.

karlhigley:
rerun tests

benfred (approved):
Since we need this for the couple of pr's - lets get this merged now.

I have some ideas on how to improve the dependency handling, but that can wait until we have the fit_schema changes in.

PR #1079, opened 2021-08-31
Adding Schema plumbing to ops

jperez999:
This PR adds schema plumbing to individual operators. Each operator is responsible for updating the schema with new columns (and new column names), updated dtypes, tags and properties to add to the schema for each column of focus (selected in input or newly created). This data will be stored in dictionaries with output_column names as keys.


nvtabular/ops/operator.py
```diff
@@ -84,7 +84,72 @@ def compute_output_schema(self, input_schema: Schema, col_selector: ColumnSelect
...
+        names_original, names_transformed = [], []
+        for original_column in col_selector.names:
+            # logic works for all ops except TE
+            new_cols = [new_column for new_column in output_names if original_column in new_column]
+            if new_cols:
+                for column in new_cols:
+                    if column not in names_transformed:
+                        names_original.append(original_column)
+                        names_transformed.append(column)
+        # if name is same from input schema means it was a used column,
```
karlhigley, line 97: It's a small thing, but some whitespace between the sections would make this easier to read and review.
```suggestion

        # if name is same from input schema means it was a used column,
```

nvtabular/ops/operator.py
```diff
@@ -84,7 +84,72 @@ def compute_output_schema(self, input_schema: Schema, col_selector: ColumnSelect
...
+        # in op logic to mutate against selector columns
+        new_names = [
+            name
+            for name in output_names
+            if name not in names_transformed and name not in input_schema.column_schemas
+        ]
+        new_col_schemas = []
+        # grab old column schemas and change the names
```
karlhigley, line 105:
```suggestion

        # grab old column schemas and change the names
```

nvtabular/ops/operator.py
```diff
@@ -84,7 +84,72 @@ def compute_output_schema(self, input_schema: Schema, col_selector: ColumnSelect
...
+        for orig_col_name, new_column_name in zip(names_original, names_transformed):
+            orig_col_schema = input_schema.column_schemas[orig_col_name]
+            new_col_schemas.append(orig_col_schema.with_name(new_column_name))
+        # create column Schemas for new columns
```
karlhigley, line 109:
```suggestion

        # create column Schemas for new columns
```

nvtabular/ops/operator.py
```diff
@@ -84,7 +84,72 @@ def compute_output_schema(self, input_schema: Schema, col_selector: ColumnSelect
...
+        for new_col_name in new_names:
+            if new_col_name not in input_schema.column_schemas:
+                # needs to be filled in
+                new_col_schemas.append(ColumnSchema(new_col_name))
+        # add in all tags properties and update dtypes
```
karlhigley, line 114:
```suggestion

        # add in all tags properties and update dtypes
```

nvtabular/ops/operator.py
```diff
@@ -84,7 +84,72 @@ def compute_output_schema(self, input_schema: Schema, col_selector: ColumnSelect
...
+    def _add_properties(self, column_schema):
+        # get_properties should return the additional properties
+        # for target column
+        target_column_properties = self._get_properties()[column_schema.name]
+        return column_schema.with_properties(target_column_properties)
+
+    def _update_dtype(self, column_schema):
+        return column_schema.with_dtype(self._get_dtype())
+
+    def _get_dtype(self):
```
karlhigley, line 135: Since each op has to override these, it seems like the name should make sense from the perspective of the op, so that the sub-class definitions of these functions read cleanly and make sense in context. Maybe `output_dtype`, `output_tags`, and `output_properties`?

nvtabular/tags.py
```diff
@@ -0,0 +1,24 @@
+from enum import Enum
+
+
+class DefaultTags(Enum):
```
karlhigley, line 4: I know the name of this class came from the original tagging POC code, but I'm not sure if these tags are really `Default` in a meaningful way. Maybe this could just be `Tags`?

nvtabular/tags.py
```diff
@@ -0,0 +1,24 @@
...
+    # Feature types
+    CATEGORICAL = "categorical"
+    CONTINUOUS = "continuous"
+    LIST = "list"
+    IMAGE = "image"
```
karlhigley, line 9: I don't think we're using `image` for anything yet, and we don't even have any functionality that support would using images (e.g. loading pre-trained image embeddings), so I'm inclined to leave this out. Not sure about the others in this list, but this one I'm pretty sure we can drop (and add back later when needed.)

nvtabular/tags.py
```diff
@@ -0,0 +1,24 @@
...
+    TEXT = "text"
+    TEXT_TOKENIZED = "text_tokenized"
+    TIME = "time"
+
+    # Feature context
+    USER = "user"
+    ITEM = "item"
+    ITEM_ID = ["item", "item_id"]
```
karlhigley, line 17: Seems like the type of the tags here should be consistent, so either `str` or `list` based on what's here now. I lean toward making them single strings and handling combinations by setting multiple tags on the schema object. Thoughts, @benfred?

tests/unit/ops/test_ops_schema.py
```diff
@@ -0,0 +1,66 @@
...
+
+from nvtabular import ColumnSchema, ColumnSelector, Schema, ops
+
+
+@pytest.mark.parametrize("properties", [{}, {"p1": "1"}])
+@pytest.mark.parametrize("tags", [[], ["TAG1", "TAG2"]])
+@pytest.mark.parametrize(
+    "op",
+    [
+        (ops.Bucketize, [1]),
```
karlhigley, line 11: It seems like this could be simpler if we went ahead and constructed the operator object with these args right here in this list. Then we wouldn't need [the code below that unpacks these tuples](https://github.com/NVIDIA/NVTabular/pull/1079/files#diff-3763c2abd8f8b665145b2bffd223028183a07aa894798c24e37b0e59de3c23fcR33-R39) and passes/splats the args into the operator's constructor.

tests/unit/ops/test_ops_schema.py
```diff
@@ -0,0 +1,66 @@
...
+        op = op()
+    for x in range(5):
+        all_cols.append(str(x))
+        column_schemas.append(ColumnSchema(str(x), tags=tags, properties=properties))
+    # Turn to Schema
+    schema = Schema(column_schemas)
+    # run schema through op
+    selector = ColumnSelector(selection)
+    new_schema = op.compute_output_schema(schema, selector)
+    # check schema otherside
```
karlhigley, line 48: Add a blank line before the comments so my old eyes can parse this code into meaningful chunks to read and understand, huh? 😄

karlhigley (changes requested):

tests/unit/ops/test_ops_schema.py
jperez999: I tried it but it didnt work well... sometimes the op needed to be reinitialized anyway... and this introduced more inconsistencies. So I dont know about this one.

jperez999:
rerun tests

nvtabular/tags.py
```diff
@@ -13,11 +13,24 @@ class Tags(Enum):
...
+
     # Target related
-    TARGETS = ["target"]
-    TARGETS_BINARY = ["target", "binary"]
-    TARGETS_REGRESSION = ["target", "regression"]
-    TARGETS_MULTI_CLASS = ["target", "multi_class"]
+    TARGETS = "target"
+
+    @property
+    def TARGETS_BINARY(self):
```
karlhigley, line 27: I was thinking of these as lower-cased method names. Now that I see it this way, I'm not sure which way makes more sense. Do we want to give the impression they're interchangeable with the singular tags or not? 🤔

benfred: I don't think using a property here is going to work - properties work on a instance of a class, and usage here will be on the class itself.

Like running this

```
print(repr(Tags.TARGETS_BINARY))
print(repr(Tags.TIME))
```

prints out 
```
<property object at 0x7fdc684c3cc0>
<Tags.TIME: 'time'>
```

karlhigley: Good point! They'd have to be class methods.

karlhigley (approved):
Left a few last comments re: style etc, but fine to merge once those are addressed

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

PR #1080, opened 2021-08-31
Adjust `WorkflowNode.__add__` to combine addition nodes during graph construction

karlhigley:
It's a pain to debug graph construction involving addition nodes when they don't get combined until the `Workflow` is created.


benfred (approved):
nice!

karlhigley:
rerun tests

PR #1083, opened 2021-09-01
Move dataset schema inference into `Dataset` and automatically fit `Workflow` to schema (as needed)

karlhigley:
This PR creates an `infer_schema` method on `Dataset` that uses `sample_dtypes` to get the column names and dtypes from the ddf, then uses that information to create `Schema` and `ColumnSchema` objects to represent the dataset. This allows `Workflow.fit()` to avoid reading in all the columns of the dataset when only some of them are needed.

To make the schema fitting process smooth and break as little calling code as possible, this PR also checks whether or not the output schema of the `Workflow` has already been computed at the beginning of the `fit` and `transform` methods, and runs `fit_schema` to compute the output schema if not.

### DIFF
--- a/nvtabular/io/dataset.py
+++ b/nvtabular/io/dataset.py
@@ -29,6 +29,7 @@ from dask.utils import natural_sort_key, parse_bytes
 from fsspec.core import get_fs_token_paths
 from fsspec.utils import stringify_path
 
+from nvtabular.columns.schema import ColumnSchema, Schema
 from nvtabular.dispatch import _convert_data, _hex_to_int, _is_dataframe_object
 from nvtabular.io.shuffle import _check_shuffle_arg
 from nvtabular.utils import global_dask_client
@@ -1050,6 +1051,22 @@ class Dataset:
             return result.compute()
         else:
             return result
+
+    def infer_schema(self, n=1):
+        """Create a schema containing the column names and inferred dtypes of the Dataset
+
+        Args:
+            n (int, optional): Number of rows to sample to infer the dtypes. Defaults to 1.
+        """
+        sampled_dtypes = self.sample_dtypes(n)
+        dtypes = dict(zip(sampled_dtypes.index, sampled_dtypes))
+
+        column_schemas = []
+        for column, dtype in dtypes.items():
+            col_schema = ColumnSchema(column, dtype=dtype)
+            column_schemas.append(col_schema)
+
+        return Schema(column_schemas)
 
     def sample_dtypes(self, n=1):
         """Return the real dtypes of the Dataset
--- a/nvtabular/workflow/workflow.py
+++ b/nvtabular/workflow/workflow.py
@@ -109,10 +109,11 @@ class Workflow:
         -------
         Dataset
         """
-        if not self.output_schema:
-            self.fit(dataset)
-
         self._clear_worker_cache()
+
+        if not self.output_schema:
+            self.fit_schema(dataset.infer_schema())
+
         ddf = dataset.to_ddf(columns=self._input_columns())
         return Dataset(
             _transform_ddf(ddf, self.output_node, self.output_dtypes),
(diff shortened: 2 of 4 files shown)


nvtabular/workflow/workflow.py
```diff
@@ -112,6 +109,9 @@ def transform(self, dataset: Dataset) -> Dataset:
         -------
         Dataset
         """
+        if not self.output_schema:
+            self.fit(dataset)
```
benfred, line 113: Running the full fit here doesn't seem quite right to me - I'd rather either throw an exception (saying that you should run fit first) or call fit_schema instead of fit.

nvtabular/io/dataset.py
```diff
@@ -1051,6 +1052,22 @@ def regenerate_dataset(
         else:
             return result
 
+    def infer_schema(self, n=1):
+        """Create a schema containing the column names and inferred dtypes of the Dataset
+
+        Args:
+            n (int, optional): Number of rows to sample to infer the dtypes. Defaults to 1.
+        """
+        sampled_dtypes = self.sample_dtypes(n)
```
benfred, line 1061: :+1: 
Long term I don't think we'll need to sample the dtypes (since we should have correct dtype information on the Dataset ddf after workflow.transform calls)   - but I think this is a the right thing to do for now.

benfred (approved):

nvtabular/workflow/workflow.py
karlhigley: Hmm, okay. What's the downside of running the full `fit` here? It seems like it provides a convenience from the standpoint of the calling code, but maybe I'm missing a negative impact of some kind.

nvtabular/io/dataset.py
karlhigley: We need to know the column names and dtypes before `Workflow.fit/transform` in order to compute which columns to read from the ddf though. 🙁

nvtabular/workflow/workflow.py
karlhigley: (Made the change, but still curious.)

nvtabular/io/dataset.py
benfred: The 'sample_dtypes' function was created originally because the dtype property on the ddf wasn't correct after going through workflow.transform (since we only propagated column names through the workflow graph, not full schema information). 

For getting the column names and dtype information for the input dataset, we don't need to sample at all - and could just use the dtypes on the ddf itself (like as done here as a fallback https://github.com/NVIDIA/NVTabular/blob/e9c0795ab68d4d42b13d3a73e888c73b1483fb73/nvtabular/io/dataset.py#L1067 ) .  The problem is that we can't currently rely on this for the output of the workflow graph, since the dtypes are getting mangled in sections of code like https://github.com/NVIDIA/NVTabular/blob/e9c0795ab68d4d42b13d3a73e888c73b1483fb73/nvtabular/workflow/workflow.py#L401-L407

nvtabular/workflow/workflow.py
benfred: I'd find it surprising if calling transform (without calling fit) automatically fit the workflow - we have a 'fit_transform' function that does this if the user wants to do both fit and transform - and have explicitly called out in the docs that we expect the workflow to be have fit called before running transform: 
https://github.com/NVIDIA/NVTabular/blob/e9c0795ab68d4d42b13d3a73e888c73b1483fb73/nvtabular/workflow/workflow.py#L100-L102

My worry about merging the fit & transform functionality is if the user accidentally tries to call transform on a validation dataset, on a workflow that hasn't been fit already. While this is fine for the schema (since there isn't actual data) - this will silently do the wrong thing if we fit the full workflow here.

nvtabular/io/dataset.py
karlhigley: Yeah, seems like getting the dtypes of the input dataset is the main use for `infer_schema`, but I suppose it's possible you'd want to run multiple workflow transforms in series, in which case reading the dtypes directly from the ddf wouldn't work so well. 🤔

jperez999 (approved):

Issue #1074, opened 2021-08-27
[FEA] Add options to reserve categorical indices in the Categorify() op


gabrielspmoreira:
**Is your feature request related to a problem? Please describe.**
Some pipelines trust that categorical features will have some reserved values after preprocessing.
For example, for sequential features (e.g. session-based recommendation, time series, text data) it is important to reserve a value for padding (e.g. 0) in the model side, as not all sequences might have the same length.
It is also important to reserve a value (e.g. 1) to map Out-Of-Vocabulary (OOV) categorical values that might appear in the transform() (e.g. during evaluation or inference)

**Describe the solution you'd like**
Create one argument for `Categorify()` op to set the desired `oov_index` (for which all OOV values will be mapped during the preproc) and another argument `start_index` to set the first value that should be used to encode known values (e.g. 2). We can also have a similar option for null values (`null_index`), so that they are mapped to a known index.
In this case, if we call `Categorify(oov_index=1, null_index=2, start_index=3)`, the index 0 could be safely used for padding for example. 
I think we don't need an `padding_index`, because it is specific to sequential features. It is better to give users flexibility to reserve a number of categorical values by using `start_index`

EvenOldridge:
@gabrielspmoreira Do we really need that level of complexity?  Can't we just specify the offset?  Or does categorify work differently if you specify oov, null, and start_index ranges.

Also do we want to handle the case where people specify ridiculous values?  What if I say oov = 100000000?

I think a simple reserved offset is more straightforward and less error prone.

gabrielspmoreira:
@EvenOldridge I agree we can simplify the solution.
With the offset (`start_index`) we solve the issue of reserving 0 for padding. 
And we could reserve the first two indices after `start_index` / offset for null and OOV during `transform()` (inference). 
What do you think @benfred ?

benfred:
I think we should simplify, and just specify a start_index with a default value of 0.

The implementation of Categorify has gotten pretty complex - so I thought it might help to put some pointers on where to get started with this:
 * There are two phases in Categorify: the 'fit' which calculates the mapping of input label to categorical id, and the 'transform' which applies the mapping. We'll need to change both of these to handle this option
[...]

PR #1093, opened 2021-09-02
[HELP-REQ] Add tests for issue #1074 and update Categorify args

lesnikow:
Add tests for issue #1074 on adding a start_index argument to
the Categorify op. The motivation for this issue is to allow an
offset for encoding out-of-vocabulary and other special values.

In this commit, we add a test function to test various values of our
new start_index arg in tests/unit/ops/text_ops.py and add a
start_value arg to the class signature for Categorify(),
documentation for its intended function in Categorify()'s
docstring, a start_value arg to FitOptions(), and documentation
for this new argument in FitOptions()'s docstring.

This is a draft pull request to query and check whether the 
implemented tests capture the functionality requested in issue #1074


lesnikow:
There should be 12 failing tests with the new test method `test_categorify_lists_with_start_index()`. These failing tests have been written to query whether the tests capture the intended functionality and to guide, test the implementation of the feature.

benfred (commented):
Thanks for this PR!

The test looks like its capturing the functionality we want, I've left some comments below.

tests/unit/ops/test_ops.py
```diff
@@ -504,6 +504,41 @@ def test_categorify_lists(tmpdir, freq_threshold, cpu, dtype, vocabs):
...
+        compare = df_out["Authors"].to_arrow().to_pylist()
+
+    if start_index == 1:
+        assert compare == [[1], [1, 4], [3, 2], [2]]
+
+    if start_index == 2:
+        assert compare == [[2], [2, 5], [4, 3], [3]]
+
+    if start_index == 16:
+        assert compare == [[16], [16, 19], [18, 17], [17]]
```
gabrielspmoreira, line 541: You could maybe convert the `compare` variable to a numpy vector and then having a single assert in the end for different start_index. Something like

```python
compare = np.array(compare)
expected = np.array([[1], [1, 4], [3, 2], [2]])
assert compare == (expected + start_index)
```

gabrielspmoreira:
@benfred I see that in the `Categorify._encode()` method there is the `na_sentinel` argument that I assume defines the default value that is used to encode: (a) null values and (b) OOV values (like this [line](https://github.com/NVIDIA/NVTabular/blob/8cfdd83fa644a0d4f8ff1e298c2c298dc4a0f9b3/nvtabular/ops/categorify.py#L1165) that fillsna after a left join with the mapping table). Is my understanding correct?

Should `start_index` be summed to `na_sentinel` in the `_encode()` method? Or it would be the user responsability to as `na_sentinel` a value equal or larger than `start_index`?

gabrielspmoreira:
@benfred Another question is whether in some point NVTabular trusts in the row indices of the parquet files created to store the mapping between old and new values. 
Because trusting in the indices with start_index could not work properly

nvtabular/ops/categorify.py
```diff
@@ -1042,7 +1068,30 @@ def _encode(
     cat_names=None,
     max_size=0,
     dtype=None,
+    start_index=1
 ):
+    """The _encode method is responsible for transforming a dataframe (taking the written 
+    out vocabulary file and looking up values to translate from say string inputs to numeric 
+    outputs) 
+
+    Args:
```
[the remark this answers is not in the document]
lesnikow: Ben, thank you for your review and comments. Let me take a look at your comments and address the points you raised.

nvtabular/ops/categorify.py
```diff
@@ -174,6 +174,12 @@ class Categorify(StatOperator):
         value will be `max_size - num_buckets -1`.  Setting the max_size param means that
         freq_threshold should not be given.  If the num_buckets parameter is set,  it must be
         smaller than the max_size value.
+    start_index: int, default 1
```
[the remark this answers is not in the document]
lesnikow: Ben, we could do this, and I personally would be for starting with 0 to be more Pythonic. However from what I understand so far, Categorify() is already starting to output categorical values at 1. For instance, in `nvtabular/test/units/ops.py`, we have 

```
            # again userB has highest frequency given lowest encoding
            assert compare_authors == [2, 5, 1, 3]
            assert compare_engaging == [1, 1, 2, 4]
        else:
            # Column combinations are encoded
            compare_engaging = (
                df_out["Author_Engaging User"].to_list()
                if cpu
                else df_out["Author_Engaging User"].to_arrow().to_pylist()
            )
            assert compare_engaging == [1, 4, 2, 3]
```

So I would suspect existing tests would break with a `start_index` default of 0. What are your thoughts, or perhaps we should do this update as a separate issue to maintain the compatibility when a user does not specify a value of `start_index`?

tests/unit/ops/test_ops.py
lesnikow: Gabriel, this sounds like a reasonable idea to me, though beyond the scope of this issue for me. The motivation for this current format is consistency with the other tests currently in this module. I would suggest that we would either keep this existing assert statements as is, or update all of the similar assert comparisons in the rest of the `Categorify` class as a separate issue.

nvtabular/ops/categorify.py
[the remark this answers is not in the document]
benfred: We're starting at 0 - its just that 0 is reserved for out of vocabulary tokens. If we set ```start_index=0``` as the default, everything should work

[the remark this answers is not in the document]
lesnikow: Interesting. Yes, let me take a look and address this point.

nvtabular/ops/categorify.py
```diff
@@ -1132,6 +1205,7 @@ def _encode(
     elif dtype:
         labels = labels.astype(dtype, copy=False)
 
+    labels = labels + (start_index - 1)
```
benfred, line 1208: I don't think this quite works with on the GPU with list columns - can you move  this line up to before the 'dispatch._encode_list_column' call (move to line 1202).  This will do the start_index adjustment before we encode as a list column and should fix the unittest failures you're seeing here

lesnikow: Intriguing, yes, let me try this update as you suggest.

lesnikow: Indeed this has resolved all the failing tests related to the `Categorify()` class.

It looks like there is one failing test for `test_notebooks.py` and two for `test_s3.py`.

Would you have a suggested way that I can debug these two, if these two failing tests are related to the changes in this PR?

I have the Jenkins build information available at this URL: http://10.20.13.93:8080/job/nvtabular_tests/3390/

lesnikow: Thanks Ben, it looks like the merging main with the changes in #1079 resolved this three last failing tests. I would say this current PR #1093 is ready for review and to be merged into main.

benfred (commented):
Looks mostly good! 

One thing that needs done here is adjusting the embedding sizes returned. Inside the `_get_embedding_dask` function we need to add the start_index to the output (in a similar way we do for hash_buckets).

lesnikow:
Ben, thanks for your review and feedback, this is good to know. I was not exactly clear how to adjust the `_get_embedding_dask()` method output to include this increased embedding size while keeping the intended output format. I looked for, but could not find, documentation on `_get_embedding_dask()`, so it would be helpful for me to hear from you about the expected return format for this method to have this PR merged into `main`. I will look forward to talking with you during the time that you mentioned you will setup.

benfred:
> Ben, thanks for your review and feedback, this is good to know. I am not exactly clear how to adjust the `_get_embedding_dask()` method output to include this increased embedding size while keeping the intended output format. I looked for, but could not find, documentation on `_get_embedding_dask()`, so it would be helpful for me to hear from you about the expected return format for this method to have this PR merged into `main`. I will look forward to talking with you during the time that you mentioned you will setup to finish this PR, and to start work on my next issue 1077, once this PR is merged.

The get_embeddings_dask change is really just to include the start_index in the number of rows returned. Changing this line

https://github.com/NVIDIA/NVTabular/blob/5060ce140be3a2fc54e4734e4c9adb421b4e92fb/nvtabular/ops/categorify.py#L566-L567

to be something like :

```python
        num_rows += start_index
        embeddings[col] = _emb_sz_rule(num_rows)
```
(after passing start_index into the function) should be sufficient.   We should also a check to the unittest that verifies the output is correct. Other tests check this function with:

https://github.com/NVIDIA/NVTabular/blob/5060ce140be3a2fc54e4734e4c9adb421b4e92fb/tests/unit/ops/test_ops.py#L248

I'll set up some time later today -

lesnikow:
@benfred This is helpful, thank you. Let me make a commit based on these change and draft a test for this change based on the example you mentioned.

lesnikow:
@benfred I pushed these changes to `_get_embeddings_dask()`. For me to implement a test for the intended outputs of this method, I am a bit mystified on one part. Could you explain to me, here:

```
def _emb_sz_rule(n_cat: int, minimum_size=16, maximum_size=512) -> int:
    return n_cat, min(max(minimum_size, round(1.6 * n_cat ** 0.56)), maximum_size)
```

what is `round(1.6 * n_cat ** 0.56)` doing?

tests/unit/ops/test_ops.py
```diff
@@ -505,6 +505,42 @@ def test_categorify_lists(tmpdir, freq_threshold, cpu, dtype, vocabs):
...
+    # as well as the values for the rest of the in-vocabulary items.
+    # In this group of tests below, there are no out-of-vocabulary items, so our start index
+    # value does not appear in the expected comparison object.
+    if start_index == 0:
+        assert compare == [[1], [1, 4], [3, 2], [2]]
+    elif start_index == 1:
+        assert compare == [[2], [2, 5], [4, 3], [3]]
+    elif start_index == 16:
+        assert compare == [[17], [17, 20], [19, 18], [18]]
+
```
benfred, line 542: Can you add a check for the embedding size here?

benfred:
> def _emb_sz_rule(n_cat: int, minimum_size=16, maximum_size=512) -> int:
    return n_cat, min(max(minimum_size, round(1.6 * n_cat ** 0.56)), maximum_size)
what is round(1.6 * n_cat ** 0.56) intended to do?

It's a heuristic for figuring out the number of columns (width) of the embedding table - you can ignore it for now.

benfred (commented):
this is looking mostly good! Can you update the unittest to add a check for get_embedding_sizes?

tests/unit/ops/test_ops.py
lesnikow: Yes, commit b8458ce implements unit tests for this method and the `start_index` arg. 

The commit 3670f17 implements the type updating in the docstrings.

benfred (approved):
looks good to me! thanks

PR #1095, opened 2021-09-03
Make make_feature_column_workflow work with vocabularies of varying size

andrey-klochkov-liftoff:
`make_feature_column_workflow` fails in `Categorify` if features have vocabularies of varying sizes. This fix changes the type of parameter `vocabs` of `Categorify` to be a dictionary instead of DataFrame so that clients can provide dictionaries of varying sizes.

This fixes NVIDIA#1062.


benfred (approved):
Thanks for this!

PR #1096, opened 2021-09-03
Refactor `Workflow` code to rely on schemas instead of selectors

karlhigley:
This is a giant messy PR that needs a lot of clean-up, but it makes everything except the notebook tests pass. The problem in the notebooks is that the inference code wants to remove (e.g. label) columns from workflows by iterating through the graph and dropping them from the node inputs. I don't think that can work quite the same way anymore, but I haven't figured out how to change it yet. Nonetheless, pushing this code to share the current state and make sure it lives somewhere other than on my machine.


nvtabular/ops/target_encoding.py
```diff
@@ -215,6 +212,9 @@ def dependencies(self):
         return self.dependency
 
     def output_column_names(self, columns):
+        if hasattr(self.target, "output_columns"):
+            self.target = self.target.output_columns
+
```
karlhigley, line 217: This should move to an override of a new `fit_schema` method on the `Operator` base class.

nvtabular/workflow/node.py
```diff
@@ -18,6 +18,8 @@
 
 from nvtabular.columns import ColumnSelector, Schema
 from nvtabular.ops import LambdaOp, Operator, internal
+from nvtabular.ops.internal.concat_columns import ConcatColumns
+from nvtabular.ops.internal.subset_columns import SubsetColumns
```
karlhigley, line 22: We should be able to import these in one line

nvtabular/workflow/node.py
```diff
@@ -65,16 +67,105 @@ def selector(self, sel):
...
+            if isinstance(dep, ColumnSelector):
+                schema += Schema(dep.names)
+            elif isinstance(dep, WorkflowNode):
+                schema += dep.output_schema
+            elif isinstance(dep, list):
+                for nested_dep in dep:
+                    if isinstance(nested_dep, ColumnSelector):
+                        schema += Schema(nested_dep.names)
+                    elif isinstance(nested_dep, WorkflowNode):
+                        schema += nested_dep.output_schema
```
karlhigley, line 83: This is the same code twice; we ought to be able to write it as a recursive method

nvtabular/workflow/node.py
```diff
@@ -65,16 +67,105 @@ def selector(self, sel):
...
+                        upstream_selector += ColumnSelector(dep.output_schema.column_names)
+                    elif isinstance(dep, ColumnSelector):
+                        upstream_selector += dep
+                    elif isinstance(dep, list):
+                        subgroup_selector = ColumnSelector()
+                        for elem in dep:
+                            if isinstance(elem, WorkflowNode):
+                                subgroup_selector += ColumnSelector(elem.output_schema.column_names)
+                            elif isinstance(elem, ColumnSelector):
+                                subgroup_selector += elem
```
karlhigley, line 129: This is that same code to iterate through lists of nodes and selectors again. Should be replaced with a method we can call here

nvtabular/workflow/node.py
```diff
@@ -65,16 +67,105 @@ def selector(self, sel):
...
+                    self.selector = upstream_selector
+
+                # For addition nodes, some of the operands are parents and
+                # others are dependencies so grab schemas from both
+                self.input_schema = self.parents_schema + self.dependency_schema
+
+            # If we're a subtraction node, we have to do some gymnastics to compute
+            # the schema, because operands may be in the parents or the dependencies
+            # or both
+            elif isinstance(self.op, SubsetColumns):
```
karlhigley, line 143: These chunks that compute the schemas should be things we can do on the `Operator` sub-classes themselves

nvtabular/workflow/node.py
```diff
@@ -133,54 +224,33 @@ def __add__(self, other):
...
+            # Add self as a parent
+            self.children.append(child)
+            child.parents.append(self)
+
+        # The right operand becomes a dependency
+        if isinstance(other, list):
+            converted_elem = []
+            for elem in other:
+                if not isinstance(elem, (ColumnSelector, WorkflowNode)):
+                    elem = ColumnSelector(elem)
```
karlhigley, line 243: This is another frequently repeated piece of code that iterates through and converts strings to selectors. Maybe another extractable method?

nvtabular/workflow/node.py
```diff
@@ -199,19 +269,39 @@ def __sub__(self, other):
...
-        child.op = internal.SubsetColumns(label=f"- {list(to_remove)}")
+            # Create a child node
+            child = WorkflowNode()
+            child.op = internal.SubsetColumns(label="-")
+
+            # Add self as a parent
+            self.children.append(child)
+            child.parents.append(self)
+
+        # The right operand becomes a dependency
```
karlhigley, line 284: Longer term, we shouldn't be using dependencies for this, but `parents` are only allowed to be `WorkflowNodes`, not `ColumnSelectors`. Might be resolved by the proposed refactor to make selection into a node op?

nvtabular/workflow/node.py
```diff
@@ -252,22 +342,44 @@ def parents_schema(self):
...
     def input_columns(self):
+        if not self.input_schema:
+            raise RuntimeError(
+                "The input columns aren't computed until the workflow "
+                "is fit to a dataset or input schema."
+            )
+
         if self.selector:
+            # To maintain column groupings
             return self.selector
```
karlhigley, line 353: Do we still need to return the selector here? What's using this that needs the groupings? Could it ask for the selector directly?

nvtabular/workflow/node.py
```diff
@@ -252,22 +342,44 @@ def parents_schema(self):
...
+    @property
+    def dependency_nodes(self):
+        nodes = []
+        for dep in self.dependencies:
+            if isinstance(dep, WorkflowNode):
+                nodes.append(dep)
+            elif isinstance(dep, list):
+                for nested_dep in dep:
+                    if isinstance(nested_dep, WorkflowNode):
+                        nodes.append(nested_dep)
```
karlhigley, line 380: Same iterate through nested lists code from before, should be easier to compute than this

nvtabular/workflow/node.py
```diff
@@ -279,10 +391,18 @@ def label(self):
 
     @property
     def _cols_repr(self):
-        cols = ", ".join(map(str, self.selector[:3]))
-        if len(self.selector) > 3:
-            cols += "..."
-        return cols
+        if self.input_schema:
```
karlhigley, line 394: This fixes the string representation of the columns, but there's also an issue where nodes that are only dependencies and not parents think they're output nodes (according to the string representation.) We might be able to get around that by making the nodes that depend on them children?

nvtabular/workflow/workflow.py
```diff
@@ -124,7 +127,7 @@ def transform(self, dataset: Dataset) -> Dataset:
 
     def fit_schema(self, input_schema: Schema) -> "Workflow":
         schemaless_nodes = {
-            node: _get_schemaless_nodes(node.parents)
+            node: _get_schemaless_nodes(node.parents + node.dependency_nodes)
```
karlhigley, line 130: This is a place an `upstream_nodes` method or something would be useful so we don't have to think about this all the time

nvtabular/workflow/workflow.py
```diff
@@ -181,7 +184,10 @@ def fit(self, dataset: Dataset) -> "Workflow":
         # Get a dictionary mapping all StatOperators we need to fit to a set of any dependant
         # StatOperators (having StatOperators that depend on the output of other StatOperators
         # means that will have multiple phases in the fit cycle here)
-        stat_ops = {op: _get_stat_ops(op.parents) for op in _get_stat_ops([self.output_node])}
+        stat_ops = {
+            op: _get_stat_ops(op.parents + op.dependency_nodes)
```
karlhigley, line 188: `upstream_nodes` again

nvtabular/workflow/workflow.py
```diff
@@ -364,11 +370,16 @@ def clear_stats(self):
...
         for node in iter_nodes([self.output_node]):
-            parent_output_cols = []
-            for parent in node.parents:
-                parent_output_cols += parent.output_columns.names
-            parent_output_cols = _get_unique(parent_output_cols)
-            input_cols += list(set(node.input_columns.names) - set(parent_output_cols))
+            upstream_output_cols = []
+            upstream_nodes = []
+            upstream_nodes += node.parents or []
+            upstream_nodes += node.dependency_nodes or []
```
karlhigley, line 376: `upstream_nodes` again

nvtabular/workflow/workflow.py
```diff
@@ -440,12 +451,15 @@ def _transform_partition(root_df, workflow_nodes, additional_columns=None):
         addl_input_cols = set(node.dependency_columns.names)
 
         # Build input dataframe
-        if node.parents:
+        if node.parents or node.dependency_nodes:
```
karlhigley, line 454: `upstream_nodes` again and again below

nvtabular/workflow/workflow.py
```diff
@@ -461,6 +475,9 @@ def _transform_partition(root_df, workflow_nodes, additional_columns=None):
             unseen_columns = set(node.input_columns.names) - seen_columns
             addl_input_cols = addl_input_cols.union(unseen_columns)
 
+            # TODO: Find a better way to remove dupes
+            addl_input_cols = addl_input_cols - set(input_df.columns)
```
karlhigley, line 479: We should look at this with fresh eyes and either figure out why overlapping columns are happening or find a better fix

tests/unit/framework_utils/test_tf_feature_columns.py
```diff
@@ -21,4 +23,7 @@ def test_feature_column_utils():
     ]
 
     workflow, _ = nvtf.make_feature_column_workflow(cols, "target")
-    assert workflow.output_node.selector.names == ["vocab_1", "vocab_2", "target"]
+
+    schema = Schema(["vocab_1", "vocab_2", "target"])
+    workflow.fit_schema(schema)
```
karlhigley, line 28: There's no dataset to fit here so I used a schema, but this seems like a bad user experience to have to manually fit. Maybe we can move that inside `make_feature_column_workflow` and fit it to an inferred schema automatically?

karlhigley (commented):
Left notes about what to clean up while it's fresh in my mind

nvtabular/workflow/node.py
karlhigley: `Workflow` runs operator transforms with the result of the `input_columns` method as the selector, and the ops need the grouping, so we do in fact have to return the selector here when it's present.

karlhigley:
@oyilmaz-nvidia Adding you as a reviewer here since we ended up needing to make some changes to the way inference interacts with workflows. Tests pass so it should be okay, but never bad to get more eyes on something and you have more context on inference than we do so you may spot stuff we didn't.

jperez999 (approved):

nvtabular/workflow/node.py
jperez999: like `from nvtabular.ops.internal import ConcatColumns, SubsetColumns`

nvtabular/workflow/node.py
```diff
@@ -133,54 +175,27 @@ def __add__(self, other):
...
-        if added_node:
-            if isinstance(added_node.op, internal.ConcatColumns):
-                child.parents += added_node.parents
-                for parent in added_node.parents:
-                    parent.children.append(child)
-                    parent.children.remove(added_node)
-            else:
-                child.parents.append(added_node)
-                added_node.children.append(child)
+        if isinstance(other, WorkflowNode) and isinstance(other.op, internal.ConcatColumns):
```
jperez999, line 195: could use a comment describing why situation is different for concat columns op.

nvtabular/workflow/workflow.py
jperez999: Here your finding the columns that come from root df right? I guess we should be getting this information from the input_schema for the node op, instead of the input_df, is that way you mean?

nvtabular/workflow/node.py
karlhigley: Good point! Added one.

nvtabular/workflow/workflow.py
karlhigley: I'm not even sure what I mean TBH. I just remember writing this comment when you pointed out this was a little sketchy. 😅

PR #1100, opened 2021-09-08
remove iteration from column selector class and supporting functionality

jperez999:
removes iteration and supporting functionality from column selector. Prepping for column selector to be able to have tags/names.

### DIFF
--- a/nvtabular/columns/selector.py
+++ b/nvtabular/columns/selector.py
@@ -86,15 +86,6 @@ class ColumnSelector:
         for col_sel0 in self.subgroups:
             col_sel0._nested_check(nests=nests + 1)
 
-    # def __getitem__(self, index):
-    #     return self._names[index]
-
-    # def __len__(self):
-    #     return len(self.names)
-
-    # def __iter__(self):
-    #     return iter(self.names)
-
     def __add__(self, other):
         if other is None:
             return self
--- a/nvtabular/columns/selector.py
+++ b/nvtabular/columns/selector.py
@@ -53,6 +53,8 @@ class ColumnSelector:
                 plain_names.append(name)
             elif isinstance(name, nvtabular.WorkflowNode):
                 raise ValueError("ColumnSelectors can not contain WorkflowNodes")
+            elif isinstance(name, ColumnSelector):
+                self.subgroups.append(name)
             else:
                 self.subgroups.append(ColumnSelector(name))
         self._names = plain_names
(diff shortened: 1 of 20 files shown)


nvtabular/columns/selector.py
```diff
@@ -84,14 +86,14 @@ def _nested_check(self, nests=0):
         for col_sel0 in self.subgroups:
             col_sel0._nested_check(nests=nests + 1)
```
karlhigley, line 88:
```suggestion

```

nvtabular/columns/selector.py
```diff
@@ -84,14 +86,14 @@ def _nested_check(self, nests=0):
...
 
-    def __getitem__(self, index):
-        return self._names[index]
+    # def __getitem__(self, index):
+    #     return self._names[index]
```
karlhigley, line 90:
```suggestion
```

nvtabular/columns/selector.py
```diff
@@ -84,14 +86,14 @@ def _nested_check(self, nests=0):
...
 
-    def __len__(self):
-        return len(self.names)
+    # def __len__(self):
+    #     return len(self.names)
 
-    def __iter__(self):
-        return iter(self.names)
+    # def __iter__(self):
+    #     return iter(self.names)
```
karlhigley, line 96: Let's go ahead and remove this instead of commenting it out. If we need to retrieve it later for some reason, we can always find it in the commit history

nvtabular/ops/categorify.py
```diff
@@ -1160,9 +1164,14 @@ def _get_multicolumn_names(col_selector, df_columns, name_sep):
...
 
 
-def _is_list_col(workflow_node, df):
-    has_lists = any(dispatch._is_list_dtype(df[col]) for col in workflow_node)
-    if has_lists and len(workflow_node) != 1:
+def _is_list_col(col_selector, df):
+    if isinstance(col_selector, list):
+        import pdb
+
+        pdb.set_trace()
```
karlhigley, line 1171: Stray pdb here

nvtabular/workflow/node.py
```diff
@@ -105,6 +105,8 @@ def compute_schemas(self, root_schema):
                 left_operand = operands.pop(0)
 
                 left_operand_schema = _combine_schemas([left_operand])
+                # upstream schema gets lambda applied
+
```
karlhigley, line 109: Stray comment here from brainstorming solutions to a different problem (I think)

nvtabular/workflow/node.py
```diff
@@ -376,7 +378,7 @@ def _filter_by_type(elements, type_):
     return results
 
 
-def _combine_schemas(elements):
+def _combine_schemas(elements, input_schema=None):
```
karlhigley, line 381: I think this was also a change for later that wasn't part of the refactor in this PR

karlhigley (approved):
Left some small comments about stray code etc, but otherwise looks good

PR #1112, opened 2021-09-13
working categorify export to properties

jperez999:
This is the first example of an operator writing out operator statistics specific to a column after gathering those stats during the fit stage.

### DIFF
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -175,6 +175,15 @@ class Categorify(StatOperator):
         value will be `max_size - num_buckets -1`.  Setting the max_size param means that
         freq_threshold should not be given.  If the num_buckets parameter is set,  it must be
         smaller than the max_size value.
+    start_index: int, default 0
+        The start index where Categorify will begin to translate dataframe entries
+        into integer values, including an initial out-of-vocabulary encoding value.
+        For instance, if our original translated dataframe entries appear
+        as [[1], [1, 4], [3, 2], [2]], with an out-of-vocabulary value of 0, then with a
+        start_index of 16, Categorify will reserve 16 as the out-of-vocabulary encoding value,
+        and our new translated dataframe entry will now be [[17], [17, 20], [19, 18], [18]].
+        This parameter is useful to reserve an intial segment of non-negative translated integers
+        for special user-defined values.
     """
 
     def __init__(
@@ -192,6 +201,7 @@ class Categorify(StatOperator):
         num_buckets=None,
         vocabs=None,
         max_size=0,
+        start_index=0,
     ):
 
         # We need to handle three types of encoding here:
@@ -244,6 +254,7 @@ class Categorify(StatOperator):
         self.cat_cache = cat_cache
         self.encode_type = encode_type
         self.search_sorted = search_sorted
+        self.start_index = start_index
 
         if self.search_sorted and self.freq_threshold:
             raise ValueError(
@@ -439,6 +450,7 @@ class Categorify(StatOperator):
                     cat_names=cat_names,
                     max_size=self.max_size,
                     dtype=self.dtype,
+                    start_index=self.start_index,
                 )
             except Exception as e:
                 raise RuntimeError(f"Failed to categorical encode column {name}") from e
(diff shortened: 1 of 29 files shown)


tests/unit/ops/test_ops_schema.py
```diff
@@ -60,3 +62,195 @@ def test_schema_out(tags, properties, selection, op):
...
+@pytest.mark.parametrize(
+    "op_routine",
+    [
+        [ops.Categorify()],
+        [ops.Clip(min_value=10), ops.Categorify()],
+        [ops.Categorify(), ops.Rename(postfix="_test")],
+        [ops.Clip(min_value=10), ops.Categorify(), ops.Rename(postfix="_test")],
+    ],
+)
+def test_categorify_schema_properties_blank(properties, tags, op_routine):
```
jperez999, line 126: All subsequent tests are the same as above just with different parameters for tags and properties going into the workflow. If you try to collapse this down the way ti should be... (as a second entry to parameters in the first call, you will see that the first test of each "op_routine" will pass while all subsequent tests will fail. This is why this has been broken up. It's very ugly and hurts the eyes, but I feel the different behaviors need to be explicitly verified.

karlhigley: Yeah, unfortunate it doesn't work the way we'd like here. Since the code of the tests is mostly the same, maybe we could extract a (non-test) helper method for the body of the test? Like most of the stuff above the assertions or something along those lines?

nvtabular/workflow/workflow.py
```diff
@@ -248,6 +245,8 @@ def fit(self, dataset: Dataset) -> "Workflow":
         output_dtypes = self.transform(dataset).sample_dtypes()
         self.output_dtypes = dict(zip(output_dtypes.index, output_dtypes))
 
+        self._zero_output_schemas()
+        self.fit_schema(dataset.infer_schema())
```
karlhigley, line 249: Looks like this requires fitting the schema twice? That's probably fine for now, but I wonder if we should split `fit_schema` into computing columns/tags vs computing properties at some point in the future.

nvtabular/workflow/workflow.py
```diff
@@ -390,6 +389,17 @@ def _clear_worker_cache(self):
         else:
             clean_worker_cache()
 
+    def _zero_output_schemas(self):
+        """
+        Zero out all schemas in order to rerun fit schema after operators
+        have run fit and have stats to add to schema.
+        """
+        for node in iter_nodes(self.output_node.parents_with_dep_nodes):
```
karlhigley, line 397: Would `iter_nodes([output_node])` not work here? Seems like we should be able to iterate through all the nodes in the loop without having to handle `output_node` as a special case here.

tests/unit/ops/test_ops_schema.py
```diff
@@ -60,3 +62,195 @@ def test_schema_out(tags, properties, selection, op):
...
+        schema1 = workflow_schema_out.column_schemas[column_name]
+        assert "domain" in schema1.properties
+        embeddings_info = schema1.properties["domain"]
+        assert isinstance(embeddings_info, tuple)
+        # should always exist, represents unkown
+        assert embeddings_info[0] == 0
+        assert embeddings_info[1] == new_gdf[column_name].max() + 1
+    workflow_schema_out = {}
+    df_dict = {}
+    schema = {}
```
karlhigley, line 112: Is there some kind of coupling across test runs that makes it necessary to clear these out at the end?

nvtabular/ops/categorify.py
```diff
@@ -477,6 +477,27 @@ def _get_dtypes(self):
...
     fit_finalize.__doc__ = StatOperator.fit_finalize.__doc__
 
+    def compute_output_schema(self, input_schema: Schema, col_selector: ColumnSelector) -> Schema:
+        if not col_selector:
+            col_selector = ColumnSelector(input_schema.column_names)
+        col_selector = self.output_column_names(col_selector)
+        for column_name in col_selector.names:
+            if column_name not in input_schema.column_schemas:
+                input_schema += Schema([column_name])
+        return super().compute_output_schema(input_schema, col_selector)
```
karlhigley, line 487: Is there something different about this `compute_output_schema` method, or is this left over from before the other refactoring PR that removed the dupes?

jperez999: this was a dupe

karlhigley (approved):

jperez999:
rerun tests

jperez999:
rerun tests

PR #1115, opened 2021-09-14
select by tags

jperez999:


nvtabular/ops/fill.py
```diff
@@ -71,6 +71,13 @@ def inference_initialize(self, col_selector, inference_config):
...
             col_selector = ColumnSelector(input_schema.column_names)
+        if col_selector.tags:
+            tags_col_selector = ColumnSelector(tags=col_selector.tags)
+            filtered_schema = input_schema.apply(tags_col_selector)
+            col_selector += ColumnSelector(filtered_schema.column_names)
+
+            # zero tags because already filtered
+            col_selector._tags = []
         output_schema = Schema()
         for column_name in col_selector.names:
```
benfred, line 82: I'm wondering if we should move the 'filter schema by tags' logic into the ColumnSelector.apply and then do something like 

```suggestion
        if col_selector:
            input_schema = input_schema.apply(col_selector)

        output_schema = Schema()
        for column_name in input_schema.column_names:
```

There are a couple reasons I'm suggesting this:
 * mutating the col_selector input (zeroing out tags and then appending column names to it etc) isn't what I'd expect from this method based off the name: I'd assume that the col_selector input would be unchanged 
 * Using ColumnSelector to store a concrete list of column names is overriding the point of a column selector (abstract representation of columns that are input here). 
 * Moving the filter by tag logic into ColumnSelector.apply will let us reuse some code across operators

karlhigley: I'm with you on the point about mutation, though we haven't yet done anything to encourage people to explicitly create and store selectors. We should definitely address it, but I don't think it makes an immediate difference that's worth holding up this PR over.

For the other two points, we're aware of the issues but have a slightly different way of addressing them planned:
* What we really have is three things: abstract selectors (e.g. tags, lambdas), concrete selections (with grouping), (flat) dataset schemas. We're overloading the concept of a `ColumnSelector` here in the interest of time, but what we'd like to do (given the time) is peel apart those three ideas into separate classes.
* We definitely do want to re-use the logic for turning abstract selectors into concrete selections, but we're not entirely sure where it's going to live yet. It _could_ live in `Schema.apply()`, but it's kinda funky that the `Schema` is responsible for knowing how selection works (in addition to several other responsibilities) when we already have a class called `ColumnSelector`. This is a key area of refactoring we'd like to tackle in the first part of the next release cycle.

karlhigley: Point being: we are not under the impression that the current state of things is great, but we are under the impression that it would be useful to get this functionality in before the code freeze, even if we immediately start refactoring it after the upcoming bug-bash (as we plan to.)

benfred: Agreed - sounds like this isn't easy to do, and we can clean up after this release.

benfred (approved):

Issue #1088, opened 2021-09-02
[QST] How to efficiently read data from Google Cloud Storage (gs://)?


leiterenato:
**What is your question?**

I am using NVTabular to do the preprocessing of large parquet files stored in a bucket in Google Cloud Storage (~11GB each file - criteo dataset).
The end goal is to find a solution to efficiently preprocess large volumes of data (> 100TB) reading parquet files directly from GCS (protocol `gs`).

For an initial benchmark I used [this script](https://github.com/NVIDIA/NVTabular/blob/main/bench/examples/dask-nvtabular-criteo-benchmark.py) from the NVTabular repository.
In [line 226](https://github.com/NVIDIA/NVTabular/blob/6635972632de86061274558e0a3e224e8254b6ba/bench/examples/dask-nvtabular-criteo-benchmark.py#L216) it creates a Dataset (universal external-data wrapper) from a list of GCS paths which instantiates a '[PaquetDatasetEngine](https://github.com/NVIDIA/NVTabular/blob/6635972632de86061274558e0a3e224e8254b6ba/nvtabular/io/parquet.py#L59)' as '[self.engine](https://github.com/NVIDIA/NVTabular/blob/6635972632de86061274558e0a3e224e8254b6ba/nvtabular/io/dataset.py#L282)'.

NVTabular methods [.fit](https://github.com/NVIDIA/NVTabular/blob/6635972632de86061274558e0a3e224e8254b6ba/bench/examples/dask-nvtabular-criteo-benchmark.py#L221) and [.transform](https://github.com/NVIDIA/NVTabular/blob/6635972632de86061274558e0a3e224e8254b6ba/bench/examples/dask-nvtabular-criteo-benchmark.py#L225) receive the Dataset instance and they both call the method [Dataset.to_dff](https://github.com/NVIDIA/NVTabular/blob/6635972632de86061274558e0a3e224e8254b6ba/nvtabular/workflow/workflow.py#L177) which returns the result of [dask_cudf.read_parquet](https://github.com/NVIDIA/NVTabular/blob/6635972632de86061274558e0a3e224e8254b6ba/nvtabular/io/parquet.py#L217).
[...]

PR #1119, opened 2021-09-16
Add optimized read_parquet path for remote storage

rjzamora:
**Notes**:

- ~This PR is **blocked** by #1081 (many of the apparent changes here actually belong to that PR)~
- These changes are intended to address #1088
- The optimizations added here are (mostly) copied from [cudf#9225](https://github.com/rapidsai/cudf/pull/9225)
    - The specific fsspec-based optimizations are described in a bit more detail in that PR
    - Once NVTabular is pinned to a new enough version of cudf (one where cudf#9225 is merged), the redundant code added here (in `nvtabular/io/fsspec_utils.py`) can be removed

**Performance [UPDATED 9/20/2021]**:

- Initial tests on a local workstation (reading from GCS over the internet) are very promising: With this PR, I am able to read through a 23-file, 1TB Dataset (Criteo) on two GPUs in ~9 minutes.  Without this PR, I cannot even read in a single file from that dataset in the same amount of time.
- I have confirmed that this PR leads to favorable performance on GCP (~15 min on an 8-GPU cluster when the input data is stored in GCS).  **Without this PR, the benchmark takes hours** (it is slow enough that I didn't bother with a full run). Note that the same benchmark is expected to finish in ~2-3 min when the input data is stored on a local SSD (however, my instance did not have a large-enough SSD to test this).  Command and output:

```
(rapids-21.08) rzamora@a100-nvtabular-gcs-testing-8g:~$ python NVTabular/bench/examples/dask-nvtabular-criteo-benchmark.py --part-mem-frac 0.125 --data-path gs://workshop-datasets/criteo-parque --out-path /h
ome/rzamora/scratch/ --out-files-per-proc 3 --device-limit-frac 0.75 --device-pool-frac 0.85 --cat-cache-high device -d 0,1,2,3,4,5,6,7 > bench.PR1119.out 2>&1


Dask-NVTabular DLRM/Criteo benchmark
--------------------------------------
partition size     | 5313159168
protocol           | tcp
device(s)          | 0,1,2,3,4,5,6,7
rmm-pool-frac      | 0.85
out-files-per-proc | 3
num_io_threads     | 0
shuffle            | PER_PARTITION
cats-on-device     | False
======================================
Runtime[s]         | 899.8842217922211
======================================
```

**GCP Instance Details**:

- Machine type: a2-highgpu-8g (96 vCPUs, 680 GB memory)
- CPU platform: Intel Cascade Lake
- GPUs: 8 x NVIDIA Tesla A100
- Zone: us-central1-c
- Disk: 1024GB SSD


nvtabular/io/dataset.py
```diff
@@ -1122,6 +1122,10 @@ def sample_dtypes(self, n=1):
         ``n`` rows of that partition to infer dtype info. If no
         non-empty partitions are found, use the Dask dtypes.
         """
+
+        if hasattr(self.engine, "_dtypes") and self.dtypes is None:
+            return self.engine._dtypes
+
```
rjzamora, line 1128: @jperez999 @karlhigley - This is a temporary hack that I needed for reasonable dask-criteo benchmark performance in google cloud (reading from GCS).  That benchmark made it pretty clear that we want to avoid reading an entire partition from disk to sample the dtypes (which is slow on remote filesystems).

**My plan for a "real" fix in this PR**: I am thinking that a reasonable approach here is to push the `sample_dtypes` logic into `DatasetEngine`.  This way, the disk-backed engines can overrid the implementation with optimized logic to avoid reading an entire partition; for parquet we can only read a single row-group and use cached dtype information if it was already used to calculate `row_groups_per_part`.

nvtabular/io/fsspec_utils.py
```diff
@@ -0,0 +1,331 @@
...
+            paths.append(path_or_fob)
+            starts.append(offset)
+            ends.append(offset + nbytes)
+        blocks = fs.cat_ranges(paths, starts, ends)
+        for block, (offset, nbytes) in zip(blocks, ranges):
+            local_buffer[offset : offset + nbytes] = np.frombuffer(
+                block,
+                dtype="b",
+            )
+        return
```
rjzamora, line 318:
```suggestion
```
Lets remove this. special logic util fsspec is actually released with [fsspec#744](https://github.com/intake/filesystem_spec/pull/744) in place.

rjzamora:
I have confirmed that this PR is back to ~15 mins on the dask-criteo benchmark with 8 GPUs on GCP (and barely over 2 mins when reading from NVMe)

**Reading from GCS on GCP**:

```
Dask-NVTabular DLRM/Criteo benchmark
--------------------------------------
partition size     | 5313159168
protocol           | tcp
device(s)          | 0,1,2,3,4,5,6,7
rmm-pool-frac      | 0.85
out-files-per-proc | 3
num_io_threads     | 0
shuffle            | PER_PARTITION
cats-on-device     | False
======================================
Runtime[s]         | 912.0400369167328
======================================
```

**Reading from NVMe on a DGX-1**:

```
Dask-NVTabular DLRM/Criteo benchmark
--------------------------------------
partition size     | 4261216256
protocol           | tcp
device(s)          | 0,1,2,3,4,5,6,7
rmm-pool-frac      | 0.8
out-files-per-proc | 3
num_io_threads     | 0
shuffle            | PER_PARTITION
cats-on-device     | False
======================================
Runtime[s]         | 126.32316994667053
======================================
```

This PR also fixes a small bug from #1127 that results in the reading of every partition in the dataset when `infer_schema` is called - See problematic task-stream screenshot below (where `fit` hasn't even started after >1 min):

Screen Shot 2021-09-21 at 10 05 47 PM

benfred (approved):

PR #1127, opened 2021-09-16
Get embedding sizes refactor

jperez999:


nvtabular/ops/categorify.py
```diff
@@ -558,30 +570,31 @@ def get_embedding_sizes(source, output_dtypes=None):
...
-            # (which could transform meaning of the get_embedding_sizes)
-            queue.extend(current.parents)
-
-    for column in output:
-        dtype = output_dtypes.get(column)
-        if dtype and dispatch._is_list_dtype(dtype):
+    # nodes = list(set(nvt.workflow.node.iter_nodes([output_node])))
+    # for current in reversed(nodes):
+    #     if current.op and hasattr(current.op, "get_embedding_sizes"):
+    #         output.update(current.op.get_embedding_sizes(current.output_schema.column_names))
```
karlhigley, line 586:
```suggestion
    # nodes = list(set(nvt.workflow.node.iter_nodes([output_node])))
    # for current in reversed(nodes):
    #     if current.op and hasattr(current.op, "get_embedding_sizes"):
    #         output.update(current.op.get_embedding_sizes(current.output_schema.column_names))
```

karlhigley (approved):
This looks good! There's one stray chunk of code commented out in the middle, but I'm approving and merging anyway in the interest of time. We can submit a second PR to remove the comments.

nvtabular/io/dataset.py
```diff
@@ -1097,18 +1101,33 @@ def infer_schema(self, n=1):
...
+            }
+            for partition_index in range(_ddf.npartitions):
+                _head = _ddf.partitions[partition_index].head(n)
+
+                if len(_head):
+                    for col in _head.columns:
+                        dtypes[col] = {
+                            "dtype": dispatch._list_val_dtype(_head[col]) or _head[col].dtype,
+                            "is_list": dispatch._is_list_dtype(_head[col]),
+                        }
```
rjzamora, line 1120: I added a special optimization in #1119 to avoid loading any of the partitions from remote storage (which is super slow). It looks like this change will now skip that optimization.  I'll need to double check that this introduces a performance regression in the criteo benchmark on GPC. If it does, I suggest we fix this asap.

rjzamora: **Update**: I addressed the previous comment in #1119. While making that change, I also realized that the highlighted code above will always read every partition in the dataset (which I am assuming is due to a missing `break` statement).

PR #1129, opened 2021-09-17
Avoid using cudf.read_parquet on the client process

rjzamora:
While testing #1119, I realized that NVTabular is using the client process to read in parquet files in some places where either (1) a worker process would be better suited for the job, or (2) using the parquet metadata would suffice.

This PR uses `delayed` to move some `read_parquet` operations to the workers, and also uses parquet metadata in one place where the actual data is not necessary.

cc @jperez999


benfred (approved):

benfred:
lgtm - thanks!

rjzamora:
Oops - Looks like `_reset_df_index` is designed to mutate the `categories` argument. Fix on the way.

PR #1135, opened 2021-09-21
fix type error in dtypes

jperez999:


benfred (commented):
The change looks good - but can we add a test that would catch this?

karlhigley (approved):

karlhigley:
@benfred We agree that it should be tested, but we're not entirely sure how to write that test at the moment. Given the timeframe, we're inclined to merge a fix and address comprehensive testing as part of a further refactor that introduces standard NVT dtypes and provides translations between NVT dtypes and Python, Pandas, and cuDF types.

karlhigley:
...yes, I am aware of how testing works and what the purpose is.

The repro is helpful though.

benfred (approved):

benfred:
rerun tests

Issue #1142, opened 2021-09-23
[BUG] Long time for KerasSequenceLoader to "reload" after each epoch 


nv-dmajchrowski:
**Describe the bug**
I'm measuring time for each batch to be loaded using KerasSequenceLoader and I'm experiencing long time to load the first batch after previous epoch ends. For example: 

``` 
...
Epoch 0 Iter 455 0.001s
Epoch 0 Iter 456 0.003s
Epoch 1 Iter 000 1.162s
Epoch 1 Iter 001 0.001s
Epoch 1 Iter 002 0.001s
...
Epoch 1 Iter 455 0.001s
Epoch 1 Iter 456 0.003s
Epoch 2 Iter 000 1.570s
Epoch 2 Iter 001 0.001s
Epoch 2 Iter 002 0.001s
```

**Steps/Code to reproduce bug**

1. Use NVTabular/examples/advanced-ops-outbrain/ jupyter notebooks as an example
2. In last notebook (03-Training) add following cell (right after `train_dataset_tf` definition):

```
import time
for epoch in range(3):
    t = time.time()
    for it, _ in enumerate(train_dataset_tf):
        time.sleep(0.001)
        it_padded = str(it).zfill(3)
        print("Epoch", epoch, "Iter", f"{it_padded}", f"{time.time() - t:.3f}s")
        t = time.time()
```

**Expected behavior**
No overhead from data loading when starting new epoch

**Environment details (please complete the following information):**
 - Environment location: Docker
 - Method of NVTabular install:
   - Merlin tensorflow container - [code](https://github.com/NVIDIA-Merlin/Merlin/blob/f5fb5d5b9ad12c3cfc3c879e379d2e4d1f212daf/docker/dockerfile.tf)
   - Checkout to nvtabular v0.6.1 - `RUN cd /nvtabular && git pull && git checkout v0.6.1`
 

**Additional context**
- Measured on DGX1V 32GB

rjzamora:
[...]

PR #1147, opened 2021-09-24
Add DataLoader.epochs() method and Dataset.to_iter(epochs=) argument

rjzamora:
This PR implements [the proposed API change](https://github.com/NVIDIA/NVTabular/issues/1142#issuecomment-926205356) discussed in #1142


tests/unit/loader/test_dataloader_backend.py
```diff
@@ -120,3 +122,33 @@ def test_dataloader_empty_error(datasets, engine, batch_size):
...
+        cat_names=cat_names,
+        cont_names=cont_names,
+        batch_size=batch_size,
+        label_names=label_name,
+        shuffle=False,
+    )
+
+    # Convert to iterators and then to DataFrames
+    df1 = _concat(list(data_loader._buff.itr))
+    df2 = _concat(list(data_loader.epochs(epochs)._buff.itr))
```
rjzamora, line 147: @jperez999 - Do you like this API any better? Note that `data_loader.epochs(n)` is used here to create a multi-epoch DataLoader object from the base (single-epoch) DataLoader.

jperez999: yeah this is better. It's allowed but not always set to a given number only during iteration.  I fell like I can get behind this.

jperez999 (approved):

PR #1152, opened 2021-09-29
Fix conda environment files

benfred:
Our conda environment files were failing to install, because it was overspecifying required
versions of transitive dependencies. For instance, we specified the version of pandas required
but that conflicts with the version that we pick up from cudf (likewise pyarrow etc).

Fix by removing transitive dependencies like pandas/pyarrow etc-  and instead rely on the
fact that these are dependencies of cudf. This means we won't have to keep updating version
numbers ourselves to keep in sync with cudf. Also remove development dependencies, which can
be picked up from requirements-dev.txt instead.

### DIFF
--- a/conda/environments/nvtabular_dev_cuda11.0.yml
+++ b/conda/environments/nvtabular_dev_cuda11.0.yml
@@ -5,54 +5,17 @@ channels:
   - conda-forge
   - defaults
 dependencies:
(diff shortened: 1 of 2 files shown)


bschifferer (approved):
I could install nvtabular with the 11.2 file. I couldnt test 11.0

PR #1161, opened 2021-10-05
decouple io from schema

jperez999:


nvtabular/columns/schema_io/schema_writer_base.py
```diff
@@ -0,0 +1,12 @@
+class SchemaWriter:
+    @classmethod
+    def _read_scheam(cls, schema_path):
```
karlhigley, line 3:
```suggestion
    def _read_schema(cls, schema_path):
```

nvtabular/columns/schema_io/schema_writer_base.py
```diff
@@ -0,0 +1,12 @@
+class SchemaWriter:
```
karlhigley, line 1: This seems like it might want to be an abstract base class in order to enforce that these methods are overwritten?

nvtabular/columns/schema_io/schema_writer_base.py
```diff
@@ -0,0 +1,12 @@
...
+    @classmethod
+    def _read_scheam(cls, schema_path):
+        raise NotImplementedError("Must have logic to read schema from file")
+
+    @classmethod
+    def write_schema(cls, schema, schema_path):
```
karlhigley, line 7: Since the class names already have `Schema` in them, we can probably drop `_schema` from these method names so they're just `write`,`load`, and `_read`

nvtabular/columns/schema_io/schema_writer_base.py
```diff
@@ -0,0 +1,12 @@
...
+        raise NotImplementedError("Must have logic to write schema to file")
+
+    @classmethod
+    def load_schema(cls, schema_path):
+        raise NotImplementedError("logic to create schema object from file")
```
karlhigley, line 12: This file needs a copyright/license notice

jperez999:
rerun tests

karlhigley (approved):

PR #1162, opened 2021-10-06
Groupby null

jperez999:
adding <NA> value count for categoricals


benfred (approved):

nvtabular/ops/categorify.py
```diff
@@ -1066,6 +1068,7 @@ def _groupby_to_disk(ddf, write_func, options: FitOptions):
     level_2_name = "level_2-" + token
     level_3_name = "level_3-" + token
     finalize_labels_name = options.stat_name + "-" + token
+    options.dataset_size = len(ddf)
```
rjzamora, line 1071: This call can be pretty expensive (see #1222). Perhaps we can add `"size"` to the default groupby aggregations here and use the existing Categorify tree reduction to calculate the dataset length (rather than adding entirely new graph construction/execution step here).

PR #1179, opened 2021-10-11
Updates cudf 21.10

albert17:


nvtabular/dispatch.py
```diff
@@ -30,7 +30,7 @@
     import dask_cudf
     import rmm
     from cudf.core.column import as_column, build_column
-    from cudf.utils.dtypes import is_list_dtype, is_string_dtype
+    from cudf.api.types import is_list_dtype, is_string_dtype
```
benfred, line 33: Can we make this backwards compatible for older versions of cudf ? Something like we do for hash_object_dispatch ?

https://github.com/NVIDIA-Merlin/NVTabular/blob/b94f6b49f2d02b4ac82aa8d1951b9b625c820648/nvtabular/dispatch.py#L42-L47

nvtabular/inference/triton/benchmarking_tools.py
```diff
@@ -58,7 +58,7 @@ def _convert_df_to_triton_json(df, input_dtypes=None):
     for col in input_dtypes:
         values = df[col].values_host
         # we need to fill None values in strings
-        if cudf.utils.dtypes.is_string_dtype(values.dtype):
+        if cudf.api.types.is_string_dtype(values.dtype):
```
benfred, line 61: Should we just switch to the `_is_string_dtype` wrapper in nvtabular/dispatch.py? (so handle differences between cudf versions, and also enable cpu?)

albert17:
@rjzamora merged your PR #1197 here

@benfred added cudf backwards compatibility

PR #1212, opened 2021-10-26
TensorFlow Sparse Features

bschifferer:
Added a custom TensorFlow layer to enable tf.sparse.SparseTensor with NVTabular dataloader
Added unit tests for it
Added an example notebook


marcromeyn:
We have a [similar layer](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/main/transformers4rec/tf/tabular/transformations.py#L30) in TF-rec that will move to Merlin-models soon. So I guess the question is whether we want some duplication for now that we'll later merge or whether we hold off on the PR and move it to Merlin-models directly?

cc @karlhigley @benfred @EvenOldridge

PR #1215, opened 2021-10-27
fixes for schema saving and writing

jperez999:
This PR  accomplishes the following

- Removes auto tagging CATEGORICAL from groupby op
- Fixes schema writing, introduces transformed schema property in dataset. Previously using dataset original schema to write out, now it uses workflow output schema after transform has completed.
- Adds the ability to merge to ColumnSchemas with same name.
- Enhanced addition of schemas to handle key collisions by merger the column schemas


marcromeyn:
> @jperez999 thanks for the fixes. @marcromeyn Could you please check the schema.pbtxt outputted from [this code](https://gist.github.com/rnyak/f5f50abcb92c987fdbcec88d6a50cf84) and see if it is fine for you? We can parse the `extra-metadata` field or not? Note that continuous features do not have min-max values, would that create a problem or not? Thanks.

Yeah, that would be an issue I think. We would need to know the list-lengths for continuous features as well. One thing that could help is if we could easily output a common tag for all group-by columns so that we could distinguish multiple different sequences in a dataset.

Next to this, are there tests as well to make sure the `int_domain` is saved to the schema?

karlhigley:
@jperez999 Not sure if this is the latest, but weird that merging the main branch resulted in a test failure that didn't happen before...

rnyak:
@jperez999 @karlhigley you can merge this. It works fine now at my end.

rnyak (approved):

PR #1252, opened 2021-11-16
Modularize the Triton inference model for NVT Workflows

karlhigley:
This creates a `WorkflowRunner` base class that contains the core code for executing Workflows in Triton, and then migrates code that's specific to Tensorflow, PyTorch, or HugeCTR into sub-classes of it. This concentrates framework-specific code from being scattered throughout one large `TritonPythonModel` into classes where you can see all of it at once.


nvtabular/inference/triton/model.py
```diff
@@ -24,306 +24,45 @@
...
-            )
+        model_kind = args["model_instance_kind"]
+        model_config = json.loads(args["model_config"])
+        output_model = model_config["parameters"]["output_model"]["string_value"]
+
+        if output_model == "hugectr":
+            runner_class = HugeCTRWorkflowRunner
+        elif output_model == "pytorch":
+            runner_class = PyTorchWorkflowRunner
+        elif output_model == "tensorflow":
```
karlhigley, line 60: I'm not entirely sure that `"tensorflow"` is the right value here. It might just be `"model"`?

nvtabular/inference/triton/model.py
```diff
@@ -24,306 +24,45 @@
...
+            runner_class = TensorflowWorkflowRunner
         else:
-            output_tensors.append(Tensor("DES", np.array([[]], np.float32)))
+            runner_class = TensorflowWorkflowRunner
```
karlhigley, line 63: Also not sure if Tensorflow is the right default here, but the tests weren't passing without this.

karlhigley: (The diff is misleading due to large changes in this file)

benfred: I'm not sure what the best default here is either - but I do think the previous behaviour was to default to TF format

benfred (approved):
Thanks for cleaning this up! The previous code was a bit of a mess, and I like that you've organized this to be much more maintainable.

Issue #1235, opened 2021-11-09
[BUG] Cannot execute Workflow.fit on full Criteo Dataset located on a remote storage (Google Cloud Storage)


leiterenato:
**Describe the bug**
I cannot run Workflow.fit on a full Criteo dataset located on a remote storage (Google Cloud Storage).
Here is a gist with the error message:
https://gist.github.com/leiterenato/678636807fe168f5709a35b1771fe5d7

Dataset location: 'gs://workshop-datasets/criteo-parque/' (24 files ~12GB each).

**Steps/Code to reproduce bug**
1) Create a cluster:
```python
device_size = device_mem_size()
device_limit = int(device_limit_frac * device_size)
device_pool_size = int(device_pool_frac * device_size)
rmm_pool_size = (device_pool_size // 256) * 256

cluster = LocalCUDACluster(
    n_workers=n_workers,
    device_memory_limit=device_limit,
    rmm_pool_size=rmm_pool_size
)
```

I am using the n_workers equal to the number of GPUs, in this case 4.

2) Create the dataset definition:
fs = fsspec.filesystem('gs')
data_path = 'gs://workshop-datasets/criteo-parque/'
file_list = fs.glob(
    os.path.join(data_path, '*.parquet')
)
file_list = [os.path.join('gs://', i) for i in file_list]

dataset = nvt.Dataset(
    file_list,
    engine="parquet", 
    part_size=int(part_mem_frac * device_mem_size()),
    client=client
)

3) Create the workflow:
```python
cont_names = ["I" + str(x) for x in range(1, 14)]
cat_names = ["C" + str(x) for x in range(1, 27)]

num_buckets = 10000000
categorify_op = Categorify(max_size=num_buckets)
cat_features = cat_names >> categorify_op
cont_features = cont_names >> FillMissing() >> Clip(min_value=0) >> Normalize()
features = cat_features + cont_features + ['label']

# Create and save workflow
workflow = nvt.Workflow(features, client)
```

4) Execute FIT:
```python
workflow.fit(dataset)
````

**Expected behavior**
I expect to fit the dataset.

**Environment details:**
 - Google Cloud
 - Using the NGC merlin-training image 21.11
 - 1 x N1 instance with 96 vCPUs and 360GB RAM
 - 4 x T4s

mengdong:
with cudf 21.12 and rmm 21.10, here is what I have: 

In [26]: %cpaste
[...]

Issue #1249, opened 2021-11-16
[FEA] Add DistributedClient API and use global client objects in Workflow and Dataset


rjzamora:
**Is your feature request related to a problem? Please describe.**
In order to perform distributed ETL and/or spill data between device/host/disk, NVTabular currently requires the user to provide a Dask-Distributed (or Dask-CUDA) cluster.  Furthermore, the `nvt.Workflow` (and sometimes `nvt.Dataset`) need to be defined with an explicit `client=` argument in order for the distributed cluster to be used.  Although I feel strongly that it would be dangerous and messy for NVTabular to automatically spin up a distributed cluster by default, I do suspect that the user experience could be much better. 

**Describe the solution you'd like**
To improve the user experience of distributed ETL with NVTabular, I propose:

1. Simple `LocalCluster`/`LocalCUDACluster`-wrapper APIs be added to NVTabular so that users can enable multi-GPU processing and/or spilling without interacting with distributed/dask_cuda. I am not yet sure of the ideal API to expose in NVTabular, but perhaps something like `DistributedClient` (wrapping `distributed.Client`). This API could be used to automatically generate a local cluster (if the address of an existing cluster is not provided), and we could add a `cpu=False` kwarg to toggle between gpu and cpu mode.
2. [**DONE** in #1318] Automatically detect and **use** an existing Dask client object. NVTabular already [checks for a global dask client](https://github.com/NVIDIA-Merlin/NVTabular/blob/34d01d7e6090d6029ac40010ed79e1558f18759c/nvtabular/workflow/workflow.py#L88) in both `Workflow ` and `Dataset`.  However, these checks result in a UserWarning whenever a global client is detected.  Instead of warning the user, I propose that NVTabular automatically **use** the client object when it is detected (which is actually the defult behavior in `dask.dataframe` and `dask.array` anyway).

**Describe alternatives you've considered**
The alternative is to continue requiring the user to deploy Dask clusters/clients with `distributed`/`dask_cuda`, and require them to include an explicit `client` argument when defining an `Workflow` objects.

albert17:
I think this is a great idea, and it would be very valuable. All our examples/notebooks would be automatically running in multi-gpu and customers would be able to use multi-gpu with minimum effort and knowledge.

Happy to work into this, unless Rick wants to tackle it.

rjzamora:
>Happy to work into this, unless Rick wants to tackle it.

[...]

PR #1259, opened 2021-11-19
Reduce device-memory footprint in Categorify fit

rjzamora:
While (deeply) investingating #1235, I discovered that we are using unneccassary device memory to write out the unique values and counts for each categorical column.

- The first problem is that we are failing to pass `ignore_index=True` in the intial sort used to ensure the Null element is first.  This means that the original light-weight RangeIndex is materialized into a full int64 column (adding an unnecessary 64 bytes for every untique value).  The changes in this PR address this bug.
- The second problem is that we are always using `int64` for the "_size" column (used to store the count for each category). For the case of `on_host=True`, we *know* that a low memory footprint is a priority.  Therefore, the changes in this PR sacrifice an the time needed to calculate the maximum value of the "_size" column, and then down-cast the dtype before memory-intensive `sort_values` and `nlargest` operations.

Addresses #1235 (May fully close the issue, but I'll wait for direct confirmation before doing that)


nvtabular/ops/categorify.py
```diff
@@ -954,16 +977,69 @@ def _write_uniques(dfs, base_path, col_selector: ColumnSelector, options: FitOpt
...
+
+    # Check if we should warn user that this Column is likely
+    # to cause memory-pressure issues
+    _df_size = df.memory_usage(deep=True, index=True).sum()
+    if (_df_size > options.cardinality_memory_limit) if options.cardinality_memory_limit else False:
+        warnings.warn(
+            f"Category DataFrame (with columns: {df.columns}) is {_df_size} "
+            f"bytes in size. This is large compared to the suggested "
+            f"upper limit of {options.cardinality_memory_limit} bytes!"
+        )
```
rjzamora, line 1034: I'm very open to feedback on this `UserWarning`. This PR introduces the `cardinality_memory_limit` argument as a mechanism to check if we should take extra steps to reduce memory usage.  However, I also assume it would be useful to warn the user if/when we are unable to keep the memory usage below this number (if the code eventually hits an OOM error, this warning should be a clear hint that a high-cardinality column is likely causing problems).

nvtabular/ops/categorify.py
```diff
@@ -185,6 +186,11 @@ class Categorify(StatOperator):
         and our new translated dataframe entry will now be [[17], [17, 20], [19, 18], [18]].
         This parameter is useful to reserve an initial segment of non-negative translated integers
         for special user-defined values.
+    cardinality_memory_limit: int or str, default None
```
rjzamora, line 189: Open to suggestions on the best name here. In this PR, I want to use this attribute as a guide for deciding whether or not to perform memory-optimization logic. However, in the future, we *should* be able to use it to choose better defaults across the board.

EvenOldridge: That seems like a reasonable naming.

nvtabular/ops/categorify.py
```diff
@@ -339,6 +349,15 @@ def fit(self, col_selector: ColumnSelector, ddf: dd.DataFrame):
             if c not in cols_with_vocabs
         ]
 
+        # Define a rough row-count at which we are likely to
+        # start hitting memory-pressure issues that cannot
+        # be accommodated with smaller partition sizes.
+        # By default, we estimate a "problematic" cardinality
+        # to be one that consumes >12.5% of the total memory.
+        self.cardinality_memory_limit = parse_bytes(
+            self.cardinality_memory_limit or int(device_mem_size(kind="total", cpu=_cpu) * 0.125)
```
EvenOldridge, line 358: Does this 12.5% relate to the 10% above?  Or is this specific to CPU memory?

nvtabular/ops/categorify.py
EvenOldridge: It might make sense to add in the defaults are calculated (12.5% of host memory?)

 warnings.warn(
            f"Category DataFrame (with columns: {df.columns}) is {_df_size} "
            f"bytes in size. This is large compared to the suggested "
            f"upper limit of {options.cardinality_memory_limit} bytes!"
            f"(12.5% of host memory by default)"
        )

EvenOldridge (approved):
A few very minor changes, but I think this is otherwise good.

nvtabular/ops/categorify.py
```diff
@@ -185,6 +186,11 @@ class Categorify(StatOperator):
         and our new translated dataframe entry will now be [[17], [17, 20], [19, 18], [18]].
         This parameter is useful to reserve an initial segment of non-negative translated integers
         for special user-defined values.
+    cardinality_memory_limit: int or str, default None
+        Upper limit on the "allowed" memory usage of the internal DataFrame and Table objects used
+        to store unique categories. By default, this limit is assumed to be 10% of the total memory.
```
rjzamora, line 191:
```suggestion
        to store unique categories. By default, this limit is assumed to be 12.5% of the total memory.
```

nvtabular/ops/categorify.py
```diff
@@ -954,16 +977,69 @@ def _write_uniques(dfs, base_path, col_selector: ColumnSelector, options: FitOpt
...
         df = _concat(dfs, ignore_index=True)
+
+    # Check if we should warn user that this Column is likely
+    # to cause memory-pressure issues
+    _df_size = df.memory_usage(deep=True, index=True).sum()
+    if (_df_size > options.cardinality_memory_limit) if options.cardinality_memory_limit else False:
+        warnings.warn(
+            f"Category DataFrame (with columns: {df.columns}) is {_df_size} "
+            f"bytes in size. This is large compared to the suggested "
+            f"upper limit of {options.cardinality_memory_limit} bytes!"
```
rjzamora, line 1033:
```suggestion
            f"upper limit of {options.cardinality_memory_limit} bytes!"
            f"(12.5% of the total memory by default)"
```

nvtabular/ops/categorify.py
rjzamora: Oops. These should both be 12.5% (just pushed a correction).

Also, this is compared to 12.5% of the total host memory where the worker is running, or 12.5% of the total device memory for a single GPU, depending on whether we are running in cpu or gpu mode.

I am actually pretty unsatisfied with this answer, because this means the comparison is quite different for gpu and cpu-backed data. However, I am not sure if it is possible for us to query to query the total host memory allocated to the current Dask worker. We may be able to "fix" this in `device_mem_size` by searching for a global Dask client, but we probably need to make some of the changes suggested in #1249 first.

rjzamora:
I'm going to merge this now to get the categorify memory footprint down. However, I suspect we will want to follow up with changes to: (1) Use int32 instead of int64 by default for encodings, and (2) align the meaning of `cardinality_memory_limit` for gpu- and cpu- backed data.

PR #1265, opened 2021-11-22
Extract an abstract `graph` package from NVT Workflows

karlhigley:
This extracts a generic set of graph construction abstractions from the existing NVT Workflow code, in hopes that they can be reused to build graphs of Triton components too.

The one (breaking) change to NVT functionality this requires is that "naked" lambdas are no longer possible to `>>` onto the graph; they first have to wrapped with `LambdaOp`.


karlhigley:
rerun tests

PR #1272, opened 2021-11-24
dataset duck typing for dataloader

jperez999:
removing knowledge of dataset and everything else outside the Tags and dispatch from dataloader package.


rjzamora:
Not that I have any problem with this kind of change, but is there a motivating issue or explanation of the rational here?

karlhigley:
@rjzamora I think this is related to wanting to move the dataloaders into the new models library without that library needing to depend on NVTabular. From talking to @jperez999 earlier, it sounds like allowing the dataloaders to operate on a plain ddf makes it easier to cleanly separate datasets (in NVT) from dataloaders (in the models library in a way that lets them be used either together or separately.

rjzamora:
Sounds good @karlhigley !  This does make sense to me. However, since I occasionally find myself tracking down performance regressions in NVTabular, I like there to be a linked issue or motivation described in any PR touching the IO interface.  Your last comment seems sufficient to me :)

jperez999:
rerun tests

tests/unit/loader/test_dataloader_backend.py
```diff
@@ -122,11 +122,22 @@ def test_dataloader_empty_error(datasets, engine, batch_size):
     )
 
 
+class DatasetStub:
+    def __init__(self, dataset, schema):
+        self.data = dataset
+        self.schema = schema
+
+
```
karlhigley, line 130: It doesn't look like this actually gets used in this file:
```suggestion
```

nvtabular/loader/torch.py
```diff
@@ -87,6 +87,8 @@ def __init__(
         sparse_max=None,
         sparse_as_dense=False,
     ):
+
+        # dataset = nvt.dispatch.validate_ds(dataset)
```
karlhigley, line 91: Stray comment here:
```suggestion

```

nvtabular/loader/tensorflow.py
```diff
@@ -33,12 +32,21 @@
...
-def _validate_dataset(paths_or_dataset, batch_size, buffer_size, engine, reader_kwargs):
+dd_engine = {
+    "parquet": dd.read_parquet,
+    "csv": dd.read_csv,
+    "df": dd.DataFrame,
+}
+
+
+def _validate_dataset(
+    paths_or_dataset, batch_size, buffer_size, engine, reader_kwargs, dataclass=None
```
karlhigley, line 43: I think this extra parameter came from an intermediate version of the code, but doesn't look like it gets used here:
```suggestion
    paths_or_dataset, batch_size, buffer_size, engine, reader_kwargs
```

karlhigley (approved):

jperez999:
rerun tests

Issue #1282, opened 2021-11-30
[BUG] fitting not working correctly


jperez999:
The following example does not fit correctly:
```
@pytest.mark.parametrize("engine", ["parquet"])
def test_workflow_fit_op_rename(tmpdir, dataset, engine):
    # NVT
    schema =  dataset.schema
    for name in schema.column_names:
        dataset.schema.column_schemas[name] = dataset.schema.column_schemas[name].with_tags([nvt.tags.Tags.USER])
    selector = nvt.ColumnSelector(tags=[nvt.tags.Tags.USER])


    workflow_ops_1 = selector >> nvt.ops.Rename(postfix="_1")
    workflow_1 = nvt.Workflow(workflow_ops_1)
    workflow_1.fit(dataset)
    workflow_1.save(str(tmpdir / "one"))
    new_dataset = workflow_1.transform(dataset).to_ddf().compute()

    assert len(new_dataset.columns) > 0
    assert all("_1" in col for col in new_dataset.columns)
    ```

You end up failing on the fit looking for timestamp column.

PR #1285, opened 2021-12-01
Fix resolution of tags to column names when executing `Workflows`

jperez999:
workflow fix for #1282


karlhigley (commented):
I don't think the perms on the pre-commit-config.yaml should change, but it otherwise LGTM

karlhigley (approved):

PR #1292, opened 2021-12-02
Apply NVT graph API/DSL to building Triton ensembles

jperez999:
This PR applies the new graph package to provide an API for building Triton ensembles. For the first pass, it only supports the same ensembles that NVT already supported (i.e. request -> `Workflow` -> DL framework model -> response), and only supports Tensorflow (due to issues retrieving inputs/outputs from the saved model in other frameworks.)

Additionally, it creates a framework for chaining together multiple Python inference operators, storing them in a single Triton config, and running them in a single Triton model. This doesn't yet support `Workflow`s (which use more of the dataframe API's surface area than we wanted to implement in our virtual `InferenceDataframe` class), but should be usable for implementing operators that do things like filtering, unrolling user features, etc


nvtabular/inference/graph/ops/workflow.py
```diff
@@ -0,0 +1,101 @@
...
+    def __init__(
+        self, workflow, name=None, sparse_max=None, max_batch_size=None, label_columns=None
+    ):
+        self.workflow = workflow
+        self.sparse_max = sparse_max or {}
+        self.name = name or self.__class__.__name__
+        self.max_batch_size = max_batch_size
+        self.label_columns = label_columns or []
+
+    @classmethod
```
karlhigley, line 41: Same issue with `transform` and `from_config` here: the code is correct as written, but we can probably do without it if we use the class hierarchy a bit more cleverly.

nvtabular/inference/graph/tesnor_df.py
```diff
@@ -0,0 +1,102 @@
...
+        results["values"] = self.values
+        if self._offsets:
+            results["offsets"] = self._offsets
+        return str(results)
+
+    def append(self, values, offsets=None):
+        pass
+
+
+class TensorDataFrame:
```
karlhigley, line 71: Doesn't look like we're using this implementation in this PR. Worth saving for sure, but probably doesn't need to be included here

nvtabular/inference/triton/ensemble.py
```diff
@@ -83,6 +83,7 @@ def export_tensorflow_ensemble(
             )
             workflow.output_dtypes[column.name] = tf_dtype
 
+    # Node 2
```
karlhigley, line 86:
```suggestion
```

nvtabular/inference/triton/ensemble.py
```diff
@@ -64,7 +64,7 @@ def export_tensorflow_ensemble(
     nvtabular_backend: "python" or "nvtabular"
         The backend that will be used for inference in Triton.
     """
-
+    # remove
```
karlhigley, line 67:
```suggestion
```

tests/unit/inference/test_export.py
```diff
@@ -0,0 +1,105 @@
...
+            {},
+            "test_name",
+            tmpdir,
+            [],
+        )
+
+    # assert os.path.exists(os.path.join(repo, "config.pbtxt"))
+    tri_df = df.iloc[:10]
+    tri_df = tri_df[["x", "y", "id", "name-cat", "name-string"]]
+    response = _run_ensemble_on_tritonserver(str(tmpdir), ["output"], tri_df, "test_name")
```
karlhigley, line 58: Worth noting that this is the first test in NVT that tests a whole ensemble (including the TF model) and not just the `Workflow`

nvtabular/inference/graph/ensemble.py
```diff
@@ -0,0 +1,96 @@
+#
```
karlhigley, line 1: Seems like we might want to import some things in the `nvtabular.inference.graph` init file for convenience. At least `Ensemble` and `Node` probably?

karlhigley (commented):
Looking good overall 👍🏻 

(Left some comments re: cleanup for us to do)

nvtabular/inference/graph/ops/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# alias submodules here to avoid breaking everything with moving to submodules
+# flake8: noqa
```
karlhigley, line 18:
```suggestion
# alias submodules here to avoid breaking everything with moving to submodules
# flake8: noqa
```

jperez999: we need the flake8 noqa because of the method for imports... look at CI failures below I am going to revert this.

karlhigley (approved):

benfred (approved):
This looks like a good first step ! I'm excited to see this come along.

There are a couple things that I'd like to see added here (docstrings for important classes like Graph / Ensemble,  updating the examples to use this API, deprecating the old API etc), but I think we can handle that in a future PR

karlhigley:
Agreed on all of those suggestions! We're building out the new thing without removing the old thing for now, and will do the deprecation once we have a full set of functionality for people to migrate over to.

PR #1303, opened 2021-12-07
Uninstall NVT (removing versions from PyPI) before installing NVT in CI

karlhigley:

### DIFF
--- a/ci/build_and_test.sh
+++ b/ci/build_and_test.sh
@@ -8,7 +8,7 @@ cd $nvt_directory
 
 echo "Installing NVTabular"
 python -m pip install --user --upgrade pip setuptools wheel pybind11 numpy==1.20.3
-python -m pip uninstall nvtabular
+python -m pip uninstall nvtabular -y
 python setup.py develop --user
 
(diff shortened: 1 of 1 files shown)


albert17:
@jperez999 @karlhigley Could this be a problem if it is run while TF4Rec tests are running?

karlhigley:
I think it should be okay most of the time, because the imports get loaded at the beginning and after that the files on disk don't matter. We might get an occasional flaky test run if things line up badly between NVT and Transformers4Rec, but I think we can rerun the tests in that case.

jperez999 (approved):

jperez999:
rerun tests

ci/build_and_test.sh
```diff
@@ -8,6 +8,7 @@ cd $nvt_directory
 
 echo "Installing NVTabular"
 python -m pip install --user --upgrade pip setuptools wheel pybind11 numpy==1.20.3
+python -m pip uninstall nvtabular
```
jperez999, line 11: can you add -y ?

PR #1308, opened 2021-12-08
Set the priority and date added fields for new issues.

benfred:
This change sets the priority field in the backlog for new issues to 'needs triage'.
This will let us create a view in the project board of new issues.

Also this adds a 'date added' field and automatically sets it for new issues.
While github has dates associated with issues already, there is no way for us to expose
these as fields on the project boards (see https://github.com/github/feedback/discussions/7280 )
Automatically populating a date field ourselves lets us work around this limitation.


bschifferer (approved):

benfred:
rerun tests

PR #1309, opened 2021-12-09
Label issues not created by nvidia-merlin members

benfred:
When adding issues to the backlog, label the issue as being from a 'External' user in the
impact field if the user that created the issue isn't a member of nvidia-merlin.


EvenOldridge (approved):
This is great.  Thanks for implementing.

benfred:
Rerun tests

PR #1317, opened 2021-12-13
removed column_types.json from nvtabular

jperez999:


nvtabular/inference/triton/ensemble.py
```diff
@@ -55,16 +57,16 @@ def export_tensorflow_ensemble(
         The base name of the various triton models
     model_path:
         The root path to write out files to
-    label_columns:
-        Labels in the dataset (will be removed from the dataset)
```
benfred, line 59: Removing the label_columns as a parameter means you have to update all the criteo, rossmann, and movielens notebooks to reflect this change. This includes not just removing the label_columns from the export_tensorflow_ensemble code, but also updating the workflow to add the 'Target' tag to the label column instead.

We might want to keep this parameter, add a deprecated warning if its set - and then if its not set grab from the schema like you are doing here?

nvtabular/inference/triton/ensemble.py
```diff
@@ -178,10 +179,7 @@ def export_hugectr_ensemble(
     hugectr_params,
     name,
     output_path,
-    label_columns,
```
benfred, line 181: Similar to the TF call - you will need to update the hugectr movielens/criteo inference notebooks here too, not just to remove the passed in label_columns but also the cats/conts

benfred (changes requested):
I like that we're getting rid of this file!  

I don't think we can merge this yet though, left some comments below about why

nvtabular/inference/triton/ensemble.py
karlhigley: Indeed, notebook changes are required. Keeping this around for backwards compatibility seems reasonable to me though. Would you apply the same logic to `cats` and `conts`, or are you thinking specifically about labels here?

nvtabular/inference/triton/ensemble.py
karlhigley: Yep - looks like the tests catch these issues

nvtabular/inference/triton/ensemble.py
karlhigley: We actually ended up _adding_ `cats` and `conts` to `export_tensorflow_ensemble` to make everything work consistently. 🤷🏻

nvtabular/inference/triton/ensemble.py
```diff
@@ -30,14 +32,17 @@
...
 
 
 def export_tensorflow_ensemble(
     model,
     workflow,
     name,
     model_path,
-    label_columns,
+    cats=None,
+    conts=None,
```
benfred, line 44: The movielens notebook passes the label_column by position instead of by name

```"export_tensorflow_ensemble(model, workflow, MODEL_NAME_ENSEMBLE, MODEL_PATH, [\"rating\"])```

( https://github.com/NVIDIA-Merlin/NVTabular/blame/main/examples/getting-started-movielens/03-Training-with-TF.ipynb#L706 ).

I think this will still break that notebook, or at least cause unexpected results with the `rating` column being treated as a categorical.

Do we need to insert the cats/conts here?

jperez999:
rerun tests

nvtabular/inference/triton/ensemble.py
jperez999: I just used the keyword arg in the call (label_columns) and now everything is working again 👍  good catch though.

benfred (approved):

PR #1318, opened 2021-12-14
Deprecate client from Dataset, Workflow, and DatasetInspector

rjzamora:
Partially addresses #1249

This PR is intended to improve user experience by removing the need for users to specify a `client` object when one is already defined in the global python context.

TODO:
- [x] Remove `client` usage from examples and benchmarks


rjzamora:
@bschifferer @rnyak - It would be good to get your thoughts (since you are both developers and "power users").

After this change, the user will still be responsible for deploying a dask cluster if/when they want to perform distributed (multi-GPU) ETL.  However, the user will no longer be responsible for specifying `client=` to any NVTabular API (i.e. `Workflow` or `Dataset`). By default, NVTabular will find/use the dask client defined in the global python context. To disable this "automatic-client-detection" behavior, the user can call `nvt.utils.set_dask_client(None)`.

cc @benfred and @albert17 as well :)

bschifferer:
I think it will be easier for the user. Does it still work, when the dask_cluster is on another node?

Example:
- we have one node, which runs the jupyter server
- we have 1-n nodes, which runs the dask-workers

Is NVTabular able to auto-connect in this setup?

rjzamora:
>Is NVTabular able to auto-connect in this setup?

This PR does not change anything about what kinds of clusters NVT can connect to, or how those cluster should be deployed. If the user wants a multi-node dask cluster with a scheduler on a different node than the jupyter server, they still need to do this with a Dask-based api, and the user will still need to generate a `client` object on the jupyter server (to connect to the cluster).  The only difference is that the user will no longer need to specify the client when they define a `Workflow` or `Dataset` object.

examples/multi-gpu-movielens/01-03-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-PyTorch.ipynb
```diff
@@ -369,7 +369,7 @@
     "ratings = nvt.ColumnSelector([\"rating\"]) >> nvt.ops.LambdaOp(lambda col: (col > 3).astype(\"int8\"))\n",
     "output = cat_features + ratings\n",
     "# USE client in NVTabular workflow\n",
```
rjzamora, line 371:
```suggestion
```

examples/multi-gpu-movielens/01-03-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-TensorFlow.ipynb
```diff
@@ -374,7 +374,7 @@
     "ratings = nvt.ColumnSelector([\"rating\"]) >> nvt.ops.LambdaOp(lambda col: (col > 3).astype(\"int8\"))\n",
     "output = cat_features + ratings\n",
     "# USE client in NVTabular workflow\n",
```
rjzamora, line 376:
```suggestion
```

nvtabular/io/dataset.py
```diff
@@ -220,7 +220,7 @@ def __init__(
         part_mem_fraction=None,
         storage_options=None,
         dtypes=None,
-        client=None,
+        client="auto",
```
rjzamora, line 223:
```suggestion
```

rjzamora:
rerun tests

rjzamora:
CI failure is in `test_export_run_ensemble_triton` (added in #1292).  This seems unrelated to the current PR (probably?). cc @jperez999 (in case you have thoughts)

rjzamora:
@benfred @karlhigley @jperez999 - **Just a note**:  No need to get this in for the nearest/current 22.01 release. However, we probably want to get it in soon after that to make sure there is time to find and fix/simplify old code with explicit `client` passing.

examples/multi-gpu-toy-example/multi-gpu_dask.ipynb
```diff
@@ -591,9 +591,7 @@
     "\n",
     "### Step 5: Define our NVTabular Workflow\n",
     "\n",
-    "Now that we have our Dask cluster up and running, we can use the NVTabular API as usual. For multi-GPU execution, the only requirement is that you specify a `client` when you initialize the NVTabular Workflow.\n",
-    "\n",
-    "**IMPORTANT**: You must pass `client` to `nvt.Workflow` to enable Multi-GPU execution!"
+    "Now that we have our Dask cluster up and running, we can use the NVTabular API as usual. For NVTabular versions newer than `0.8.0`, the global `client` (created above) will be used automatically for multi-GPU (or CPU) execution."
```
rjzamora, line 594:
```suggestion
    "Now that we have our Dask cluster up and running, we can use the NVTabular API as usual. For NVTabular versions newer than `0.9.0`, the global `client` (created above) will be used automatically for multi-GPU (or CPU) execution."
```

jperez999 (approved):

PR #1321, opened 2021-12-14
MultiGPU TensorFlow uses Keras

bschifferer:
Updated the multi-GPU TensorFlow example with TF.Keras as this is easier to use


karlhigley (approved):

karlhigley:
rerun tests

karlhigley:
@bschifferer I liked this change, but not sure we ever got it to a point where we could merge it. Should we keep this open and revive it or close and re-submit a version built on `main`?

PR #1348, opened 2022-01-05
Create an explicit mapping between Operator input and output columns

jperez999:


nvtabular/graph/base_operator.py
```diff
@@ -48,6 +48,8 @@ def compute_selector(
         parents_selector: ColumnSelector,
         dependencies_selector: ColumnSelector,
     ) -> ColumnSelector:
+        self._validate_matching_cols(input_schema, selector, "computing input selector")
```
karlhigley, line 51: Could we use the method name instead of hard coded strings here? 🤔

nvtabular/graph/graph.py
```diff
@@ -80,6 +57,11 @@ def fit_schema(self, root_schema: Schema) -> "Graph":
 
         return self
 
+    def update_properties(self):
+        for node in postorder_iter_nodes(self.output_node):
+            if node.parents:
+                node.op.update_properties()
```
karlhigley, line 63: I don't think `update_properties` still exists, so we should replace this with the newer version.

nvtabular/inference/triton/ensemble.py
```diff
@@ -702,7 +702,8 @@ def _remove_columns(workflow, to_remove):
                     if column in subgroup._names:
                         subgroup._names.remove(column)
 
-    return workflow.fit_schema(new_schema)
+    workflow.fit_schema(new_schema)
+    return workflow
```
karlhigley, line 706: This looks like a debugging change, but doesn't look like we need it any more.

nvtabular/ops/categorify.py
```diff
@@ -1388,22 +1390,6 @@ def _read_groupby_stat_df(path, name, cat_cache, read_pq_func):
     return read_pq_func(path)
 
 
-def _get_multicolumn_names(col_selector, df_columns, name_sep):
```
karlhigley, line 1391: Nice that we can remove this now that we have `column_mapping`. 👍🏻

karlhigley (approved):

PR #1351, opened 2022-01-07
Allows int64 hugectr keysetfile

albert17:


albert17:
@jershi425 Initially, we can just write all values using 64 bits.

Do you know the expected dtype before reading the file?

jershi425:
I think the expected dtype should be consistent with the `dtypes`  specified in `workflow.transform().to_parquet()`.

albert17:
@jershi425 I have pushed an update. Please, let me know what you think now

jershi425:
@albert17 Thank you! I will try it out and let you know.

PR #1354, opened 2022-01-10
schema metadata propagation

jperez999:
This pr will focus on creating a more controlled method for propagating changes to schemas.

### DIFF
--- a/nvtabular/graph/graph.py
+++ b/nvtabular/graph/graph.py
@@ -30,15 +30,23 @@ class Graph:
 
     @property
     def input_dtypes(self):
-        return {
-            name: col_schema.dtype for name, col_schema in self.input_schema.column_schemas.items()
-        }
+        if self.input_schema:
+            return {
+                name: col_schema.dtype
+                for name, col_schema in self.input_schema.column_schemas.items()
+            }
+        else:
+            return {}
 
     @property
     def output_dtypes(self):
-        return {
-            name: col_schema.dtype for name, col_schema in self.output_schema.column_schemas.items()
-        }
+        if self.output_schema:
+            return {
+                name: col_schema.dtype
+                for name, col_schema in self.output_schema.column_schemas.items()
+            }
+        else:
+            return {}
 
     def fit_schema(self, root_schema: Schema, preserve_dtypes=False) -> "Graph":
         nodes = list(postorder_iter_nodes(self.output_node))
--- a/nvtabular/graph/graph.py
+++ b/nvtabular/graph/graph.py
@@ -15,7 +15,7 @@
 #
 import logging
 
-from nvtabular.graph.node import Node, iter_nodes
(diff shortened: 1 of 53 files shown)


jperez999:
rerun tests

nvtabular/graph/graph.py
```diff
@@ -55,8 +79,6 @@ def fit_schema(self, root_schema: Schema) -> "Graph":
         )
         self.output_schema = self.output_node.output_schema
 
-        return self
```
karlhigley, line 58: I think this was removed accidentally, but `fit_schema` should probably still `return self`

nvtabular/workflow/workflow.py
```diff
@@ -460,6 +476,19 @@ def _transform_partition(root_df, workflow_nodes, additional_columns=None):
...
                 selection = node.input_columns.resolve(node.input_schema)
                 output_df = node.op.transform(selection, input_df)
+
+                # Update or validate output_df dtypes
+                for col_name, col_schema in node.output_schema.column_schemas.items():
+                    output_schema = col_schema.with_dtype(output_df[col_name].dtype)
+
+                    if capture_dtypes:
+                        node.output_schema.column_schemas[col_name] = output_schema
+                #     else:
```
karlhigley, line 486: We should still try this, but it can probably be removed from this PR for the time being.

tests/unit/ops/test_lambda.py
```diff
@@ -153,3 +155,94 @@ def test_lambdaop_misalign(cpu):
...
+    label_feature = (
+        label >> ops.LambdaOp(lambda col: col.astype(np.float32)) >> ops.Rename(postfix="_1st")
+    )
+    b_labels = (
+        label_feature["c_1st"]
+        >> ops.LambdaOp(lambda col: col.astype(np.float64))
+        >> ops.Rename(postfix="_2nd")
+    )
+
+    # workflow = nvt.Workflow(label_feature + b_labels)
```
karlhigley, line 242: I think we can either remove this commented line or replace the line below with it

tests/unit/ops/test_lambda.py
```diff
@@ -153,3 +155,94 @@ def test_lambdaop_misalign(cpu):
...
+    workflow = nvt.Workflow(label_feature)
+    workflow.fit(dataset)
+
+    output_schema = workflow.output_node.output_schema
+
+    assert output_schema["c_renamed"].dtype == dtype
+
+
+@pytest.mark.parametrize("cpu", _CPU)
+def test_lambdaop_blah(cpu):
```
karlhigley, line 219: This test should get a better name 😄

karlhigley (approved):
LGTM once the (minor) comments are addressed

nvtabular/graph/graph.py
jperez999: fit schema still has return self. Its just up further in the code.

PR #1361, opened 2022-01-14
[Experimental] Add ModelEncode Operator

rjzamora:
This PR adds an experimental `ModelEncode` Operator, which is designed to apply an arbitrary `model` to your data, and add one or more new columns with the model output (prediction).  The new test added here (`test_model_encode`) does not use a real tf/pt-based `Dataloader` and model, but it is absolutely my intention to support such a use case.

I am not completely sure if it makes sense to add this as an operator, but the `transform` logic is relatively simple (so it seemed reasonable to write it up).  Feedback is very welcome.

cc @marcromeyn


nvtabular/ops/model_predict.py
```diff
@@ -0,0 +1,100 @@
...
+            [predict_func(self.model, batch) for batch in iterator_func(df)]
+        )
+
+        # Return result
+        return new_df
+
+    def column_mapping(self, col_selector):
+        column_mapping = super().column_mapping(col_selector)
+        column_mapping[self.column_name] = col_selector.names
+        return column_mapping
```
rjzamora, line 98: Does this seem reasonable @karlhigley ?

karlhigley: Yeah, looks perfect 👍🏻

karlhigley: By default, this will pull the dtype/tags/properties from the first input column, which may or may not be what you want. If you have a model, you can probably pull the dtype of the output from the model and return it from `def output_dtype` though.

karlhigley: (If the input columns aren't also returned as outputs, you can drop the `super()` call.)

rjzamora: Right now, this Op passes through the input columns (unchanged), and adds a new column for the column prediction. Does this mean that I need to define something to specify the dtype of the new column?

Also, I just realized that it probably makes sense to allow the user to optionally drop some (or all) of the input columns.

rjzamora:
After reviewing the initial `ModelPredict` implementation here, I wonder if it makes sense to revise the Op to (1) be called `ModelEncode`, (2) allow the Model to add an arbitrary number of columns (rather than just one).  It feels like we want an Op that allows the user to encode a collection of columns into one or more new columns using a predefined "model".

I also feel that we need to add a simple test case to use tensorflow and/or pytorch for the model definition.

karlhigley:
@rjzamora @marcromeyn @benfred Is my understanding correct that this PR has been superseded by `merlin-core` and the use of `merlin.io.Dataset` in `merlin-models` to batch predict / bulk export item vectors?

karlhigley:
Closing since the PR is stale and I haven't heard anything, but open to re-opening if this is still a thing we want to do

PR #1362, opened 2022-01-14
Raise an error when the actual dtype produced by an operator doesn't match the schema

jperez999:
This adds a very helpful dtypes guard to transform_partition. Which ensures robustness by verifying that all column schema dtypes correspond to the dtypes of the output_df for each op transform.


tests/unit/ops/test_ops_schema.py
```diff
@@ -51,27 +53,28 @@ def test_schema_out(tags, properties, selection, op):
...
+                expected_tags = op._compute_tags(
                     ColumnSchema(col_name), Schema([input_schema.column_schemas[col_name]])
                 ).tags
-                output_properties = op._compute_properties(
+
+                expected_properties = op._compute_properties(
                     ColumnSchema(col_name), Schema([input_schema.column_schemas[col_name]])
                 ).properties
+
+                # should not be exactly the same name, having gone through operator
```
karlhigley, line 70: I think we can remove the comment above

karlhigley (approved):

PR #1376, opened 2022-01-31
Fixes hash_crossed with cudf 21.12

albert17:

### DIFF
--- a/nvtabular/ops/hashed_cross.py
+++ b/nvtabular/ops/hashed_cross.py
@@ -17,7 +17,7 @@ from typing import Dict, Union
 
 import numpy
 
-from nvtabular.dispatch import DataFrameType, _hash_series, _is_series_object, annotate
+from nvtabular.dispatch import DataFrameType, _hash_series, annotate
 
 from .operator import ColumnSelector, Operator
 
--- a/nvtabular/ops/hashed_cross.py
+++ b/nvtabular/ops/hashed_cross.py
@@ -17,7 +17,7 @@ from typing import Dict, Union
 
 import numpy
 
-from nvtabular.dispatch import DataFrameType, _hash_series, annotate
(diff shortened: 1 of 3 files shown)


nvtabular/ops/hashed_cross.py
```diff
@@ -59,7 +59,10 @@ def transform(self, col_selector: ColumnSelector, df: DataFrameType) -> DataFram
         for cross in _nest_columns(col_selector.names):
             val = 0
             for column in cross:
-                val ^= _hash_series(df[column])  # or however we want to do this aggregation
+                if _is_series_object(val):
+                    val ^= _hash_series(df[column])  # or however we want to do this aggregation
+                else:
+                    val = _hash_series(df[column])
```
bdice, line 65:
```suggestion
                # Currently only __xor__ is implemented for cudf Series, not __rxor__.
                val = df[column] ^ val
```

bdice: Regarding the comment, `# or however we want to do this aggregation`, it might be best to use a hash-combining function so that equivalent values don't end up giving the same hash. I haven't looked at how this function is used, so please forgive me if this is not helpful. 😄

For example, hashing two equivalent columns make the bits negate one another with a plain XOR operator (`x ^ x == 0`). This can happen with shifted data as well: hashing the rows (or columns) of `[[1, 2, 3], [2, 3, 1], [3, 1, 2]]` would give the same hashes for every row (column), despite every row (column) being unique. Here is an example of a hash-combining function that might be helpful here.
https://github.com/rapidsai/cudf/blob/57ff6f55b9fd44e8a8e10282d3f95d5f38e299ef/cpp/include/cudf/detail/utilities/hash_functions.cuh#L125-L132

albert17: Thaanks!! :)

benfred (approved):

PR #1380, opened 2022-02-02
Import `Distributed` and `Serial` execution-manager utilities from merlin-core

### LINKED ISSUE #1249: [FEA] Add DistributedClient API and use global client objects in Workflow and Dataset

rjzamora:
Closes #1249

This is a follow-up to #1318.  In that PR, we added automatic client detection for distributed execution with Dask. That PR added the `set_dask_client` utility to enable the user to control which client to use for distributed execution (and when). However, if stopped short of automating the deployment of local clusters.  It also stopped short of providing context-manager support for distributed and/or serial execution.

This PR adds new `Distributed` and `Serial` context-manager APIs to NVTabular. For users who require distributed execution on a single machine, these APIs remove the need to directly interact with external libraries like `distributed` and `dask_cuda`. For all distributed-execution applications, these APIs can be used as convenient context managers (to hopefully make user code more readable and convenient).

Simple example (where a default local cluster will be deployed):

```python
workflow = nvt.Workflow(["col"] >> ops.Normalize())
with nvt.Distributed():
    dataset = nvt.Dataset(...)
    workflow.transform(dataset).to_parquet(...)
```

More complex example (with user-defined LocalCUDACluster options):

```python
workflow = nvt.Workflow(["col"] >> ops.Normalize())
with nvt.Distributed(
    cluster_type="cuda",
    n_workers=4,
    local_directory="/raid/dask-space",
    protocol="ucx",
    device_memory_limit=0.8,
    rmm_pool_size="20GB",
    log_spilling=True,
):
    dataset = nvt.Dataset(...)
    workflow.transform(dataset).to_parquet(...)
```

For workflows where a `with` block is not convenient. The user also has the option to define (and later `deactivate`) the execution manager manually.  It is also possible to nest the `Distributed` and `Serial` execution managers if/when
the user wants to execute a smaller subset of the full workflow in serial.


rjzamora:
@benfred @bschifferer @rnyak @albert17 - Interested in your feedback on this general design.

karlhigley (approved):

rjzamora:
rerun tests

rjzamora:
I don't think the current conflicts can be resolved, so I copied most of this into [merlin-core#21](https://github.com/NVIDIA-Merlin/core/pull/21). After the new utilities are merged in merlin-core, we can update this PR to simply import `Distributed`/`Serial` in the `nvtabular` namespace (to enable `with nvt.Distributed:`).  However, none of the logic would live in nvt anymore, so we could also close this altogether. @karlhigley - Let me know if you have any thoughts on how/where users should import the new context managers.

karlhigley:
Hey @rjzamora, sorry for the conflicts! This PR is great, and I definitely want to get it integrated into NVT and/or `merlin-core`. 😅 

Since `Dataset` (and AFAIK the majority of Dask-related code) has migrated to `merlin-core` and is also used in `merlin-models`, I think maybe these context managers should be imported from `merlin.io`. That way they can also be used in the other Merlin libraries where managing distributed execution will also be useful.

rjzamora:
>sorry for the conflicts!

No worries! The re-org process is bound to cause conflicts at times.

>Since Dataset (and AFAIK the majority of Dask-related code) has migrated to merlin-core and is also used in merlin-models, I think maybe these context managers should be imported from merlin.io. That way they can also be used in the other Merlin libraries where managing distributed execution will also be useful.

I definitely agree that it makes sense to put the context managers in merlin-core (i.e. [merlin#21](https://github.com/NVIDIA-Merlin/core/pull/21)), but the `io` namespace doesn't feel quite right to me. I used `utils` in merlin#21, but I'm open to other suggestions if you have time to review that PR :)

karlhigley:
Yeah, I'm with you on `io` not feeling quite right. I think that's where the code these are most closely related to (i.e. the code for Dask-based classes like `Dataset`) lives though, right? Wondering if the name of the package still makes sense in the context of `merlin-core`.

(Having the same issue with the `graph` package which, having moved to `core` should probably now be called `dag`.)

PR #1384, opened 2022-02-04
Migrate io, graph, dispatch, worker, and utils to `merlin.core`

karlhigley:
This PR removes code that's now present in `merlin.core`, re-exports methods from those submodules in their original locations in NVT, and updates method names from private (`_method`) to public (`method`) where necessary.


jperez999 (approved):

tests/unit/test_triton_inference.py
```diff
@@ -205,7 +205,7 @@ def test_concatenate_dataframe(tmpdir, output_model):
         tmpdir,
         workflow,
         df,
-        "test_concatenate_dataframe",
+        "testconcatenate_dataframe",
```
benfred, line 208:
```suggestion
        "test_concatenate_dataframe",
```

tests/unit/workflow/test_workflow_ops.py
```diff
@@ -14,11 +14,12 @@
...
 #
 
+from merlin.graph import ops as graph_ops
+
 from nvtabular import ops as workflow_ops
-from nvtabular.graph import ops as graph_ops
 
 
-def test_column_concat_op():
+def test_columnconcat_op():
```
benfred, line 22:
```suggestion
def test_column_concat_op():
```

benfred (commented):
looks good! A couple minor nit-picks below,

Also - I think we should eventually deprecate nvtabular.dispatch. There are some parts of the code here that uses merlin.core.dispatch and others that use the re-exports from nvtabular.dispatch - which is a little inconsistent . This can be done in a future PR though

benfred (approved):

rjzamora:
It looks like this removed the dask client-handling logic, making #1380 completely unmergable... I guess I will need to re-implement everything in merlin/core?  Note that I do like the merlin-core idea, but now the client-handling logic and test coverage are in two different places.

karlhigley:
@rjzamora I made as close to a straight copy/paste into `merlin-core` as possible, so hoping that it's mostly possible to pick the code from #1380 up and drop it in `merlin-core`, with minor changes to the imports and the removal of the leading `_` from dispatch methods. I tried to move the relevant tests for the sub-packages (like `io`) that moved to `merlin-core` too, but I'm not sure I got everything. Are there more tests that should move?

karlhigley:
Yeah, looks `test_dask_nvt.py` stayed in NVTabular, which I think is because the existing tests there make use of NVT ops that aren't accessible to the tests in `merlin-core`. From what I can tell though, the new tests added in #1380 don't use NVT ops and could live in a similarly named test file in `merlin-core` (based on their limited dependencies.)

rjzamora:
@karlhigley - I think this PR did the best it could to migrate the appropriate tests. The fact that utils were being tested in `test_dask_nvt.py` was my fault (my earlier "complaint" is mostly with myself). I'm pretty sure [merlin#21](https://github.com/NVIDIA-Merlin/core/pull/21) is doing what you had in mind.  Moving forward, if we find tests that don't really depend on nvt-specific functionality, we should just try to move them from nvt to merlin-core. It seems like you are already doing this, but I'm just establishing that this is probably the best we can do.  Perhaps we could also run some of the nvt tests in merlin-core CI to be sure integration is fully covered.

karlhigley:
Yeah, we might want to start thinking about how to do Merlin-level integration tests across the increasing number of libraries we have. `merlin-models` consumes output from `nvtabular`, `merlin-systems` will consume output from `merlin-models`, everything will depend on `merlin-core`, etc

PR #1399, opened 2022-02-17
Itemsize pb

jperez999:
This PR introduces the "dtype_itemsize" property. This will allow us to maintain full dtype descriptions when exporting the protobuf tensorflow metadata config.


nvtabular/graph/schema.py
```diff
@@ -42,14 +42,22 @@ def __post_init__(self):
                 dtype = np.dtype(self.dtype.numpy_dtype)
             elif hasattr(self.dtype, "_categories"):
                 dtype = self.dtype._categories.dtype
-            else:
+            # CAUTION: fun fact np.dtype(None) == np.float64, must be avoided
```
karlhigley, line 45: 🥇

karlhigley (commented):
Looks pretty good to me! Definitely run this by @benfred though, since he's working on schemas in `merlin-core` etc. We're in the tricky period where we have two copies of this code, so want to keep everyone in sync on what's changing.

PR #1403, opened 2022-02-18
Fix/remove columns

jperez999:
adds on top of @karlhigley's PR. to pass all test. paves way for pytorch  inference notebooks. starts from:  https://github.com/NVIDIA-Merlin/NVTabular/pull/1389


nvtabular/framework_utils/torch/utils.py
```diff
@@ -113,20 +113,21 @@ def process_epoch(
...
-            if device:
-                x_cat = x_cat.to(device)
-                x_cont = x_cont.to(device)
-                y = y.to(device)
+            x, y = batch
+            # x_cat, x_cont, y = transform(batch)
+            # if device:
+            #     x_cat = x_cat.to(device)
+            #     x_cont = x_cont.to(device)
+            #     y = y.to(device)
```
karlhigley, line 121: Does this commented code need to be removed to make this work? Does it still work if it's uncommented?

nvtabular/framework_utils/torch/models.py
```diff
@@ -107,4 +114,4 @@ def forward(self, x_cat, x_cont):
         if self.max_output:
             x = self.max_output * torch.sigmoid(x)
         x = x.view(-1)
-        return x
+        return {"predictions": x}
```
karlhigley, line 117: Does the output have to change here? Looks like the places where predictions happen are also updated to fetch `predictions`, so wondering if this change washes out and can be dropped

jperez999: so actually when you remove this you get the following error from triton: `Failed to process the request(s) for model 'test_name_pt', message: ValueError: output of the forward function should be a dict` this is why I added that change it needs to be a dictionary on the way out for triton to accept it.

karlhigley: Makes sense!

karlhigley:
Think this has been superseded by later PRs?

PR #1414, opened 2022-03-01
Add TagAs and related wrapper classes

### LINKED ISSUE #1404: [FEA] Simplify AddMetadata Tag for NVTabular
Currently, we provide the functionality for the user to tag columns with the operator `AddMetadata`.
The use case is that users will use the operator mainly for adding tags. 

Should we provide a wrapper called `AddTag` or `TagAs` to simplify the AddMetadata operator?
Should we provide multiple wrappers for common tags - e.g.

`TagAsUserID()`, `TagAsItemID()`, `TagAsUserFeatures()`, `TagAsItemFeatures()`, etc.

radekosmulski:
Looking at the issues, I thought #1404 would be a nice way of getting introduced to the codebase so I went ahead and implemented a proof of concept of what this might look like.

I can see that there has been no discussion on this feature request (also, only now learned about the existence of the project board), fully understood if this is not something that might be needed. 

Anyhow, I welcome any feedback and comments and am happy to make any changes that might be needed 🙂 

Thank you!

resolves #1404


radekosmulski:
The test failing in CI fails on `main` as well.

Also, now after `black` and `pylint` I guess there wouldn't be a whole lot of difference if `TagAs<something>` were to be just child classes of `TagAs` as they would take up as much space and removing the `TagAsMeta` would probably improve readability.

nvtabular/ops/add_metadata.py
```diff
@@ -36,6 +34,42 @@ def transform(self, col_selector: ColumnSelector, df: DataFrameType) -> DataFram
...
+    def __init__(self, tags=None, properties=None):
+        super().__init__(tags)
+        self.properties = properties or {}
+
     @property
     def output_properties(self):
         return self.properties
+
+
+# Wrappers for common features
```
karlhigley, line 53: These classes look good. 👍🏻 

They should use the canonical tags (`Tags.USER` for user features, etc) from `merlin-core` though: https://github.com/NVIDIA-Merlin/core/blob/main/merlin/schema/tags.py#L20

nvtabular/ops/add_metadata.py
```diff
@@ -36,6 +34,42 @@ def transform(self, col_selector: ColumnSelector, df: DataFrameType) -> DataFram
     def output_tags(self):
         return self.tags
 
+
+class AddMetadata(AddTags):
```
karlhigley, line 38: Seems like the goal with `AddTags` is to provide a more intuitive operator name, so it seems like the simplest way to do that is probably to make `AddTags` a subclass of `AddMetadata` where the `init` only accepts tags and `output_tags` returns them, and then to create a second sub-class called `AddProperties` that does the same thing but via `output_properties`

karlhigley: (That way we can maintain the ability to add tags and properties simultaneously with `AddMetadata`, and still provide operators with more intuitive names.)

radekosmulski:
thank you very much for the feedback, @karlhigley! appreciate it! 🙂 I just pushed an updated version, happy to make any changes that might be needed

nvtabular/ops/add_metadata.py
```diff
@@ -39,3 +40,38 @@ def output_tags(self):
...
+
+
+# Wrappers for common features
+class TagAsUserID(Operator):
+    @property
+    def output_tags(self):
+        return [Tags.USER_ID]
+
+
+class TagAsItemID(AddTags):
```
karlhigley, line 62: I think these next few classes should just subclass `Operator` like the first one

tests/unit/ops/test_ops_schema.py
```diff
@@ -27,6 +27,12 @@
         ops.Normalize(),
         ops.TargetEncoding(["1"]),
         ops.AddMetadata(tags=["excellent"], properties={"domain": {"min": 0, "max": 20}}),
+        ops.AddTags(tags=["excellent"]),
+        ops.AddProperties(properties={"domain": {"min": 0, "max": 20}}),
+        ops.TagAsUserID(),
+        ops.TagAsItemID(),
+        ops.TagAsUserFeatures(),
+        ops.TagAsItemFeatures(),
```
karlhigley, line 35: @jperez999 Is this sufficient to test that these operators add the tags we're expecting them to, or is there another place we need to add tests for that?

jperez999: no this is good. This is our test framework for testing all metadata implementations.

nvtabular/ops/add_metadata.py
radekosmulski: apologies, that was an oversight on my part, fixed this now

karlhigley: No worries! You're taking tickets off my queue, so I'm not going to complain. 😀

karlhigley:
Thanks for the contribution, @radekosmulski! 🚀

radekosmulski:
np at all, glad I can be of help! 🙂

Issue #1427, opened 2022-03-04
[BUG] output schema (dtype) not being correctly inferred with groupby when agg is string


radekosmulski:
**Describe the bug**
When `agg`  is string,  `nvt.ops.Groupby` infers the output `dtype` to be that of the input column which results in an error:
```
TypeError: Improperly matched output dtypes detected in product_id_count, int64 and int32
```

This does not occur when `agg` is `dict`.

**Steps/Code to reproduce bug**
The following code:

```
sales = {
    "product_id": [1, 2, 1],
    "week_of_sale": [1, 1, 2]
}

df_sales = make_df(sales)

total_sales_per_week = ['product_id', 'week_of_sale'] >> nvt.ops.Groupby(
    groupby_cols=['week_of_sale'], 
    aggs='count'
)

workflow = nvt.Workflow(total_sales_per_week)
dataset = nvt.Dataset(df_sales)
workflow.fit(dataset)
workflow.transform(dataset).to_ddf().compute()
```
results in:
```
Failed to transform operator <nvtabular.ops.groupby.Groupby object at 0x7fc78f1592d0>
Traceback (most recent call last):
  File "/home/radek/workspace/NVTabular/nvtabular/workflow/workflow.py", line 486, in _transform_partition
    f"Improperly matched output dtypes detected in {col_name},"
TypeError: Improperly matched output dtypes detected in product_id_count, int64 and int32
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_3801/3727759829.py in <module>
     14 dataset = nvt.Dataset(df_sales)
     15 workflow.fit(dataset)
---> 16 workflow.transform(dataset).to_ddf().compute()

~/anaconda3/envs/nv/lib/python3.7/site-packages/dask/base.py in compute(self, **kwargs)
    286         dask.base.compute
    287         """
--> 288         (result,) = compute(self, traverse=False, **kwargs)
    289         return result
    290 

~/anaconda3/envs/nv/lib/python3.7/site-packages/dask/base.py in compute(traverse, optimize_graph, scheduler, get, *args, **kwargs)
    569         postcomputes.append(x.__dask_postcompute__())
    570 
--> 571     results = schedule(dsk, keys, **kwargs)
    572     return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
    573 

[...]

PR #1428, opened 2022-03-04
add basic test for groupby with string agg

radekosmulski:
This is a rudimentary test to demonstrate #1427


karlhigley:
Resolved by #1430

PR #1434, opened 2022-03-07
docs: Ensure that parent review directory exists

mikemckiernan:
I reviewed the failure in https://github.com/NVIDIA-Merlin/NVTabular/runs/5456159239?check_suite_focus=true#step:6:12

My assessment is that the `mv` command failed because the `review` directory does not exist on the `gh-pages` branch.  As an aside, the directory was certainly created and present during the evolutionary process of developing the workflow in another repository.

### DIFF
--- a/.github/workflows/docs-preview-pr.yaml
+++ b/.github/workflows/docs-preview-pr.yaml
@@ -66,7 +66,8 @@ jobs:
           echo "PR_NO=${PR}" >> $GITHUB_ENV
           echo "MERGE_STATUS=${MERGED}" >> $GITHUB_ENV
           echo "PR_ACTION=${ACTION}" >> $GITHUB_ENV
-          echo "REVIEW_DIR=review/pr-${PR}" >> $GITHUB_ENV
+          echo "REVIEW_DIR=review/" >> $GITHUB_ENV
+          echo "PR_REVIEW_DIR=review/pr-${PR}" >> $GITHUB_ENV
 
           # Remove the pr artifact directory so that it does not
           # appear in listings or confuse git with untracked files.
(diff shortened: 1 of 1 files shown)


karlhigley:
The notebook tests on this PR are failing due to an unrelated issue with the inference code. I'll put in a fix for that.

mikemckiernan:
> The notebook tests on this PR are failing due to an unrelated issue with the inference code. I'll put in a fix for that.

Ah.  I just slacked to Julio that I don't have access to Jenkins to see the issue.  Not a biggie for now.

I suppose the bigger issue is that (understandably) GH is running the workflow file from `main` rather than the workflow file in the PR.  I guess that I'm kind of stuck w/o the ability to merge (or reliant on help from reviewers).  I hate guess-o-rama.

karlhigley:
Yeah, I don't think we have the repos configured to run workflows from PRs. @benfred is that configuration something we can change?

mikemckiernan:
> Yeah, I don't think we have the repos configured to run workflows from PRs. @benfred is that configuration something we can change?

I'm OK with *not* enabling that.  That sounds a little dicey with regard to untrusted code.  I just thought that team membership would make me "trusted" and run from the PR.

karlhigley:
I think we have to approve running actions on code from non-team members anyway, so we might be okay w.r.t. untrusted code even if we do allow running actions from PRs.

mikemckiernan:
Still no good because the workflow file from `main` ran rather than the corrected (as best as I can see) one that I submitted in the PR.

https://github.com/NVIDIA-Merlin/NVTabular/runs/5457273851?check_suite_focus=true

Julio and I were chatting.  My $0.02 is that it makes sense for GH _not_ to run a workflow file from a PR.  Especially not one that runs `on.workflow_run`.

I regret that this has become a hassle.  Sorry about that.

karlhigley (approved):

PR #1450, opened 2022-03-14
doc: Update matrix for 22.03

mikemckiernan:
* Updates from Alberto's version script.

* Some updates after using NGC for more difficult
  to discover values.


mikemckiernan:
Until we can replace this matrix with automatically updated information in the Merlin repository, here's the 22.03 update.

https://nvidia-merlin.github.io/NVTabular/review/pr-1450/resources/support_matrix.html

I removed whatever was in the left-most column and added 22.03 on the right.  The version that I removed ought to be available from the version picker.

karlhigley (approved):

EvenOldridge:
Looks great!  Thanks @mikemckiernan!

PR #1474, opened 2022-03-24
fix(docs): Restore the version picker

mikemckiernan:
This is the process that make sense to me--please review the code and tell me if my intent doesn't match the code.

* For PRs, create a doc preview like we've done before.
* Small enhancement to the doc preview--tolerate the PR failing to build (due to pre-commit) and add a PR comment if one doesn't already exist.
* Stop clobbering the `main` build because the doc preview does not cover the previous six releases.
* Add a nightly build that updates `main` and the last six releases.

If you can spot a way to make this more robust, PLMK.

I have some evidence of success:

* Output from the nightly build with `main` and six previous releases: https://mikemckiernan.github.io/NVTabular/main/


karlhigley (approved):

karlhigley:
It looks like this is having trouble with `natsort`, which I think should be installed but maybe isn't?

jperez999:
rerun tests

mikemckiernan:
rerun tests

PR #1476, opened 2022-03-28
Add missing `--user` flag to `natsort` CI install

karlhigley:


mikemckiernan (approved):
Thank you.  I can not defend missing `--user` on that one when it's completely surrounded by others that specify it.

PR #1487, opened 2022-03-30
Remove imports to deprecated modules

benfred:
Functionality like dispatch,io,utils,graph has moved to the merlin-core package.
This change fixes imports to use the merlin-core packages directly, instead of
the API compatability stubs included here

### DIFF
--- a/nvtabular/__init__.py
+++ b/nvtabular/__init__.py
@@ -15,10 +15,14 @@
 #
 import warnings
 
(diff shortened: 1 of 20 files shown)


karlhigley (approved):

viswa-nvidia:
Its a nice to have - there are a bunch of warnings being thrown in the notebooks etc, and this will help clean them up

PR #1494, opened 2022-04-01
Note in the README that installing with pip runs only on CPU

karlhigley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ NVTabular can be installed with `pip` by running the following command:
 pip install nvtabular
 ```
 
-Note: Installing NVTabular with Pip will require installing some additional dependencies (like CUDA Toolkit) manually. Prefer either Conda or Docker where possible.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,7 +59,7 @@ NVTabular can be installed with `pip` by running the following command:
 pip install nvtabular
 ```
 
-Note: Installing NVTabular with Pip will require installing some additional dependencies (like CUDA Toolkit) manually. Prefer either Conda or Docker where possible.
+Note: NVTabular installed with pip runs only on CPU, and may require installing some additional dependencies manually. Prefer installing with either Conda or Docker where possible.
```
mikemckiernan, line 62:
```suggestion
> Installing NVTabular with Pip causes NVTabular to run on the CPU only and might require installing additional dependencies manually.
> When you run NVTabular in one of our Docker containers, the dependencies are already installed.
```

mikemckiernan (approved):
Tweak based on a similar note that I added to models.

Issue #1455, opened 2022-03-16
[BUG] Workflow typerror during integrations tests


albert17:
```
tests/integration/test_nvt_tf_inference.py:206: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

root_df =       Store  DayOfWeek  ... StateHoliday_fw  Promo_fw
0       386          2  ...               0         4
1       97...       2  ...               0         7
1023    511          2  ...               0         7

[1024 rows x 75 columns]
workflow_nodes = [<Node + output>], additional_columns = None
capture_dtypes = False

    def _transform_partition(root_df, workflow_nodes, additional_columns=None, capture_dtypes=False):
        """Transforms a single partition by appyling all operators in a WorkflowNode"""
        output = None
    
        for node in workflow_nodes:
            node_input_cols = _get_unique(node.input_schema.column_names)
            node_output_cols = _get_unique(node.output_schema.column_names)
            addl_input_cols = set(node.dependency_columns.names)
    
            # Build input dataframe
            if node.parents_with_dependencies:
                # If there are parents, collect their outputs
                # to build the current node's input
                input_df = None
                seen_columns = None
    
                for parent in node.parents_with_dependencies:
                    parent_output_cols = _get_unique(parent.output_schema.column_names)
                    parent_df = _transform_partition(root_df, [parent], capture_dtypes=capture_dtypes)
                    if input_df is None or not len(input_df):
                        input_df = parent_df[parent_output_cols]
                        seen_columns = set(parent_output_cols)
                    else:
                        new_columns = set(parent_output_cols) - seen_columns
                        input_df = concat_columns([input_df, parent_df[list(new_columns)]])
                        seen_columns.update(new_columns)
    
                # Check for additional input columns that aren't generated by parents
                # and fetch them from the root dataframe
                unseen_columns = set(node.input_schema.column_names) - seen_columns
                addl_input_cols = addl_input_cols.union(unseen_columns)
    
                # TODO: Find a better way to remove dupes
                addl_input_cols = addl_input_cols - set(input_df.columns)
    
                if addl_input_cols:
                    input_df = concat_columns([input_df, root_df[list(addl_input_cols)]])
            else:
                # If there are no parents, this is an input node,
                # so pull columns directly from root df
[...]

PR #1495, opened 2022-04-01
fix normalize

jperez999:
This PR is related to #1455. The normalize operator was using an itemsize 64 output dtype clashing with the other defaults of 32. We change to itemsize 32 in order to allow integration tests to pass.

### DIFF
--- a/ci/build_and_test.sh
+++ b/ci/build_and_test.sh
@@ -11,7 +11,7 @@ python -m pip install --user --upgrade pip setuptools wheel pybind11 numpy==1.20
 python -m pip uninstall nvtabular merlin-core -y
 python -m pip install --user --upgrade --no-cache-dir merlin-core@git+https://github.com/NVIDIA-Merlin/core.git
 python -m pip install --user natsort==8.1.0
-python setup.py develop --user --no-deps
+python setup.py develop
 
 # following checks requirement requirements-dev.txt to be installed
 echo "Running black --check"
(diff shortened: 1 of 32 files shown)


karlhigley:
Looks like this change breaks the TF4Rec test with a dtype discrepancy over there instead 🤔

ci/build_and_test.sh
```diff
@@ -11,7 +11,7 @@ python -m pip install --user --upgrade pip setuptools wheel pybind11 numpy==1.20
 python -m pip uninstall nvtabular merlin-core -y
 python -m pip install --user --upgrade --no-cache-dir merlin-core@git+https://github.com/NVIDIA-Merlin/core.git
 python -m pip install --user natsort==8.1.0
-python setup.py develop --user --no-deps
+python setup.py develop
```
benfred, line 14: This is breaking CI - do you need to make this change?

nvtabular/ops/normalize.py
```diff
@@ -94,16 +94,19 @@ def supports(self):
         )
 
     def clear(self):
+        """Clear previously recorded stats"""
```
benfred, line 97: Can we just use the docstrings from the parent class? I'd rather not maintain two copies of the same content just to make interrogate happy

benfred (approved):

jperez999:
Not the correct fix

PR #1511, opened 2022-04-13
Fix the embedding size lookup in `Categorify` op

karlhigley:
When columns are jointly encoded, the embedding size needs to be looked up by the shared category name, instead of the column name.


karlhigley:
rerun tests

karlhigley:
rerun tests

PR #1530, opened 2022-04-28
remove horovod example, no longer supported

jperez999:
we already have torch distributed example!


karlhigley:
I think there's also a notebook that should be removed?

mikemckiernan:
It's cool by me to remove the notebook if it no longer applies.  I'm chiming in because I planned to make that notebook public in the doc:

<https://nvidia-merlin.github.io/NVTabular/review/pr-1529/examples/multi-gpu-movielens/01-03-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-PyTorch.html?highlight=torch_trainer#>

Please send me a signal to confirm that we want to remove the notebook.

karlhigley:
@mikemckiernan Yeah, the notebook has to go, since we're dropping support for the PyTorch/Horovod combo and removing Horovod from the Torch containers. Doesn't matter to me which PR it gets removed in as long as it disappears somehow.

bschifferer:
@jperez999 where is the "torch distributed example"? Do we know if there are performance differences between using torch.distributed and horovod?

bschifferer:
I think the notebook ( https://github.com/NVIDIA-Merlin/NVTabular/blob/main/examples/multi-gpu-movielens/01-03-MultiGPU-Download-Convert-ETL-with-NVTabular-Training-with-PyTorch.ipynb ) generated torch_trainer.py. 

We need to remove the notebook, as well / update to generate `torch_trainer_dist.py` ?

jperez999:
@bschifferer Sorry I don't know the diff off the top of my head. And I don't think my old perf runs are good for current comparison.

karlhigley (approved):

PR #1532, opened 2022-04-29
Fix getting started 03-Training-with-HugeCTR.ipynb  nb's training without printing out auc and loss metrics issue

rnyak:
This is fixing https://github.com/NVIDIA-Merlin/NVTabular/issues/1521


rnyak:
rerun tests

bschifferer (approved):

PR #1540, opened 2022-05-11
Add release-drafter workflow for generating changelogs

benfred:
Automatically generate changelogs using release-drafter
https://github.com/release-drafter/release-drafter . This will
categorize PR's into different sections of the changelog based
on labels (breaking/bug/feature etc).

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,0 +1,23 @@
+name: Release Drafter
+
+on:
+  push:
+    # branches to consider in the event; optional, defaults to all
+    branches:
+      - main
+  # pull_request event is required only for autolabeler
+  pull_request:
+    # Only following types are handled by the action, but one can default to all as well
+    types: [opened, reopened, synchronize]
+  # pull_request_target event is required for autolabeler to support PRs from forks
+  pull_request_target:
+    types: [opened, reopened, synchronize]
+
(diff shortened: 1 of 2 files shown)


.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,23 @@
...
+  push:
+    # branches to consider in the event; optional, defaults to all
+    branches:
+      - main
+  # pull_request event is required only for autolabeler
+  pull_request:
+    # Only following types are handled by the action, but one can default to all as well
+    types: [opened, reopened, synchronize]
+  # pull_request_target event is required for autolabeler to support PRs from forks
+  pull_request_target:
```
mikemckiernan, line 13: Initially I planned to request that we drop this `on` condition. However, it seems that the diceyness involves `pull_request_target` and a checkout.  There's no checkout, so /lgtm.

mikemckiernan (commented):
I still don't know how or where the "in progress" release notes are viewable but I'm game for this change because the outcome is clear on other projects' Releases pages.

mikemckiernan (approved):

benfred:
thanks for the review @mikemckiernan - I've updated the whitespace before $CHANGES . I'm betting we'll need some tweaks to this, but I thought we could use NVT as a proof of concept while we iron out the bugs -

PR #1551, opened 2022-05-17
Improved docstrings of GroupBy op to reinforce the required usage of dataset.shuffle_by_keys() 

gabrielspmoreira:


nvtabular/ops/groupby.py
```diff
@@ -46,10 +49,15 @@ class Groupby(Operator):
...
 
+        workflow.fit(dataset)
+        dataset_transformed = workflow.transform(dataset)
+
     Parameters
     -----------
     groupby_cols : str or list of str
         The column names to be used as groupby keys.
+        WARNING: Ensure the dataset was partitioned by those
+        groupby keys (see above an example).
```
benfred, line 60:
```suggestion
        groupby keys (see above for an example).
```

benfred (approved):
Thanks! left a nitpick on some of the wording, but looks good aside from that

PR #1563, opened 2022-05-25
revamp how we use the proto file

jperez999:
This PR introduces a new file which is charged with interacting with protobuf on the system and gathering the required messages and enums from the previously created protobuf file. This will decouple the proto compilation, from the usage.  Which alleviates the file cannot be created because already exists in the pool.

### DIFF
--- a/nvtabular/inference/triton/model_config_util.py
+++ b/nvtabular/inference/triton/model_config_util.py
@@ -4,24 +4,23 @@ from google.protobuf.message_factory import MessageFactory
 def_pool = Default()
 msg_factory = MessageFactory(pool=def_pool)
 file_descriptor = def_pool.FindFileByName("model_config.proto")
-if file_descriptor:
-    ModelConfig = msg_factory.CreatePrototype(file_descriptor.message_types_by_name["ModelConfig"])
-    ModelInput = msg_factory.CreatePrototype(file_descriptor.message_types_by_name["ModelInput"])
-    ModelOutput = msg_factory.CreatePrototype(file_descriptor.message_types_by_name["ModelOutput"])
-    ModelEnsembling = msg_factory.CreatePrototype(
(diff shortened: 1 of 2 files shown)


jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

nvtabular/inference/triton/model_config_util.py
```diff
@@ -0,0 +1,24 @@
+from google.protobuf.descriptor_pool import Default
+from google.protobuf.message_factory import MessageFactory
+
+def_pool = Default()
+msg_factory = MessageFactory(pool=def_pool)
+file_descriptor = def_pool.FindFileByName("model_config.proto")
+if file_descriptor:
```
karlhigley, line 7: What happens if this isn't found? (I'm not sure we need an else clause, I just don't understand how it works in the file doesn't exist.)

jperez999: so your right by the time you hit this code the protoc will have compiled the code (during install) and then we will always have all the pieces guaranteed. Otherwise the install did not occur correctly.

benfred (commented):
with this change, do we still need to compile the proto file to python in setup.py? If not - we should probably remove from there

jperez999:
> with this change, do we still need to compile the proto file to python in setup.py? If not - we should probably remove from there

we do still need to compile it. But during compilation it gets registered. so we should only interact with the model_config_util. This way no matter where the access comes from the object is what we pull and not the actual proto instantiations.

karlhigley:
It looks like this failed on test coverage? Haven't seen that before. 🤔

benfred:
Don't think this is necessary anymore - closing

PR #1566, opened 2022-05-26
Don't build model_config_pb2

benfred:
Instead of building the model_config.proto file, just use the one thats
installed with tritonclient instead


benfred:
rerun tests

jperez999 (approved):

PR #1580, opened 2022-06-10
add casting for additional aggs in groupby

### LINKED ISSUE #1574: [BUG] Groupby aggregations on non float64 columns throws an error
**Describe the bug**
When running `Groupby` on columns with a dtype other than `float64`, I receive the following error:

```
TypeError: Dtype discrepancy detected for column measurement_std: operator Groupby reported dtype `float32` but returned dtype `float64`.
```

image

Here is code that can be copied over
```
import cudf
import nvtabular as nvt

gdf = cudf.DataFrame(data={'cat': ['a', 'a', 'b'], 'timestamp': [1, 2, 1], 'measurement': [0.1, 0.2, 0.5]})

gdf.measurement = gdf.measurement.astype('float32')

out = ['measurement', 'cat'] >> nvt.ops.Groupby('cat', aggs=['std'])


ds = nvt.Dataset(gdf)
wf = nvt.Workflow(out)
res = wf.fit_transform(ds)

res.compute()
```

**Expected behavior**
The transformation to work, results to be returned without error.

**Environment details (please complete the following information):**
`merlin-pytorch-training:22.05` image with all Merlin repos updated and pip installed today

radekosmulski:
resolves #1574


karlhigley:
This looks good! 👍🏻 

Seems like the test assertions might need some updates to check the new aggregations though?

radekosmulski:
I added some asserts for new aggregations. I thought that getting an assert for `nunique` and `mean` would be a good start, given how different they are. Essentially, I duplicated the logic that was there before, and I perform calculations on the lists and compare them to the output from aggregations.

But unfortunately, this is now failing in a new way... If I want to aggregate a `float32` column into a list, I get the following error:
```
TypeError: Dtype discrepancy detected for column z-list: operator Groupby reported dtype `float32` but returned dtype `float64`.
```

I looked briefly around, but have a hard time figuring out where to even start troubleshooting this. This doesn't seem to be an issue with casting like it was the case with the aggregations that were missing.

nvtabular/ops/groupby.py
```diff
@@ -184,7 +186,14 @@ def _compute_dtype(self, col_schema, input_schema):
         dtype = col_schema.dtype
         is_list = col_schema.is_list
 
-        dtypes = {"count": numpy.int32, "mean": numpy.float32}
+        dtypes = {
+            "count": numpy.int32,
+            "nunique": numpy.int32,
```
rnyak, line 191: @radekosmulski not sure we can do `nunique` agg with groupby. did you test it? thanks.

radekosmulski: With the changes in this PR `nunique` and a couple of other aggs work now 🙂 

image

But please don't merge it just yet -- the tests are passing in CI because the transformations don't run on a GPU, but when you run the tests for `Groupby` on a GPU, they fail.

Problem is, I didn't get a good handle on how to fix this today. The list transformation that is throwing the error upon schema dtype verification is different than these other groupby aggs, 

I don't even know where to start looking for where the issue might lie apart from just trying to figure out what is happening where via doing `set_trace` in various places in code... Will think on how to approach this, maybe asking someone to pair on this or to give me some pointers would be a good idea here 🙂

radekosmulski: The casting that you mentioned, is what I have done for the other aggregations to make them work, but the `list` one is a bit more tricky.

I can have a list of `object`s
image

And also this seems to fail only when run on the CPU but works on the GPU

image

There is something much more tricky going on here 🙂 

BTW just going to a list of strings works
image

I think ideally what we might want is for the list op to honor the original dtype of the column, but somewhere, somehow, `float32` still gets cast to `float64` on the CPU (I guess it must be something `dask_cudf` is doing) but I am a bit unsure where to start unraveling this in our code.

Maybe another session of looking at `_apply_aggs` from ops.grouby might help, but doing this hasn't been very efficient for me so far 🙂

karlhigley: Yeah, honoring the original dtypes makes sense to me too, but as you said, it's tricky to figure out exactly when and where dtypes are modified by the underlying libraries and code. Casting floats to a consistent precision is kind of a workaround for that.

I can see how casting list columns is trickier, that might require some thought. 🤔

radekosmulski:
Not sure if this might be helpful, but let me maybe try to document what I'm seeing

Just before returning from `_apply_aggs` in `Groupby`, when running on the CPU, here are a couple of facts about the `df`
image

And here is the same spot but on the GPU
image

The `df` gets returned from `_apply_aggs` and it then gets returned straight from `transform`.

Before it gets returned, the situation looks quite similar, it's just a column of lists with Python 64bit floats inside. Not a lot of opportunity to maneuver here, just sharing this for broader consideration, I am not sure how (or where) to address this. There doesn't seem to exist an opportunity to cast this in `Groupby` AFAICT.

radekosmulski:
I pushed to origin and not to my fork and so the new changes were not showing here 🤦‍♂️ 

This now has the test that is failing on the CPU, this is what I have been talking about.

radekosmulski:
Ok, I am thinking that this might actually be good enough! 🙂 

There are no explicit tests for the new behavior with casting (and thus the aggs running), but the fact that the test executes with these new additions is actually a testament to this working (it would fail without the changes added here).

I was unable to write a test for these aggregations actually working (summing correctly, etc) because in `test_groupby_op` I didn't manage to find a way to write it where the test would work across all the "keys" scenarios.

I am not sure how to think about this. I can add a test for the behavior of these aggs (for instance, `nunique` across partitions might be interesting) or maybe this what we have here is enough. The redeeming quality here is that what we have there now does implicitly test these aggregations work on the schema side of things.

tests/unit/ops/test_groupyby.py
```diff
@@ -161,5 +161,46 @@ def test_groupby_selector_cols():
     )
     workflow = Workflow(groupby).fit_schema(input_schema)
 
-    # If groupby_cols aren't included in the selector, they shouldn't be in the output
+    # If groupby_cols arenat included in the selector, they shouldn't be in the output
```
karlhigley, line 164:
```suggestion
    # If groupby_cols aren't included in the selector, they shouldn't be in the output
```

karlhigley (approved):

karlhigley:
Made one minor suggestion about a typo, but otherwise LGTM

PR #1581, opened 2022-06-13
Merlin offers three containers

mikemckiernan:
Beginning with the 22.06 release, the Merlin
team provides the following three containers
that are capable of training and inference:

* `merlin-hugectr`
* `merlin-tensorflow`
* `merlin-pytorch`

This is a change from the 22.05 release that
used six containers--one for training and a
separate container for inference.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -60,46 +60,53 @@ Rossmann operates over 3,000 drug stores across seven European countries. Histor
 
 ## Running the Example Notebooks
 
-You can run the example notebooks by [installing NVTabular](https://github.com/NVIDIA/NVTabular#installation) and other required libraries. Alternatively, Docker containers are available on http://ngc.nvidia.com/catalog/containers/ with pre-installed versions. Depending on which example you want to run, you should use any one of these Docker containers:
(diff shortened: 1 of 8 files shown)


examples/README.md
```diff
@@ -60,46 +60,61 @@ Rossmann operates over 3,000 drug stores across seven European countries. Histor
...
 
-```
-root@2efa5b50b909:
-```
+   ```sh
+   root@2efa5b50b909:
+   ```
 
-2. If jupyter-lab is not installed, install jupyter-lab with `pip` by running the following command:
+1. Install JupyterLab with `pip` by running the following command:
```
benfred, line 95: I think we have jupyter lab installed on our containers now by default https://github.com/NVIDIA-Merlin/Merlin/blob/027587cab360141fac7498d461c7a26ed886fe1d/docker/dockerfile.merlin#L295 - we can probably remove this section?

benfred (approved):

examples/README.md
mikemckiernan: I removed it from NVT, systems, and models. I'll catch Merlin on the next update.  Thanks for pointing it out.

PR #1589, opened 2022-06-16
Allow disabling strict checking of Workflow operator outputs vs schema

karlhigley:
This allows disabling strict checking of operator outputs against operator output schemas, which may allow some `Workflows` to run in cases where they otherwise wouldn't. This `TypeError` was introduced in order to catch cases where operators were fibbing about their own output, but we haven't fully been able to resolve all of those cases, so this provides an escape hatch that may help in some cases. Setting `strict=False` here is an acknowledgement that downstream code (e.g. in Merlin Models) may not run as expected in all cases (for example due to dtype discrepancies), but I suspect we may still be okay in many cases, so this gives another option to try.

Related to #1580 (though not an actual fix for the issue reported therein)


mikemckiernan:
I appreciate the help.  Thank you.

radekosmulski:
Fundamentally, this change is awesome 🙂 I really like this way of thinking where we are saying "okay, feel free to use the underlying machinery at your own risk. We provide you all this functionality, but if you want to step off the beaten track, feel free to do so". We are dealing with the fact that NVTabular is built on top of these other libraries very elegantly here.

So I would like to see this change merged.

But if the #1589 would get merged at some point (which I also think is important, we are significantly expanding the set of aggregations that we support and cast floats64 to floats32 which makes a lot of sense, it's easy to eat up a lot of memory with these aggregations really quickly), so if #1589 gets merged, the unit test here will stop raising an error and will fail.

ATM I don't see a good way to test the functionality in this PR (unless we find some other operator or aggregation that we currently don't support and don't plan to support in near future, or can play a trick on the operator and amend the schema mid-flight). In fact, same goes for #1589 -- I think the functionality is good there, it is just the additional test that is problematic.

radekosmulski:
Maybe even with #1589 getting merged this could be tested via doing a list aggregation of floats32 on the CPU? 🤔 This would be a genuine scenario that I am aware of which wouldn't run without `strict=False`.

tests/unit/workflow/test_workflow.py
```diff
@@ -667,3 +667,25 @@ def test_workflow_saved_schema(tmpdir):
     for node in postorder_iter_nodes(workflow2.output_node):
         assert node.input_schema is not None
         assert node.output_schema is not None
+
+
+def test_workflow_strict_mode_disabled():
```
radekosmulski, line 672: This would do the trick even with #1580 merged

```
def test_workflow_strict_mode_disabled():
    df = make_df({"cat": ["a", "a", "b"], "timestamp": [1, 2, 1], "measurement": [0.1, 0.2, 0.5]})
    df["measurement"] = df["measurement"].astype("float32")

    grouped = ["measurement", "cat"] >> ops.Groupby("cat", aggs=["list"])
    workflow = Workflow(grouped)

    dataset = Dataset(df, cpu=True)
    result = workflow.fit_transform(dataset, strict=True)

    # Strict mode should catch the dtype discrepancy
    # between the schema and the output (float32 vs float64)
    with pytest.raises(TypeError):
        result.compute()

    # Disabling strict mode should allow the workflow to run
    result_ddf = workflow.fit_transform(dataset, strict=False)
    result = result_ddf.compute()
```

radekosmulski (commented):
I would rather be looking to guidance from you on coding than the other way around 🙂 I am not sure what is the etiquette re using this review functionality on GitHub, if I am messing something up or something doesn't seem right, please let me know 🙂 

Just wanted to share the idea on the test with you, not sure if there doesn't exist a better way to handle this.

karlhigley:
@radekosmulski Thanks for the comments, I think you're entirely on target there! Re: PR etiquette etc, I'm really digging your thorough and well-thought out comments. The only thing I'd mention is that for small changes, you can create suggestions from the comment box with the button that has the "piece of paper plus/minus" icon, which lets you directly propose code changes the author can incorporate with a button click. I used it to make the change you proposed by creating a suggestion on my own PR, but we could skip a step. 😺

radekosmulski:
Thank you very much @karlhigley, great to hear! 🙂 I also really appreciate the bouncing of ideas back and forth and the discussion working on this together 🙂

Good to know about the ability to add changes that can immediately be committed from the web GUI! Could I also please ask you one more question? On #1580 I squashed my commits when I thought I was done, but I saw this kills the earlier discussion a little bit (some earlier code comments disappear, etc, which I don't think happens if you just keep pushing ). Aaaah and now I see the web GUI gives you an option to squash and commit at the end?

So if I am reading this right, the workflow would be to just keep piling commits on top of each other and then squashing and merging via the web GUI?

Thank you very much! 🙂

karlhigley:
@radekosmulski I am also normally inclined to clean up my commits, but since all PRs get squashed before merging, I've mostly stopped. Piling commits on top of each other seems to be the preferred workflow.

radekosmulski:
Thank you for your answer @karlhigley! Good to know! 🙂

viswa-nvidia:
@karlhigley , please link this to an initiative

karlhigley:
This PR is obsolete after the extraction of executors into Core

PR #1592, opened 2022-06-18
column names can now include aggregations in ops.Groupby

radekosmulski:
I found a bug where if a column contained the name of one of the aggregations in `Groupby`, and a `Groupby` op was applied to it, the `dtype` could be cast incorrectly.

The behavior is outlined in the unit test and the proposed code changes fix this.


jperez999:
rerun tests

karlhigley (approved):

radekosmulski:
Thank you very much @karlhigley for helping me with this!!!! 🙂 Much appreciated

PR #1595, opened 2022-06-22
fix notebook input output paths for int testing on read only data

jperez999:
This fixes the rossman test. Currently reading and writing data to the same place. This is not acceptable on blossom where the data is stored in a read only drive.


benfred (commented):
thanks for the fix! talking over slack though, and I think you've fixed in another way.

I also think we might need to iterate on this a bit, we need to be able to pass the output of this notebook to the inference tests - and I'm not sure this will work for that. (as part of https://github.com/NVIDIA-Merlin/Merlin/issues/343 ) .

PR #1604, opened 2022-07-06
Adjust imports in the `TritonPythonModel` for `Workflows`

karlhigley:
For reasons we don't fully understand yet, `InferenceRequest` no longer seems to be importable from `triton_python_backend_utils`. This adjusts the style of the imports to match the examples in the Python back-end examples, and removes the reference to `InferenceRequest`.

### DIFF
--- a/nvtabular/inference/triton/workflow_model.py
+++ b/nvtabular/inference/triton/workflow_model.py
@@ -26,18 +26,10 @@
 
 import json
 import logging
-import os
-from typing import List
+import pathlib
 
 import numpy as np
-from triton_python_backend_utils import (
-    InferenceRequest,
-    InferenceResponse,
-    Tensor,
-    get_input_tensor_by_name,
-    get_output_config_by_name,
-    triton_string_to_numpy,
-)
+import triton_python_backend_utils as tpb
 
 import nvtabular
 from merlin.core.dispatch import is_list_dtype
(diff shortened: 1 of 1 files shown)


jperez999 (approved):

nvtabular/inference/triton/workflow_model.py
```diff
@@ -50,15 +42,18 @@
...
+    """
+    Triton model used for running NVT Workflows
+    """
+
     def initialize(self, args):
         # Arg parsing
-        workflow_path = os.path.join(
-            args["model_repository"], str(args["model_version"]), "workflow"
-        )
+        repository_path = pathlib.Path(args["model_repository"]).parent.parent
```
oliverholworthy, line 51: the change to what `model_repository` represents might be something to do with this PR? https://github.com/triton-inference-server/python_backend/pull/138

Based on the name `model_repository`, if it's intended to be the same as the tritonserver `model_repository` I'd expect this to match the directory of the model repository passed to triton (the one with all the model directories in it). However even before the changes in release 22.06, it seems like it was set to "{model_repository}/{model_name}".

karlhigley: Yeah, I am confused by what it's intended to convey. This seems to work, but it looks pretty weird to me.

PR #1609, opened 2022-07-12
Extract Python and Dask `Executor` classes from `Workflow`

karlhigley:
We'd like to re-use some of the mechanics of graph execution (both local and distributed) in other parts of Merlin, so this is a step in the direction of disentangling graph execution from `Workflow` itself. It removes direct dependencies on Dask from `Workflow` and centralizes them in `MerlinDaskExecutor`, which `Workflow` can then use in conjunction with a Merlin operator DAG to run distributed computations.

In the future, we'd like to use these `Executor` classes in Merlin Systems too, so that we can run the full process of generating recommendations (also represented as a Merlin DAG) interchangeably either in Triton (using `MerlinPythonExecutor`) or on Dask (using `MerlinDaskExecutor`.)


rjzamora:
I suppose this would (partially) intersect with https://github.com/NVIDIA-Merlin/core/issues/70

karlhigley:
Yeah, good point @rjzamora. I would like to be able to do Dask computations across all the Merlin libraries, and also use Merlin graphs to run computations without Dask in some contexts (e.g. in Triton), so I ended up with a somewhat different design.

viswa-nvidia:
arbitration: which initiative is this under ?

nvtabular/workflow/executor.py
```diff
@@ -0,0 +1,247 @@
...
+                self._executor.apply,
+                nodes,
+                additional_columns=additional_columns,
+                capture_dtypes=capture_dtypes,
+                meta=output_dtypes,
+                enforce_metadata=False,
+            )
+        )
+
+    def fit(self, ddf, nodes):
```
oliverholworthy, line 192: 📄 missing nodes in Paramaters docstring here

nvtabular/workflow/executor.py
```diff
@@ -0,0 +1,247 @@
...
+    def __init__(self, client=None):
+        self._executor = MerlinPythonExecutor()
+
+        # Deprecate `client`
+        if client is not None:
+            set_client_deprecated(client, "Workflow")
+
+    def __getstate__(self):
+        # dask client objects aren't picklable - exclude from saved representation
+        return {k: v for k, v in self.__dict__.items() if k != "client"}
```
oliverholworthy, line 144: ❓ I'm wondering where the client attribute is being set on the object (that this code is trying exclude). I don't see a `self.client` in here. Could be something outside this module doing something I suppose. Not suggesting we remove this now since it was here before and to reduce risk it makes sense to keep.

oliverholworthy:
😃 This PR is a great example of separating changes into well defined commits that makes reviewing a refactor like this easy to follow. 🚀  It looks like a great step in the direction toward being able to run these transforms in different modes. 

I imagine we may identify further changes as we try to use this in Systems. In the interest of keeping the changes relatively small, it seems like in a merge-able state to me.

karlhigley:
@viswa-nvidia This PR was opened on the premise that we'd be working on offline batch recs generation in 22.08, as we'd planned before session-based bumped it out of the way. Since we still plan to work on offline batch (albeit later than we'd originally hoped), this PR is still relevant but not tied to one of the pieces of work slated for 22.08.

nvtabular/workflow/executor.py
karlhigley: I think this is leftover from a version before I realized I should use `set_client_deprecated`, so it is likely safe to remove. I know from the process of writing this that NVT tests will fail when saving a Workflow if there's a non-serializable client attribute on this object, so if it's problematic to remove, we'll find out quickly.

oliverholworthy (approved):

karlhigley:
I'm not able to reproduce these test failures locally, even in the `merlin_ci_runner` image. Going to try a re-run 🤷🏻

karlhigley:
rerun tests

karlhigley:
rerun tests

karlhigley:
rerun tests

karlhigley:
rerun tests

karlhigley:
The tests for this keep hanging on the multi-GPU Jenkins machine. Not sure if it's an issue with this PR specifically, or NVTabular PRs in general...

PR #1636, opened 2022-07-27
 Fix `GroupBy` column names when grouping and outputting same column

karlhigley:
In this case, the piece of code that should be connecting the column name with the aggregation name incorrectly tries to join every character of the column name with the name separator, resulting in a column called (for example) `c_u_s_t_o_m_e_r___i_d`. We can avoid that by explicitly checking if we have a string (which is iterable like a list) instead of an actual list.


jperez999 (approved):

radekosmulski:
looks good, will be great to have this 🙂 

Seems some tests are failing, but after approval from @jperez999 it is actually letting me merge?! I am afraid to press the 'Squash and merge' button to see what would happen... But this doesn't seem right?

image

radekosmulski (approved):

karlhigley:
Having updated the branch to the latest `main`, it looks like this PR doesn't work on CPU with Pandas. 😅

PR #1641, opened 2022-08-09
Update the `Categorify` operator to set the domain max correctly

oliverholworthy:
## Goal

Reduce the resulting `int_domain.max` property by one on a ColumnSchema after transforming with `Categorify`. To match the data correctly.

## Motivation / Context

This PR was motivated by work on https://github.com/NVIDIA-Merlin/Merlin/issues/479

We are using the `domain.max` to compute the vocab size / cardinality when creating embedding tables in Merlin Models. This off-by-one error is resulting in some confusion when creating the correct shape embedding dimensions from pretrained embedding data.

## Example

```python
import uuid
import pandas as pd

from merlin.io import Dataset
from merlin.transforms import Workflow
from merlin.transforms.ops import Categorify

df = pd.DataFrame({"id": [str(uuid.uuid4()) for _ in range(2)]})
dataset = Dataset(df)
```

`dataset`

```
                                     id
0  fc5f18c4-919f-4496-9209-1ae34aa4230d
1  738f873f-5fa7-4345-9daa-b1f714c9f1aa
```

`dataset.schema`

```
  name tags   dtype  is_list  is_ragged
0   id   ()  object    False      False
```

After the Categorify op, these ids are transformed to integers {1, 2} with 0 reserved for out-of-vocabulary. So we have a cardinality of 3 (including the zero). 

```python
workflow = Workflow(["id"] >> Categorify())
transformed_dataset = workflow.fit_transform(dataset)
```

`transformed_dataset`:

```
   id
0   2
1   1
```

`transformed_dataset.schema`

```
  name                tags  dtype  is_list  is_ragged properties.num_buckets  \
0   id  (Tags.CATEGORICAL)  int64    False      False                   None   

   properties.freq_threshold  properties.max_size  properties.start_index  \
0                          0                    0                       0   

               properties.cat_path  properties.domain.min  \
0  .//categories/unique.id.parquet                      0   

   properties.domain.max properties.domain.name  \
0                      3                     id   

   properties.embedding_sizes.cardinality  \
0                                       3   

   properties.embedding_sizes.dimension  
0                                    16  
```

However, with the current implementation the int_domain.max value after the transform in this example is 3. This is the same value as the cardinality. However, the maximum integer value is one less than the cardinality here which is 2.


karlhigley (approved):

oliverholworthy:
rerun tests

karlhigley:
rerun tests

karlhigley:
rerun tests

PR #1651, opened 2022-08-15
Update `versioneer` from 0.19 to 0.23

oliverholworthy:
### Goals :soccer:

Restore editable install support with latest version of `setuptools` (without requiring setting any environment variables to activate legacy support).

### Implementation Details :construction:

Ran the following

```shell
pip install --upgrade versioneer
versioneer install
```

There are a few changes since we last configured versioneer (0.19).

The relevant change we're after to fix the editable install is in release 0.23 which added a patch for compatibility with the new  `setuptools` release.

### Testing Details :mag:

Checked that the following works locally:

```shell
pip install -e .
```


karlhigley (approved):

oliverholworthy:
I didn't intend to remove the  `ci` label here. This was me trying to add the `chore` label. Demonstrates something unusual about the way the label code works in Github.

image

karlhigley:
Eh, it was debatable label anyway 🤷🏻

PR #1673, opened 2022-09-08
Remove poetry config

benfred:
We added poetry config to allow nspect to scan our dependencies.

This poetry config has caused issues before with customers, with dependencies not being up to date - since our main dependencies are specified in the requirements.txt files. Also the version number included in the poetry config is months out of date, and still shows a pre 1.0 version.

Since we are now going a different route for nspect scanning, remove this config.


mikemckiernan (approved):
I haven't heard of poetry being used with this project, so removing the content seems right to me.  Thanks.

PR #1675, opened 2022-09-08
chore: Add pybind11 as a tox requirement

mikemckiernan:
To avoid asking developers to install pybind11
on their host or mixing virtual envs with tox,
it seems simplest to make pybind11 a requirement
before tox attempts to create any environment.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,6 +3,7 @@
 
 [tox]
 envlist = python3.8,test-gpu,test-cpu
+requires = pybind11
 
 [testenv]
 commands =


karlhigley (approved):

mikemckiernan:
rerun tests

benfred:
I'm totally ok with this change - but we are specifying pybind11 as a build dependency https://github.com/NVIDIA-Merlin/NVTabular/blob/d3c7aa9afed9a41bbd71831bd441f1dc243bbddc/pyproject.toml#L5, and it *should* be automatically installed by pip when building nvt . 

What was the error you were seeing?

PR #1685, opened 2022-09-30
Fix joint Categorify with list columns

rjzamora:
Closes https://github.com/NVIDIA-Merlin/NVTabular/issues/1676

The intial concatenation step within the logic to jointly-encode columns was not accounting for the possibility that a subset of the columns may be list dtypes. This PR itrodces a simple fix to flatten list columns before concatenation.


rjzamora:
cc @bschifferer

karlhigley (approved):

PR #1686, opened 2022-10-03
Fix data path for reading csv files in the getting started 01-Download-Convert nb

rnyak:
[01-Download-Convert.ipynb](https://github.com/NVIDIA-Merlin/NVTabular/blob/main/examples/getting-started-movielens/01-Download-Convert.ipynb) gives error because the data path is not properly set in `df_lib.read_csv(os.path.join(INPUT_DATA_DIR, "movies.csv"))` .  It should be like below:

`df_lib.read_csv(os.path.join(INPUT_DATA_DIR, "ml-25m", "movies.csv"))`

### DIFF
--- a/examples/getting-started-movielens/01-Download-Convert.ipynb
+++ b/examples/getting-started-movielens/01-Download-Convert.ipynb
@@ -1,488 +1,488 @@
 {
-    "cells": [
-        {
-            "cell_type": "code",
-            "execution_count": null,
-            "metadata": {},
-            "outputs": [],
-            "source": [
-                "# Copyright 2021 NVIDIA Corporation. All Rights Reserved.\n",
-                "#\n",
-                "# Licensed under the Apache License, Version 2.0 (the \"License\");\n",
-                "# you may not use this file except in compliance with the License.\n",
-                "# You may obtain a copy of the License at\n",
-                "#\n",
-                "#     http://www.apache.org/licenses/LICENSE-2.0\n",
-                "#\n",
-                "# Unless required by applicable law or agreed to in writing, software\n",
-                "# distributed under the License is distributed on an \"AS IS\" BASIS,\n",
-                "# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\n",
-                "# See the License for the specific language governing permissions and\n",
-                "# limitations under the License.\n",
-                "# =============================================================================="
-            ]
-        },
-        {
-            "cell_type": "markdown",
-            "metadata": {},
-            "source": [
-                "<img src=\"http://developer.download.nvidia.com/compute/machine-learning/frameworks/nvidia_logo.png\" style=\"width: 90px; float: right;\">\n",
-                "\n",
-                "# Getting Started MovieLens: Download and Convert\n",
-                "\n",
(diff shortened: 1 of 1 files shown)


examples/getting-started-movielens/01-Download-Convert.ipynb
```diff
@@ -1,488 +1,488 @@
...
+       "4                                       Comedy  "
+      ]
+     },
+     "execution_count": 5,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "movies = df_lib.read_csv(os.path.join(INPUT_DATA_DIR, \"ml-25m\", \"movies.csv\"))\n",
```
benfred, line 212: I see that @jperez999 removed the "ml-25m" part of the directory in https://github.com/NVIDIA-Merlin/NVTabular/pull/1622/files . IIRC, this was to make the integration tests work appropriately (?)

However, running this notebook in jupyter lab myself (after deleting the `~/nvt-examples/movielens` directory), and it is throwing errors because it can't find the file. This change seems like it fixes that error, but  I'm wondering if we will need to update the CI infrastructure to match? @jperez999 what do you think?

rnyak:
rerun tests

examples/getting-started-movielens/01-Download-Convert.ipynb
jperez999: so if ml-25m is supposed to be in the directory path... it should be part of the INPUT_DATA_DIR env var.

examples/getting-started-movielens/01-Download-Convert.ipynb
```diff
@@ -1,488 +1,488 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "INPUT_DATA_DIR = os.environ.get(\n",
+    "    \"INPUT_DATA_DIR\", os.path.expanduser(\"~/nvt-examples/movielens/data/\")\n",
```
jperez999, line 75: Change should be made here. That way it can be injected into using our testing apparatus.

jperez999 (changes requested):

examples/getting-started-movielens/01-Download-Convert.ipynb
rnyak: @jperez999 not sure I understood that comment how I will fix INPUT_DATA_DIR. the issue occurs after  extracting the zip file because it automatically generates a new folder `ml-25m` and puts the csv files in there. the current code ignores `ml-25m` folder so it cannot find the datasets.

but if we add `ml-25m` to data read path,  this works:
`
df_lib.read_csv(os.path.join(INPUT_DATA_DIR, "ml-25m", "movies.csv"))
`
you say,  it should only have two parts, path and file name, then can we not do like below:

```
df_lib.read_csv(os.path.join(INPUT_DATA_DIR, "ml-25m/movies.csv"))
```

bschifferer:
@rnyak I think we can close the PR. We will remove the example from the NVTabular repository and we will review the full movielens/criteo examples in the Merlin repository

PR #1687, opened 2022-10-03
[WIP] Refactor Categorify code to use Dask's ACA function

rjzamora:
In order to improve graph-optimization opportunities in Dask, I propose that the `Categorify` code be refactored to use [`apply_concat_apply`](https://github.com/dask/dask/blob/bdb21aedecb36e755d49eadc5e3873192c975411/dask/dataframe/core.py#L6357).

NOTE: Doing this should also make it easier to split the unique groups between multiple partitions (and avoid OOM errors when high-cardinality groups are written to disk or merged to a local partition). Multi-partition category support will be addressed in a follow-up PR.

**TODO**:

- [ ] Use benchmarks to check for performance regressions
- [x] File an issue in Dask to add a `concat_func` kwarg to aca (currently need to "hack" the `Layer` attribute to handle `on_host`) (See: https://github.com/dask/dask/issues/9539)

### DIFF
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -17,7 +17,6 @@ import os
 import warnings
 from copy import deepcopy
 from dataclasses import dataclass
-from operator import getitem
 from pathlib import Path
 from typing import Optional, Union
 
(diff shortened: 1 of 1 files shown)


rjzamora:
**Update**: I spent some time benchmarking, and I now suspect ACA is not the best approach for `Categorify`. Currently heading in a slightly different direction.

rjzamora:
Closing this in faver of https://github.com/NVIDIA-Merlin/NVTabular/pull/1692 for now.

PR #1689, opened 2022-10-06
Fix test imports

jperez999:
This PR adds an import or skip for the example tests that rely on the tensorflow framework. These tests should not run when tensorflow is not present.


karlhigley:
This looks like it has a fair number of unrelated changes 😅

jperez999:
rerun tests

PR #1692, opened 2022-10-11
Refactor Categorify

rjzamora:
**Note**: When ready, this PR should supersede #1687

## Current Algorithm

The current `Categorify` implementation was not designed for extremely large datasets. It was also not designed to handle categorical columns containing too many unique values to fit in a single `cudf.DataFrame`.  The problem with large-scale datasets is the fact that the `fit` code does not use a proper tree reduction. The problem with high-cardinality columns is the fact that all unique groups must be written to (and read from) disk using a single `cudf.DataFrame`.

In both cases, the `fit` stage of `Categorify` is what limits scalability the most.  The existing `fit` algorithm is as follows:

Screen Shot 2022-10-11 at 11 53 15 AM

1. For each Dask-DataFrame partition, we perform the "Level-1" groupby aggregation for each categorical column. The output of the "Level 1" task is a dictionary of  `cudf/pandas.DataFrame` results. Each dictionary value corresponds to the local groupby result for a distinct categorical-column + column-hash combination. The number of hash-based "splits" to use for each categorical column depends on the `tree_width` argument. In the figure above, there is only one categorical column (for simplicity), and `tree_width=3`.
2. All "Level 1" outputs are combined by a distinct "Level 2" task for each unique categorical-column + column-hash combination. For the case above, where there is a single categorical column, the number of "Level 2" tasks is determined by `tree_width`. (**Potential for OOM**: Large-scale datasets may require that a significant amount of data be aggregated in each of these tasks - Even with a large `tree_width` setting)
3. All "Level 2" tasks for the same categorical column are combined by a distinct "Level 3" task. The number of "Level 3" tasks is equivalent to the number of categorical columns in the system (one in the case above). This task is responsible for writing out the final unique-value summary to disk for a specific categorical column. (**Potential for OOM**: High-cardinality columns will need to fit entirely in memory for these tasks to succeed).

## New Algorithm

This PR modifies the existing `Categorify` algorithm in two important ways: First, we perform a "proper" tree reduction to avoid OOM errors for large-scale datasets. Second, we persist unique values as a directory of parquet files (rather than a **single** parquet file) for high-cardinality columns. The general algorithm for the new `Categorify.fit` is as follows:

Screen Shot 2022-10-11 at 11 53 25 AM

1. Same as step 1 for the *old* algorithm.
2. We perform a seperate tree reduction for each distinct categorical-column + column-hash combination. Note that "Level 2" tasks are split into "Level 2(a, b, c, ...)" tasks (depending on the necessary height of the tree reduction).
3. Rather than combining all "Level 2" tasks into a single parquet file for each categorical column, we use the "Level 2" tasks to create a temporary Dask-DataFrame collection for each categorical column. **Note that this is only done if the temporary collection is expected to contain more than one partition**. If a temporary DataFrame collection **is** created, it is written to disk with `ddf.to_parquet(..., compute=False)`, and the `Delayed` result is used in place of the "Level 2" keys as the "Level 3" task dependency.
4. The "Level 3" task will now (optionally) read the temporary DataFrame collection for a specific categorical columns from disk and then re-write the unique values to disk with a proper index, proper ordering, and a null-value.

**Important limitations**: The new algorithm will currently "fall back" to writing a single Parquet file for "complex" `Categorify` options (e.g. multi-column joint/combo encoding). The number of supported cases can certaibly be improved. However, the primary target for this PR is single-column encoding.


rjzamora:
Note that I am unable to reproduce the `test_horovod_multigpu` CI failure locally.

bench/examples/dask-nvtabular-criteo-benchmark.py
```diff
@@ -34,15 +34,15 @@
...
 from nvtabular import Dataset, Workflow
 from nvtabular import io as nvt_io
 from nvtabular import ops
-from nvtabular.utils import _pynvml_mem_size, device_mem_size, get_rmm_size
 
 
 def setup_rmm_pool(client, pool_size):
     # Initialize an RMM pool allocator.
-    # Note: RMM may require the pool size to be a multiple of 256.
+    # Note: RMM may require the pool size to be a multiple of 256
```
rjzamora, line 45:
```suggestion
    # Note: RMM may require the pool size to be a multiple of 256.
```
Unnecessary change.

nvtabular/ops/categorify.py
```diff
@@ -1299,7 +1611,7 @@ def _encode(
     path,
     df,
     cat_cache,
-    na_sentinel=-1,
+    na_sentinel=0,
```
karlhigley, line 1614: How does this interact with using the 0th position to encode unknown values? I can imagine this either simplifying things or causing potential issues, but I'm not sure which is happening here.

bschifferer:
I am getting warnings after ~60min 

```
2022-12-01 11:43:50,990 - distributed.worker_memory - WARNING - Worker is at 80% memory usage. Pausing worker.  Process memory: 66.92 GiB -- Worker memory limit: 83.60 GiB
2022-12-01 11:43:51,271 - distributed.worker_memory - WARNING - Worker is at 78% memory usage. Resuming worker. Process memory: 65.94 GiB -- Worker memory limit: 83.60 GiB
2022-12-01 11:44:10,006 - distributed.worker_memory - WARNING - Worker is at 85% memory usage. Pausing worker.  Process memory: 71.17 GiB -- Worker memory limit: 83.60 GiB
2022-12-01 11:44:47,166 - distributed.worker_memory - WARNING - Worker is at 80% memory usage. Pausing worker.  Process memory: 66.93 GiB -- Worker memory limit: 83.60 GiB
2022-12-01 11:44:48,579 - distributed.worker_memory - WARNING - Worker is at 79% memory usage. Resuming worker. Process memory: 66.47 GiB -- Worker memory limit: 83.60 GiB
2022-12-01 11:45:04,678 - distributed.worker_memory - WARNING - Worker is at 85% memory usage. Pausing worker.  Process memory: 71.88 GiB -- Worker memory limit: 83.60 GiB
2022-12-01 11:45:08,420 - distributed.worker_memory - WARNING - Worker is at 80% memory usage. Pausing worker.  Process memory: 66.88 GiB -- Worker memory limit: 83.60 GiB
2022-12-01 11:45:09,325 - distributed.worker_memory - WARNING - Worker is at 78% memory usage. Resuming worker. Process memory: 66.02 GiB -- Worker memory limit: 83.60 GiB
2022-12-01 11:45:09,570 - distributed.worker_memory - WARNING - Worker is at 79% memory usage. Resuming worker. Process memory: 66.83 GiB -- Worker memory limit: 83.60 GiB

```

rjzamora:
>I am getting warnings after ~60min

Thats too bad :/

rjzamora:
@bschifferer - Do you have  high-cardinality columns in your data? If so, can you try setting `split_out=2` (or perhaps more if the cardinality is really huge) for the those columns?  Using `split_out>1` is typically a bit slow, but will avoid the need to ever have all uniques for a given column in memory at once.

nvtabular/ops/categorify.py
karlhigley: Huh, yeah, I'm confused too! I don't really understand why that would be configurable either, so it's tempting to remove it. I feel like this could be a Chesterton's Fence situation though. 😅

rjzamora:
**Update**: I am using this PR as a space to explore the necessary changes to approach something similar to the proposal in https://github.com/NVIDIA-Merlin/NVTabular/issues/1748 (since this PR is already re-writing most of Categorify anyway)

nvtabular/ops/categorify.py
```diff
@@ -207,10 +215,40 @@ def __init__(
...
+
+        # Warn user if they set num_buckets without setting max_size or
+        # freq_threshold - This setting used to hash everything, but will
+        # now just use multiple indices for OOV encodings at transform time
+        if num_buckets and not (max_size or freq_threshold):
+            warnings.warn(
+                "You are setting num_buckets without using max_size or "
+                "freq_threshold to restrict the number of distinct "
+                "categories. Are you sure this is what you want?"
+            )
```
rjzamora, line 250: @rnyak - This is one detail I am still unsure about. I really don't want to make it easy for users to "hash everything" (what currently happens when the user **only** sets `num_buckets`).  However, it is still strange for the user to ***only** set `num_buckets` (now for a slightly-different reason).  My issue with the original behavior is that there is really no reason for `Catogorify` to be used at all for the case of "hash everything," because we are doing a lot of unnecessary work that can be accomplished by a simple lambda op.

rnyak: @rjzamora I am ok to take `num_buckets` out of Categorify. So far, we dont know a Merlin user who only wants to use `num_buckets` (IIRC). Not sure anyone wants to really hash entire item catalog only with num_buckets.. then how they will reverse back to original ids is a question mark. but I leave the decision to you and @karlhigley.

rjzamora: >I am ok to take num_buckets out of Categorify

Just to clarify, I'm not really proposing that we remove the `num_buckets` option, but that we slightly change it's meaning to **only** do what it used to do when `max_size` or `freq_threshold` was set:

- **OLD meaning**: We hash everything, **UNLESS** `max_size` or `freq_threshold` is also set (in which case we use `num_buckets` to be the number of hash-buckets to use for infrequent OOV values at transformation time).
-  **NEW meaning**: We **always** take `num_buckets` to be the number of hash-buckets to use for infrequent OOV values at transformation time. Even if neither `max_size` nor `freq_threshold` is set, we will still use a hash+modulo for OOV values at transformation time if `num_buckets>1`.

jperez999 (approved):

PR #1696, opened 2022-11-01
Upload binary wheels for nvtabular

benfred:


jperez999 (approved):

karlhigley:
@jperez999 @benfred Do y'all have any context on what this was supposed to fix? It looks like `bdist_wheel` is building a platform-specific wheel for `linux_x86_64`, which PyPI won't allow us to upload.

PR #1703, opened 2022-11-17
Update metrics keys in example notebook tests

karlhigley:


jperez999 (approved):

oliverholworthy:
> The PR checks are apparently running against different versions than the container build checks, because this fix makes the tests pass in the container but makes them fail here

This seems to be because the Jenkins GPU tests are currently picking up whatever version of merlin models happens to be installed in the CI container (an older version). Instead of installing the latest version.

Perhaps we need to add `-rrequirements/test.txt` to the `test-gpu` config in `tox.ini` ([`requirements/test.txt` has merlin models:](https://github.com/NVIDIA-Merlin/NVTabular/blob/v1.6.0/requirements/test.txt#L28-L29))

Or maybe clearer to add it to the `commands` inside `tox.ini` like we have for core/dataloader?

jperez999:
Yeah my vote is to add this to the tox ini so that we pull in latest version of models.

oliverholworthy (approved):

karlhigley:
rerun tests

karlhigley:
rerun tests

karlhigley:
This PR is blocked on (other) weird failures in the example notebooks 😞

karlhigley:
rerun tests

edknv:
The notebook test failures seem to related to the environment variables not being available in tox, same as here: https://github.com/NVIDIA-Merlin/Merlin/pull/724

PR #1705, opened 2022-11-18
Set `is_ragged=False` if value_count min=max in loader backend

oliverholworthy:
Set `is_ragged=False` if value_count min=max in loader backend.

If https://github.com/NVIDIA-Merlin/core/pull/111 is merged, this will fail because a check has been introduced that if is_ragged=True, that value_count min != max

### DIFF
--- a/nvtabular/loader/backend.py
+++ b/nvtabular/loader/backend.py
@@ -38,14 +38,17 @@ def _augment_schema(
     for col in sparse_names or []:
         cs = schema[col]
         properties = cs.properties
+        is_ragged = not sparse_as_dense
         if sparse_max and col in sparse_max:
             properties["value_count"] = {"min": sparse_max[col], "max": sparse_max[col]}
+            if sparse_max[col] == sparse_max[col]:
+                is_ragged = False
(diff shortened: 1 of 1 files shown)


karlhigley:
@oliverholworthy Does this also need to go into the `dataloader` package?

oliverholworthy:
rerun tests

oliverholworthy:
> @oliverholworthy Does this also need to go into the `dataloader` package?

This augment schema logic is only in NVTabular it seems. Or did you mean: does this functionality need to be moved into the dataloader?

jperez999:
yes it does, otherwise it will break. https://github.com/NVIDIA-Merlin/dataloader/blob/main/merlin/loader/loader_base.py#L105

jperez999:
Then that should be the only changes necessary in the dataloader repo. Sorry for posting earlier seems @oliverholworthy already had that done.

karlhigley:
rerun tests

karlhigley:
Looks like this has some linting issues, but LGTM once those are resolved

karlhigley (approved):

PR #1706, opened 2022-11-18
Allow `tox` tasks to use the `OPAL_PREFIX` env variable

karlhigley:
Following @edknv's work in NVIDIA-Merlin/Merlin#724

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -21,6 +21,8 @@ commands =
     python -m pytest --cov-report term --cov=merlin -rxs tests/unit
 
 [testenv:test-gpu]
+passenv =
+    OPAL_PREFIX
 sitepackages=true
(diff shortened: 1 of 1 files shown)


edknv (approved):

edknv:
We might also need to add `NR_USER` to passenv because the horovod test is not skipped anymore: https://github.com/NVIDIA-Merlin/NVTabular/blob/21117cfc4c113b30036afcb97b6daa5f377996db/tests/unit/loader/test_tf_dataloader.py#L530

karlhigley:
Cherry picked these commits into #1703, since they otherwise block each other from being merged

karlhigley:
Superseded by #1703

PR #1707, opened 2022-11-21
Update NVT operators and workflows to use Merlin dtypes

karlhigley:
Depends on https://github.com/NVIDIA-Merlin/core/pull/170


nvtabular/ops/reduce_dtype_size.py
```diff
@@ -63,14 +64,14 @@ def compute_output_schema(self, input_schema, selector, prev_output_schema=None)
             column = input_schema[column]
 
             dtype = column.dtype
-            if np.issubdtype(column.dtype, np.integer):
+            if column.dtype.elemtype.value == "int":
```
karlhigley, line 67: This would be good to wrap up in a helper method so that calling code doesn't need to know the details of the dtype implementation to check whether or not a dtype is an integer type

karlhigley:
The failing test here is due to changes in the Merlin Models `dataset` package in https://github.com/NVIDIA-Merlin/models/pull/912 that haven't been merged yet

jperez999 (approved):
I think this is good, but we need to wait on core changes to be able to reference the main branch of core with the merlin dtype changes.

PR #1708, opened 2022-11-22
Migrate `nvtabular.inference` to `merlin.legacy.inference`

karlhigley:
This is moving to the `merlin.legacy.inference` package in the Systems repo. That consolidates all the Triton code in one repo and should make it easier for us to figure out when we can remove it completely.

The corresponding PR on the other side is https://github.com/NVIDIA-Merlin/systems/pull/233


tests/unit/test_notebooks.py
```diff
@@ -111,75 +110,6 @@ def _nb_modify(line):
         _run_notebook(tmpdir, notebook_path, _nb_modify)
 
 
-def test_movielens_example(tmpdir):
```
karlhigley, line 114: This does lose test coverage of the Movielens example in this repo. Since that example would now rely on multiple Merlin libraries, I'd propose moving it to the top-level Merlin repo if we want to keep it.

jperez999 (approved):

karlhigley:
Still think removing this code from this library is a good idea, but up in the air where it live, so closing this.

PR #1711, opened 2022-11-22
Migrate the legacy examples to the Merlin repo

karlhigley:
We may (or may not) want to keep these examples but they've overstayed their welcome in the NVTabular repo, which is burdened with the accumulation of a lot of historical cruft. Since some of these examples use inference code that's moving to Systems, it makes more sense for them to live in the Merlin repo (if we want to keep them.)


jperez999 (approved):

karlhigley:
Hmm, haven't been able to reproduce these test failures locally, so...

karlhigley:
rerun tests

karlhigley:
rerun tests

bschifferer:
Thanks @karlhigley - that is a great clean up PR.

We need to update multiple documentation before we can merge it.
- https://github.com/NVIDIA-Merlin/NVTabular/blob/main/examples/README.md
- https://github.com/NVIDIA-Merlin/NVTabular/blob/main/README.md

Let me try to make a proposal to update the READMEs.

@mikemckiernan can you test, if we break any links in our documentations?

bschifferer (changes requested):
Update READMEs + test if we break any links

mikemckiernan:
I added Karl's fork and checked out the `migrate/legacy-examples` branch.  I already have a handful of changes to make the TOC behave reasonably and I still have some broken links to fix.

I'll fix the remaining broken links and then push (to Karl's fork?) on Monday unless someone tells me otherwise.

karlhigley:
@mikemckiernan If you can't push to my fork, I can push to a branch on the main repo that we can share or we can sync on the changes and I can make them on my branch.

Not trying to dump a bunch of new work on you here; trying to find a way to either prompt the right folks to either make some concrete decisions about these examples or move them out of my way. Maintaining and working around these outdated examples isn't free, but we often treat it like it is because that's the status quo. These changes are attempt to remove the status quo as an option from the decision making process—it's fine if we're willing to make the updates to keep them around, but if not, that's fine too.

In either case, I'm happy to make the necessary changes, but I don't always know where to find them. Point me in the right direction?

bschifferer:
Let me know, if I can support fixing the links

karlhigley:
As far as I can tell, these are the places that explicitly link to that path on Github:
Screenshot from 2022-11-28 09-18-17
This only includes absolute URLs; there could be others using relative paths but I haven't figured out how to find them yet.

karlhigley:
@bschifferer Updating the text of the READMEs is a bit outside my wheelhouse, so I'd propose we go ahead and merge the migration PRs and _then_ update the READMEs. We can't really do that work in advance of the PRs being merged, so that's kind of the only way to split it. As far as broken links go, they appear to be largely in the README and the examples themselves, so that might be reasonable thing to do as part of the second set of PRs.

mikemckiernan:
> @mikemckiernan If you can't push to my fork, I can push to a branch on the main repo that we can share or we can sync on the changes and I can make them on my branch.
> 
> Not trying to dump a bunch of new work on you here; trying to find a way to either prompt the right folks to either make some concrete decisions about these examples or move them out of my way. Maintaining and working around these outdated examples isn't free, but we often treat it like it is because that's the status quo. These changes are attempt to remove the status quo as an option from the decision making process—it's fine if we're willing to make the updates to keep them around, but if not, that's fine too.
> 
> In either case, I'm happy to make the necessary changes, but I don't always know where to find them. Point me in the right direction?

A) I don't see this as a bunch of new work.  It's tech debt and I'm glad to play any role in the fixup.

B) I never shared how _I_ check for broken links, so I can't blame anyone for not doing it.  I'll share shortly and either we can all follow along or we'll discover a better way to handle them.

mikemckiernan (approved):
I've made a number of updates in this PR too, but I certainly approve the updates that Karl made to address the duplication and stale information.

bschifferer:
@karlhigley @mikemckiernan can we merge the PR?

bschifferer (approved):

bschifferer:
rerun tests

PR #1717, opened 2022-11-28
Split test suite into `unit`, `integration`, `notebook`, and `dataset`

karlhigley:


karlhigley:
I still think is a good idea, but I think we can do it a bit more incrementally.

PR #1720, opened 2022-11-30
Handle data loader as an iterator

oliverholworthy:
The dataloader has been updated to act as an iterator in https://github.com/NVIDIA-Merlin/dataloader/pull/65 

This means that calling `iter(loader)` no longer resets the state of the loader. To reset the state `.stop()` can be used.

This PR updates the tests and notebooks to handle this


mikemckiernan (approved):
Thanks very much for implementing the update quickly.

PR #1721, opened 2022-11-30
swtiching to rmm managed memory instead of pool size

jperez999:
This PR switches to rmm managed memory, from using a defined pool size. This is to avoid failures in integration tests in scaling criteo notebooks. The integration test in question is: `tests/integration/test_criteo.py` for hugectr specifically.


jperez999:
closing, this file was moved to another repository.

PR #1723, opened 2022-12-06
Use pre-commit for lint checks in GitHub Actions Workflow

oliverholworthy:
Moves lint checks to separate lint GitHub Actions workflow that uses pre-commit. This ensures consistency of checks between local pre-commit hooks and in CI.


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,14 @@
...
+  push:
+    branches: [main]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+      - uses: pre-commit/action@v2.0.3
```
nv-alaiacano, line 14: This seems to be deprecated and the recommendation is to use https://pre-commit.ci, which appears to be a service that maybe we don't want to sign up for right now. Plus the repo is still active so 🤷‍♂️.

However:

* `pre-commit/action@v2.0.3` uses `https://github.com/pre-commit/pre-commit-hooks` version v4.4.0 [here](https://github.com/pre-commit/action/blob/main/.pre-commit-config.yaml)
* pre-commit-hooks 4.4.0 uses flake8 version 6.0.0 [here](https://github.com/pre-commit/pre-commit-hooks/blob/a41c4b94b220171e928ff3e6c9bef34e71267f46/.pre-commit-config.yaml#L35-L38)

so we should probably use that same version in our `.pre-commit-config.yaml` at the very least. Or perhaps copy their `pre-commit-config.yaml` into our repo(s)?

nv-alaiacano (approved):

.github/workflows/lint.yaml
oliverholworthy: If we're concerned about the deprectation notices, we could copy the contents of this file

https://github.com/pre-commit/action/blob/v2.0.3/action.yml

into our lint workflow to run `pre-commit run --all-files` ourselves if that makes things clearer what's going on here?

PR #1724, opened 2022-12-06
Address virtual developer review feedback

mikemckiernan:
- Remove the temporary warning that the SMX moved to the Merlin repository.
- Remove repo link from README and add an Additional Resources link to the NVT developer.nvidia.com page.
- Update page-level TOC in cloud integration file. Need SME input and check if we want to update for NVT or move to the Merlin repo and expand the scope to all of Merlin.


mikemckiernan:
rerun tests

PR #1725, opened 2022-12-09
Remove echo from command in tox.ini

oliverholworthy:
Removes echo command from tox.ini config. 

- This breaks with the recent release of tox 4
- Can be made to work with `allowlist_externals`. However, we don't really need this here since the output of each command is printed


nv-alaiacano (approved):

oliverholworthy:
rerun tests

PR #1727, opened 2022-12-09
Update package requires_dist to remove extras that are not required

oliverholworthy:
- Update requires_dist of package to enable us to publish the package again to pypi.
- Keeps only the `gpu` dependencies as extra requires
  - Updating these to use the rapids pip package names

## Motivation

- [The last two releases failed to publish to pypi due to an error in our package requires_dist](https://github.com/NVIDIA-Merlin/NVTabular/actions/runs/3528066078/jobs/5918246198)
  - `Invalid value for requires_dist. Error: Can't have direct dependency`
  - The issue is with the dependencies that are specified as URLs to git repos (e.g. sphinx-multiversion)

### Before

```
['merlin-core (>=0.2.0)',
 'merlin-dataloader (>=0.0.2)',
 'scipy',
 "merlin-core (>=0.2.0) ; extra == 'cpu'",
 "merlin-dataloader (>=0.0.2) ; extra == 'cpu'",
 "scipy ; extra == 'cpu'",
 "cpplint (>=1.5) ; extra == 'dev'",
 "codespell ; extra == 'dev'",
 "interrogate (==1.5.0) ; extra == 'dev'",
 "black (==22.6.0) ; extra == 'dev'",
 "flake8 (==3.9.2) ; extra == 'dev'",
 "isort (==5.10.1) ; extra == 'dev'",
 "pylint (==2.7.4) ; extra == 'dev'",
 "bandit (==1.7.4) ; extra == 'dev'",
 "flake8-nb (==0.5.2) ; extra == 'dev'",
 "Sphinx (<3.6) ; extra == 'docs'",
 "jinja2 (<3.1) ; extra == 'docs'",
 "markupsafe (==2.0.1) ; extra == 'docs'",
 "sphinx-multiversion @ git+https://github.com/mikemckiernan/sphinx-multiversion.git ; extra == 'docs'",
 "sphinxcontrib-copydirs @ git+https://github.com/mikemckiernan/sphinxcontrib-copydirs.git ; extra == 'docs'",
 "sphinx-external-toc (<0.4) ; extra == 'docs'",
 "sphinx-rtd-theme ; extra == 'docs'",
 "natsort ; extra == 'docs'",
 "myst-nb ; extra == 'docs'",
 "linkify-it-py ; extra == 'docs'",
 "merlin-core (>=0.2.0) ; extra == 'gpu'",
 "merlin-dataloader (>=0.0.2) ; extra == 'gpu'",
 "scipy ; extra == 'gpu'",
[...]

PR #1728, opened 2022-12-09
Add check for urls present in requires_dist

oliverholworthy:
- [We failed to publish the pypi pacakge in the last release due to invalid values in our package requires_dist](https://github.com/NVIDIA-Merlin/NVTabular/actions/runs/3528066078/jobs/5918246198)
- This PR adds a check to our build process to return an error if we have urls in our package `requires_dist`


oliverholworthy:
I'm expecting the `CPU CI / build` to fail on this new check. And then hopefully pass after #1727 is merged in

mikemckiernan (approved):
You're welcome to request a review from someone more skilled in CI, but I certainly approve the idea of adding a test after we've found a problem.  Thanks.

oliverholworthy:
Updated with an error message if this fails. And moved the check to a dedicated script in `./ci/check_dist.sh`

Example Failure: https://github.com/NVIDIA-Merlin/NVTabular/actions/runs/3697897020/jobs/6263443772#step:7:86

image

PR #1736, opened 2022-12-19
Fix max_size behavior in Categorify

rjzamora:
Closes https://github.com/NVIDIA-Merlin/NVTabular/issues/1735

Changes both combo- and joint-encoding logic to increase the null-value count to include any categories that will be dropped to satisfy the `max_size` argument.

**Warning**: This change means the 0th row in `unique.*.parquet` will no longer correspond  to actual null-value count, but to the number of values in the dataset that *would* be mapped to the "null category" at transformation time (including **both** null and low-frequency values).  cc @rnyak - Please confirm that this is the expected behavior.

### DIFF
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -1071,6 +1071,7 @@ def _combo_encode(df, name_size_multi: str, col_selector: ColumnSelector, option
     # Combo-encoding utility (used by _write_uniques)
 
     # Account for max_size and num_buckets
+    add_nulls = 0
     if options.max_size:
         max_emb_size = options.max_size
         if isinstance(options.max_size, dict):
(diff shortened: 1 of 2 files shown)


karlhigley:
@rjzamora @rnyak Is something we should move forward with now? Do we wait until we've settled the issues raised in the [proposal re: categorify encoding issues and reverse mapping](https://docs.google.com/document/d/1JE5JGK-vHEaYzkvqZGWU-lzpCKdP0oN4gq840JxPCg4/edit)?

karlhigley (approved):

rjzamora:
>Is something we should move forward with now? Do we wait until we've settled the issues raised in the [proposal re: categorify encoding issues and reverse mapping](https://docs.google.com/document/d/1JE5JGK-vHEaYzkvqZGWU-lzpCKdP0oN4gq840JxPCg4/edit)?

I'd probably defer to @rnyak and yourself.

I'm not personally up-to-date on the discussion you cited, but I'll try to catch up. I do think it is important that we establish the connection and separation between the the unique-value statistics and the encoding options (in addition to the actual encoding defaults and practices).

rjzamora:
It seems like there is some consensus in the offline discussion that we should be reserving distinct indices for null and out-of-vocabulary (OOV) values. In the latest plan (as of my writing this), we will also reserve index 0 for padding.  Given the direction of that discussion, I suggest we hold off on merging this particular PR.

If we ultimately commit ourselves to the current proposal, I suggest we revise the Categorify logic in a separate PR (or two).  It probably makes sense to start by designing a comprehensive set of tests that cover the desired behavior (including the structure of the unique-value files) for the various encoding options.

karlhigley:
In order to keep this from getting merged prematurely, I'll convert it to a draft then.

rnyak:
this should be solved as part of this ticket https://github.com/NVIDIA-Merlin/NVTabular/issues/1748

rjzamora:
Closing this since the solution will need to align with https://github.com/NVIDIA-Merlin/NVTabular/issues/1748

Issue #1737, opened 2022-12-20
[BUG] LambdaOp may fail to deserialize if the processing function is declared in another module


willb:
**Describe the bug**

By default, Cloudpickle serializes functions and modules by reference. Deserializing workflows that contain user functions declared in other modules can fail if the Python source files for those modules are not available. This means that serializing a workflow and publishing it in a Docker container requires either (1) that all user source files are published in the same paths in the container as they were when the workflow was serialized or (2) that user modules are not serialized by reference.

**Steps/Code to reproduce bug**

Assume that `identity.py` contains the identity function, called `identity`.  

```python
with open("identity.py", "w") as of:
    of.write("""
def identity(col):
    return col

""")
```

Then run this code:

```python
import nvtabular as nvt
import identity

wf = nvt.Workflow(
    ["col_a"] >> nvt.ops.LambdaOp(identity.identity)
)

wf.save("identity-workflow")
```

This serialized workflow will fail to deserialize if `identity.py` is not available in the same location, e.g., in a Docker container or on another host.

```python
import sys
import os

del sys.modules["identity"]
os.unlink("identity.py")

try:
    wf_ref = nvt.Workflow.load("identity-workflow")
except ModuleNotFoundError as mnfe:
    print("Failed to load workflow")
    print(str(mnfe))

```

**Expected behavior**

Serialized workflows should be largely self-contained and resilient to missing source files containing user transformation code.

**Environment details (please complete the following information):**
 - Environment location:  Docker
 - Method of NVTabular install: Docker
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used:
 
`docker run -v /home/willb/devel/nvtabular-sandbox:/workspace/host --runtime=nvidia --rm -it -p 8888:8888 -p 8797:8787 -p 8796:8786 --ipc=host --cap-add SYS_NICE nvcr.io/nvidia/merlin/merlin-training:22.03 /bin/bash`

**Additional context**

[This notebook](https://gist.github.com/willb/8e99c75b0544065e81bd0b04eed9c1e0) demonstrates the minimal reproducer as well as a workaround.

I expect to submit a patch shortly that will enhance `Workflow.save` in two ways:

1. to allow clients to provide explicit direction that certain modules should be serialized by value, and 
2. to allow clients to request optional automatic identification of such modules.

nv-alaiacano:
Closing because this was resolved.

PR #1741, opened 2022-12-23
Add support for serializing modules involved in LambdaOp execution by value

willb:
These commits address issue #1737 in two ways:

1. by allowing users to direct `Workflow.save` to serialize an explicit list of named modules by value, and
2. by allowing users to direct `Workflow.save` to use a heuristic to automatically infer the external modules involved in a workflow.

Taken together, these commits will make it easier to serialize workflows that are resilient to execution in environments without source files for all of the modules that were available when the workflow was created, e.g., Docker containers.

The main contribution is the addition of a `modules_byvalue` parameter to `Workflow.save`.  If this is passed an explicit list of modules, `Workflow.save` will direct `cloudpickle` to serialize these modules by value.  If it is passed the string `"auto"`, `Workflow.save` will employ a [heuristic approach that will be useful for many real-world uses of LambdaOp](https://github.com/NVIDIA-Merlin/NVTabular/commit/1e82777ba7adf6bec7d5bc59d882d2b258193802).

I believe that it would be harmless and useful to make `"auto"` the default but have not done so in this PR.

For more background on some of the tradeoffs that would be involved in making the heuristic more precise, please see [this blog post](https://chapeau.freevariable.com/2017/12/module-frontier.html). I do not believe that many of the cases I identified in that post (explicit imports within functions, list comprehensions, etc.) are likely to present in realistic NVTabular LambdaOp client code, and thus took a relatively simple approach.


karlhigley (approved):

karlhigley:
The CI tests on this PR appear to hang during `test_lambda.py` on GPU, which seems likely to be related to these changes but is somewhat difficult to troubleshoot without an actual error message.

willb:
> The CI tests on this PR appear to hang during `test_lambda.py` on GPU, which seems likely to be related to these changes but is somewhat difficult to troubleshoot without an actual error message.

Thanks, @karlhigley — I’ll try and figure out what’s going on here.

willb:
Thanks, @karlhigley!

karlhigley:
Thank you, @willb! Sorry it took us so long to get it merged, but this looks great. We literally ran into this problem again today, so the timing is fortuitous.

PR #1742, opened 2022-12-29
Fix output error occurring due to  check if it is a dict or not

rnyak:
The Transformers4Rrec [end-to-end example](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/main/examples/end-to-end-session-based/02-End-to-end-session-based-with-Yoochoose-PyT.ipynb) gives an error `Failed to process the request(s) for model 'mymodel_pt', message: ValueError: output of the forward function should be a dict` when we send request to TIS. the reason for that is due to recent changes in the prediction tasks' outputs in TF4Rec repo (see https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/546). This PR provides a quick fix to avoid such error. However, for long term solution Oliver's draft PR (see https://github.com/NVIDIA-Merlin/systems/pull/252) might be preferable.

### DIFF
--- a/nvtabular/inference/triton/model/model_pt.py
+++ b/nvtabular/inference/triton/model/model_pt.py
@@ -157,9 +157,7 @@ class TritonPythonModel:
 
                 # Call forward function to get the predictions
                 # Forward function should return a dict with the "predictions" bucket
-                out = self.model(input_dict, training=False)
-                # Get the predictions from the out
-                pred = out
+                pred = self.model(input_dict, training=False)
                 if pred is None:
                     raise KeyError(
                         "output of the forward function should have a bucket named as predictions"
--- a/nvtabular/inference/triton/model/model_pt.py
+++ b/nvtabular/inference/triton/model/model_pt.py
@@ -158,11 +158,11 @@ class TritonPythonModel:
                 # Call forward function to get the predictions
                 # Forward function should return a dict with the "predictions" bucket
                 out = self.model(input_dict, training=False)
-                if not isinstance(out, dict):
-                    raise ValueError("output of the forward function should be a dict")
+                # if not isinstance(out, dict):
+                # raise ValueError("output of the forward function should be a dict")
 
                 # Get the predictions from the out
-                pred = out.get("predictions")
+                pred = out
                 if pred is None:
                     raise KeyError(
                         "output of the forward function should have a bucket named as predictions"
--- a/nvtabular/inference/triton/model/model_pt.py
+++ b/nvtabular/inference/triton/model/model_pt.py
@@ -158,9 +158,6 @@ class TritonPythonModel:
                 # Call forward function to get the predictions
                 # Forward function should return a dict with the "predictions" bucket
                 out = self.model(input_dict, training=False)
-                # if not isinstance(out, dict):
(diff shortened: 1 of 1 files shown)


nvtabular/inference/triton/model/model_pt.py
```diff
@@ -158,11 +158,8 @@ def execute(self, requests):
                 # Call forward function to get the predictions
                 # Forward function should return a dict with the "predictions" bucket
                 out = self.model(input_dict, training=False)
```
karlhigley, line 160:
```suggestion
                pred = self.model(input_dict, training=False)
```

nvtabular/inference/triton/model/model_pt.py
```diff
@@ -158,11 +158,8 @@ def execute(self, requests):
...
-                if not isinstance(out, dict):
-                    raise ValueError("output of the forward function should be a dict")
-
                 # Get the predictions from the out
```
karlhigley, line 161:
```suggestion
```

nvtabular/inference/triton/model/model_pt.py
```diff
@@ -158,11 +158,8 @@ def execute(self, requests):
...
-                pred = out.get("predictions")
+                pred = out
```
karlhigley, line 162:
```suggestion
```

karlhigley (approved):
Looks fine but could be cleaned up a bit. Left suggestions that can be merged directly from the comments.

karlhigley:
The code modified by https://github.com/NVIDIA-Merlin/systems/pull/252 was never intended to be used directly—it was added to Systems as a starting point for refactoring toward the operator API, and some parts of the old code still exist there. If something is using the legacy serving code (e.g. `export_pytorch_ensemble`) from Systems instead of from NVTabular, it should be reverted to using the NVT version which is mostly frozen in order avoid regressions where we have insufficient test coverage.

Making minor fixes to the NVT code in order to keep it working with the rest of the Merlin ecosystem is fine if needed, but we've been avoiding making any major changes there. The corresponding code in Systems is open to major changes and therefore unstable; it's also highly likely to be removed in a future Systems version.

karlhigley:
See https://github.com/NVIDIA-Merlin/systems/pull/265 for links to the open PRs related to legacy (non-Torchscript Python) serving for PyTorch. Possible that the fixes in https://github.com/NVIDIA-Merlin/systems/pull/252 should be applied to NVTabular instead?

sararb:
> See [NVIDIA-Merlin/systems#265](https://github.com/NVIDIA-Merlin/systems/pull/265) for links to the open PRs related to legacy (non-Torchscript Python) serving for PyTorch. Possible that the fixes in [NVIDIA-Merlin/systems#252](https://github.com/NVIDIA-Merlin/systems/pull/252) should be applied to NVTabular instead?

Thank you for the context @karlhigley, with Ronay we weren't sure if the legacy function will be included in systems or not.   So It makes sense to me to include the changes proposed by Oliver in https://github.com/NVIDIA-Merlin/systems/pull/265 in NVTabular instead (as a separate PR from this quick fix).

sararb (approved):

PR #1743, opened 2023-01-03
Replace `flake8` with `ruff` (which is equivalent but faster)

karlhigley:
`ruff` is a very fast Python linter written in Rust that covers the flake8 rules in addition to a bunch of other popular linters. This sets up a config for it that covers the existing flake8 config, and adds additional linters while generating minimal errors with the existing code base. Since `pylint` is much slower than `ruff` (and I suspect slower than `flake8` too), this config executes the `ruff` checks first in order to fail fast.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,14 +5,14 @@ repos:
     hooks:
       - id: absolufy-imports
   - repo: https://github.com/timothycrosley/isort
-    rev: 5.10.1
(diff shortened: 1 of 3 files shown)


karlhigley:
Hmm, looks like we're running the pre-commit hooks on all files in the `lint` action, and it doesn't seem to be picking up settings from `pyproject.toml`.

jperez999 (approved):

PR #1751, opened 2023-01-19
Update NVTabular for `ColumnSchema` API changes from Merlin Core

karlhigley:
Since `is_list` and `is_ragged` have become derived properties computed from the shape, it's no longer possible to directly set them from the constructor. They can be smuggled in through the properties, after which they'll be used to determine an appropriate shape that results in the same `is_list` and `is_ragged` values on the other side.

(This is a first step toward capturing and using more comprehensive shape information, with the goal of putting `Shape` in place while breaking as little as possible. There will be subsequent changes to directly capture more shape information, but this gets us part-way there.)

Depends on https://github.com/NVIDIA-Merlin/core/pull/195, https://github.com/NVIDIA-Merlin/models/pull/957

### DIFF
--- a/nvtabular/ops/list_slice.py
+++ b/nvtabular/ops/list_slice.py
@@ -129,6 +129,8 @@ class ListSlice(Operator):
 
     def _compute_dtype(self, col_schema, input_schema):
         col_schema = super()._compute_dtype(col_schema, input_schema)
+        if self.pad:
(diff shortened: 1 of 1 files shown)


karlhigley:
The tests won't pass on this until after changes get merged to Core and Models (since some of the examples depend on the synthetic data generator.)

jperez999 (approved):

PR #1753, opened 2023-01-25
Bump Merlin Core dependency to 0.7.0

karlhigley:
NVT now depends on executors, which only started being available in Core in 0.7.0. This dependency mismatch is causing not infrequent problems for users trying to install NVT.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,3 +1,3 @@
-merlin-core>=0.2.0
+merlin-core>=0.7.0
 merlin-dataloader>=0.0.2
 scipy


karlhigley:
@jperez999 It looks like this version didn't get bumped previously because there's no version of `s3fs` that depends on `fsspec 2022.5.0`:
```
    merlin-core 0.7.0 depends on fsspec==2022.5.0
    s3fs 2022.7.0 depends on fsspec==2022.7.0
    merlin-core 0.7.0 depends on fsspec==2022.5.0
    s3fs 2022.3.0 depends on fsspec==2022.3.0
```
Can we move forward from `fsspec 2022.5.0` at this point?

karlhigley:
Bumping the dependencies in Core seems to work, but I might be missing something about why we had `fsspec` pinned in the first place: https://github.com/NVIDIA-Merlin/core/pull/201

jperez999 (approved):

PR #1754, opened 2023-01-25
Use `cibuildwheel` to build binary distributions for PyPI

karlhigley:


karlhigley:
It doesn't look like the CPU CI action ran after these changes, which seems...bad.

karlhigley:
With `cibuildwheel` and the current config, the dist folder ends up looking like this, with a variety of Linux wheels (but no Mac or Windows wheels):
gnome-shell-screenshot-kmoiu

jperez999 (approved):

PR #1760, opened 2023-02-13
Clean up the way shapes are computed and specified

karlhigley:
This replaces shapes specified implicitly via the `is_list`/`is_ragged` flag or via the `value_count` property with more explicitly computed and specified shapes.

Depends on NVIDIA-Merlin/core#215

### DIFF
--- a/nvtabular/loader/backend.py
+++ b/nvtabular/loader/backend.py
@@ -26,7 +26,7 @@ def _augment_schema(
     padded_cols=None,
     padded_lengths=None,
     pad=False,
-    batch_size=0
+    batch_size=0,
 ):
     labels = [labels] if isinstance(labels, str) else labels
     for label in labels or []:
@@ -38,7 +38,7 @@ def _augment_schema(
 
     for col in padded_cols or []:
         cs = schema[col]
-        dims = Shape(((1,batch_size), None))
+        dims = Shape(((1, batch_size), None))
 
         if not cs.shape.dims[1].is_unknown:
             dims = dims.with_dim(1, cs.shape.dims[1])
@@ -47,13 +47,9 @@ def _augment_schema(
             dims = dims.with_dim_min(1, padded_lengths[col])
         if padded_lengths and col in padded_lengths:
             dims = dims.with_dim_max(1, padded_lengths[col])
-        
+
         schema[col] = ColumnSchema(
-            name=cs.name,
-            tags=cs.tags,
-            dtype=cs.dtype,
-            properties=cs.properties,
-            dims=dims
+            name=cs.name, tags=cs.tags, dtype=cs.dtype, properties=cs.properties, dims=dims
         )
 
     return schema
--- a/nvtabular/loader/backend.py
+++ b/nvtabular/loader/backend.py
@@ -40,6 +40,8 @@ def _augment_schema(
         properties = cs.properties
         if sparse_max and col in sparse_max:
             properties["value_count"] = {"max": sparse_max[col]}
+        if sparse_as_dense:
+            properties["value_count"]["min"] = properties["value_count"]["max"]
         schema[col] = ColumnSchema(
             name=cs.name,
             tags=cs.tags,
--- a/nvtabular/loader/backend.py
+++ b/nvtabular/loader/backend.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 
+from merlin.dtypes.shape import Shape
 from merlin.schema import ColumnSchema, Tags
 
 
@@ -22,9 +23,10 @@ def _augment_schema(
     cats=None,
     conts=None,
     labels=None,
-    sparse_names=None,
-    sparse_max=None,
-    sparse_as_dense=False,
+    padded_cols=None,
+    padded_lengths=None,
+    pad=False,
+    batch_size=0
 ):
     labels = [labels] if isinstance(labels, str) else labels
     for label in labels or []:
@@ -34,21 +36,24 @@ def _augment_schema(
     for label in conts or []:
         schema[label] = schema[label].with_tags(Tags.CONTINUOUS)
 
-    # Set the appropriate properties for the sparse_names/sparse_max/sparse_as_dense
-    for col in sparse_names or []:
+    for col in padded_cols or []:
         cs = schema[col]
-        properties = cs.properties
-        if sparse_max and col in sparse_max:
-            properties["value_count"] = {"max": sparse_max[col]}
-        if sparse_as_dense:
-            properties["value_count"]["min"] = properties["value_count"]["max"]
+        dims = Shape(((1,batch_size), None))
+
+        if not cs.shape.dims[1].is_unknown:
+            dims = dims.with_dim(1, cs.shape.dims[1])
+
+        if pad:
+            dims = dims.with_dim_min(1, padded_lengths[col])
+        if padded_lengths and col in padded_lengths:
+            dims = dims.with_dim_max(1, padded_lengths[col])
+        
         schema[col] = ColumnSchema(
             name=cs.name,
             tags=cs.tags,
             dtype=cs.dtype,
-            is_list=True,
-            is_ragged=not sparse_as_dense,
-            properties=properties,
+            properties=cs.properties,
+            dims=dims
         )
 
     return schema
--- a/nvtabular/ops/groupby.py
+++ b/nvtabular/ops/groupby.py
@@ -18,6 +18,7 @@ import numpy
 from dask.dataframe.utils import meta_nonempty
 
 from merlin.core.dispatch import DataFrameType, annotate
+from merlin.dtypes.shape import DefaultShapes
 from merlin.schema import Schema
 from nvtabular.ops.operator import ColumnSelector, Operator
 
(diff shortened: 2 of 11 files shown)


nvtabular/ops/groupby.py
```diff
@@ -18,6 +18,7 @@
 from dask.dataframe.utils import meta_nonempty
 
 from merlin.core.dispatch import DataFrameType, annotate
+from merlin.dtypes.shape import DefaultShapes
```
oliverholworthy, line 21: Is this going to be a new feature of dtypes in core?

nvtabular/ops/list_slice.py
```diff
@@ -129,7 +129,7 @@ def transform(self, col_selector: ColumnSelector, df: DataFrameType) -> DataFram
 
     def _compute_dtype(self, col_schema, input_schema):
```
oliverholworthy, line 130: Is this method  required to be overriden with this change? or could it be removed?

karlhigley: It could be removed, but that breaks some of the tests even though the functionality works without it. (There are actually two places like this.)

nvtabular/ops/groupby.py
karlhigley: Shapes are implemented as a subfield of dtypes in Core, but this isn't really intended to be a feature of dtypes in particular. We might want to hide that implementation detail a bit more thoroughly by adjusting the imports.

As far as the defaults go, we just thought it was easier to read and understand than needing to remember which shapes mean what.

oliverholworthy: The `DefaultShapes.LIST` object seems like it could be useful, but the tests are failing to find this in core `ImportError: cannot import name 'DefaultShapes' from 'merlin.dtypes.shape`

nvtabular/ops/groupby.py
```diff
@@ -199,18 +197,26 @@ def _compute_dtype(self, col_schema, input_schema):
...
+        agg = self._find_agg(col_schema, input_schema)
+        dtype = agg_dtypes.get(agg, col_schema.dtype)
+
+        return col_schema.with_dtype(dtype)
+
+    def _compute_shape(self, col_schema, input_schema):
+        agg_is_lists = {"list": True}
+
+        agg = self._find_agg(col_schema, input_schema)
+        is_list = agg_is_lists.get(agg, col_schema.is_list)
```
oliverholworthy, line 209: In the case where we have fallback to the second argument of the `.get` here and check the col_schema.is_list attribute. If we have a fixed list, would we like to preserve that info in the shape later on instead of turning in into a ragged list which is presumably default shape corresponding to `DefaultShapes.LIST`

nvtabular/ops/groupby.py
karlhigley: Expected since there is (or was) an outstanding Core PR that adds it

karlhigley: Looks like it's still open: https://github.com/NVIDIA-Merlin/core/pull/215

nvtabular/loader/backend.py
```diff
@@ -22,9 +23,10 @@ def _augment_schema(
     cats=None,
     conts=None,
     labels=None,
-    sparse_names=None,
-    sparse_max=None,
-    sparse_as_dense=False,
+    padded_cols=None,
```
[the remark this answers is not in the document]
karlhigley: We couldn't really figure out what this function was supposed to do without the renames, so we went for it. The function name starts with an `_` so we've appropriately signaled that external code shouldn't depend on its stability. The two places in the NVTabular that use it call it via argument order instead of specifying the names, so I understand the caution but I think we're okay.

nvtabular/ops/groupby.py
karlhigley: I'm not actually sure. Are there `GroupBy` aggregations that don't change the shape of list columns?

nvtabular/ops/groupby.py
karlhigley: Merged now

nvtabular/loader/backend.py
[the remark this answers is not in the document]
oliverholworthy: looks like it will be a safe change, and the names make things clearer :). We may consider removing this functon along with the other loader code to follow-up on [the promise here](https://github.com/NVIDIA-Merlin/NVTabular/blob/v1.8.1/nvtabular/loader/__init__.py#L23) in one of the upcoming releases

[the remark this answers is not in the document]
oliverholworthy: And by that point we'll hopefully have updated the dataloader API to a point where this augment_schema function is no longer required either here or the copies [in Transformers4Rec](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/v0.1.16/merlin_standard_lib/utils/misc_utils.py#L243) and [in Merlin Models](https://github.com/NVIDIA-Merlin/models/blob/v0.11.0/merlin/models/loader/backend.py#L19). The existence of this function suggests to me that there's something missing from the dataloader API as it currently exists.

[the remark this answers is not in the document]
karlhigley: I think the something that's missing is "transforms implemented as operators that provide schema tracking" 😺

nvtabular/ops/groupby.py
karlhigley: From looking at [the list of aggregations](https://github.com/NVIDIA-Merlin/NVTabular/blob/main/nvtabular/ops/groupby.py#L192-L202), I think everything changes the shape, either from list to scalar or from scalar to list.

karlhigley: I wonder if the default there should actually be `False` 🤔

oliverholworthy: In practice it seems that it may not matter whether it's `col_schema.is_list` or `False`. I tried running Groupby with a an agg that is not `"list"` on a dataframe with list feautres and we get errors in both cudf and pandas

```python
from merlin.io import Dataset
import nvtabular as nvt
import cudf

df = cudf.DataFrame({"a": [1, 1, 2], "b": [[10], [20], [20]]})
workflow = nvt.Workflow(["a", "b"] >> nvt.ops.Groupby(groupby_cols=["a"], aggs=["sum"]))
workflow.fit_transform(Dataset(df)).compute()
# => Raises DataError: All requested aggregations are unsupported.
``` 

Some of these aggs, like `sum` or `mean` could in theory work on list features if we wanted them to.

Pandas for example, handles sum across lists as concatenation.

```python
import pandas as pd

df = pd.DataFrame({"a": [1, 1, 2], "b": [[10], [20], [20]]})
df.groupby("a").sum()
# =>
          b
a          
1  [10, 20]
2      [20]

```

or if numpy arrays, then as an element-wise sum

```python
df = pd.DataFrame({"a": [1, 1, 2], "b": [np.array([10]), np.array([20]), np.array([20])]})
df.groupby("a").sum()
# =>
      b
a      
1  [30]
2  [20]
```

`var/std/mean/median` also work and in this example return scalars. If the element type contained an array of more then 1 dimension, then `mean` could start returning a list type too

since `cudf` doesn't appear to support aggregating across list columns this is probably something we don't need to be concerned about for now.

```python
import cudf
df = cudf.DataFrame({"a": [1, 1, 2], "b": [[10], [20], [20]]})
df.groupby("a").sum()
# => Raises DataError: All requested aggregations are unsupported.

df["b"].sum()
# => Raises TypeError: cannot perform sum with type list
```

oliverholworthy: So I think `False` could be fine as the default here, and as far as this PR goes, it's no less clear than it was before.

If we need groupby aggregations for list columns as a future feature of NVTabular this will need to be revisited. I suppose even if cudf and pandas don't natively support this we could implement this ourselves by extracting the cupy/numpy arrays from the series in our own agg function to handle list column aggregations.

nvtabular/loader/backend.py
```diff
@@ -34,21 +36,20 @@ def _augment_schema(
...
-    # Set the appropriate properties for the sparse_names/sparse_max/sparse_as_dense
-    for col in sparse_names or []:
+    for col in padded_cols or []:
         cs = schema[col]
-        properties = cs.properties
-        if sparse_max and col in sparse_max:
-            properties["value_count"] = {"max": sparse_max[col]}
-        if sparse_as_dense:
-            properties["value_count"]["min"] = properties["value_count"]["max"]
+        dims = Shape(((1, batch_size), None))
```
oliverholworthy, line 41: What is the `batch_size` required for?

oliverholworthy (approved):

nvtabular/loader/backend.py
karlhigley: Not required for anything specific, just following the principle that the schema should always accurately reflect the data to the greatest extent possible. Here we have shape information since we know the batch size, so we fill it in in case that helps something downstream. I don't know if it actually will, but it seemed like the right thing to do.

nvtabular/ops/groupby.py
karlhigley: That's great to know; I really appreciate your thoroughness in testing this out. This probably warrants a further update to the shapes here, I'll open a separate issue.

karlhigley: Tracked here: https://github.com/NVIDIA-Merlin/NVTabular/issues/1763

PR #1774, opened 2023-03-08
Workflow syntax error

mikemckiernan:
This PR updates a boo-boo, but does not solve the already known problem of the API docs not rendering.


karlhigley (approved):

karlhigley:
@mikemckiernan Pretty sure that was me, so...oops! Sorry about that, and thanks for the fix.
 
@jperez999 The test failure here is unrelated; could you click the big admin merge override button to force it through and cover up my error here? 😅

PR #1775, opened 2023-03-09
Fix docs tox environment

alexanderronquillo:
The [API documentation](https://nvidia-merlin.github.io/NVTabular/main/api.html) was empty because `autosummary` couldn't import the source modules. I fixed this by installing `merlin-core` in the docs tox environment.

> **Note:** When I generate the docs now, I see some import warnings for other packages. However, I opted to ignore those for now because I was having trouble installing one of them (cudf) into the environment, and it didn't prevent the docs from generating.

I've also included a couple of small README changes:

- Fixed a typo
- Added a note about using Python 3.9 to generate documentation

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,7 +20,7 @@ Follow the instructions below to build the docs.
 
    This runs Sphinx in your shell and outputs to `docs/build/html/`.
 
-   > **Note:** Your virtual environment should use Python 3.9, not 3.10+:
+   > **Note:** Currently, we lock our Sphinx version to 3.6, which doesn't support Python 3.10. If you're using Python 3.10, install an earlier version of Python to build the docs:
    >
    > ```shell
    > sudo apt update
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,13 +12,22 @@ Follow the instructions below to build the docs.
 1. To build the docs, you need to install a developer environment and run `tox`:
 
    ```shell
-   python3 -m vevn .venv
+   python3 -m venv .venv
    source .venv/bin/activate
    python -m pip install -U pip tox
    tox -e docs
    ```
 
    This runs Sphinx in your shell and outputs to `docs/build/html/`.
+
+    > **Note:** Your virtual environment should use Python 3.9, not 3.10+:
+    >
+    > ```shell
+    > sudo apt update
+    > sudo apt install python3.9-venv
+    > sudo apt install python3.9-dev
+    > python3.9 -m venv .venv
+    > ```
 
 1. Start an HTTP server and review your updates:
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,14 +20,14 @@ Follow the instructions below to build the docs.
 
    This runs Sphinx in your shell and outputs to `docs/build/html/`.
 
-    > **Note:** Your virtual environment should use Python 3.9, not 3.10+:
-    >
-    > ```shell
-    > sudo apt update
-    > sudo apt install python3.9-venv
-    > sudo apt install python3.9-dev
-    > python3.9 -m venv .venv
-    > ```
+   > **Note:** Your virtual environment should use Python 3.9, not 3.10+:
+   >
+   > ```shell
+   > sudo apt update
+   > sudo apt install python3.9-venv
+   > sudo apt install python3.9-dev
+   > python3.9 -m venv .venv
+   > ```
 
 1. Start an HTTP server and review your updates:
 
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -79,6 +79,7 @@ commands =
 changedir = {toxinidir}
 deps = -rrequirements/docs.txt
 commands =
+    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/core.git@{posargs:main}
```
mikemckiernan, line 82: This is an admittedly inconsequential comment, but I'd just remove the `{posargs:main}` from the line:
* If you were to exercise this option by specifying a branch for the core repo, then the next line would croak because the git ref isn't a valid build type for Sphinx.
* In the next line, `posargs` is available so that `tox -e docs linkcheck` is possible.  Conversely, there's no `linkcheck` git ref for the core repo, so line 82 would croak.

However, no one might ever run linkcheck (that'd be sad) and no on might ever try to build the docs against a specific ref of the core repo.  So, the end game is that it's up to you whether you want to revise or not.

mikemckiernan (approved):
Huge thanks for doing the sleuthing.  Up to you whether you want to change how the core library is installed or not.

tox.ini
alexanderronquillo: Ah, good catch. Copy pasta.

docs/README.md
```diff
@@ -12,14 +12,23 @@ Follow the instructions below to build the docs.
...
-   python3 -m vevn .venv
+   python3 -m venv .venv
    source .venv/bin/activate
    python -m pip install -U pip tox
    tox -e docs
    ```
 
    This runs Sphinx in your shell and outputs to `docs/build/html/`.
 
+   > **Note:** Your virtual environment should use Python 3.9, not 3.10+:
```
oliverholworthy, line 23: what is it about Python 3.10+ that is unsupported in the docs build?

tox.ini
```diff
@@ -79,6 +79,7 @@ commands =
 changedir = {toxinidir}
 deps = -rrequirements/docs.txt
 commands =
+    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/core.git
     python -m sphinx.cmd.build -P -b {posargs:html} docs/source docs/build/{posargs:html}
```
oliverholworthy, line 83: An alternative could be to put this in `requirements/docs?` What part of the docs requires a feature from the core package? And does it matter what version is installed?

docs/README.md
alexanderronquillo: When I build with Python 3.10, I get this error:

```
ImportError: cannot import name 'Union' from 'types' (/usr/lib/python3.10/types.py)
```

According to [this Issue](https://github.com/sphinx-doc/sphinx/issues/9562) and [this PR](https://github.com/zopefoundation/zope.schema/pull/109), Python 3.10 isn't supported until Sphinx version 4.2.0.

We're limiting our Sphinx version to 3.6:

```
Sphinx<3.6
```

tox.ini
alexanderronquillo: `autosummary` has to import the python modules so it can read the docstrings. The `nvtabular` modules import modules from `merlin-core`, which wasn't included in the tox environment, so it raised a "failed import" error.

docs/README.md
oliverholworthy: that makes sense. And is there a reason we need Python 3.9 over 3.8?

alexanderronquillo: That's a good point. Can you review the new copy?

oliverholworthy: comment looks clearer now, thanks for updating

tox.ini
oliverholworthy: If in general the docs environment need to have all modules imported installed, then would it make sense to do a `pip install .` before the build so that anything specified in the package `install_requires` get's installed first?

alexanderronquillo: I think this makes sense. I added it in a new commit.

tox.ini
```diff
@@ -79,6 +79,7 @@ commands =
 changedir = {toxinidir}
 deps = -rrequirements/docs.txt
 commands =
+    python -m pip install .
```
oliverholworthy, line 82: I suggested this pip install in another comment, however I think it may not be strictly necessary, since tox installs the package dependencies as part of it's environment setup. And merlin-core is specified in the package install requires. 

I tried running this `docs` and `docs-multi` locally without this and both seem to produce the API docs with content, so might be some other reason they didn't populate

oliverholworthy (approved):

tox.ini
alexanderronquillo: hmm. ok, yeah then I'm stumped on what the problem would have been.

PR #1776, opened 2023-03-10
Enable `DatasetGen` usage in CPU environment

oliverholworthy:
Enable `DatasetGen` usage in CPU environment.

I ran into an error from this function when running in a CPU environment where the offsets were trying to call cupy.cumsum on numpy data.


karlhigley (approved):
This seems like a good fix for now 👍🏻 

In the longer run, I think we'd like to consolidate the dataset generation code and probably favor the code in Merlin Models over this version. Not sure how much is here that MM doesn't have though, so possible we can just get rid of this version.

PR #1777, opened 2023-03-10
Enable `Workflow.transform` to be run with a DataFrame type

oliverholworthy:
Enable `Workflow.transform` to be run with a DataFrame type.

### DIFF
--- a/nvtabular/workflow/workflow.py
+++ b/nvtabular/workflow/workflow.py
@@ -21,7 +21,8 @@ import sys
 import time
 import types
 import warnings
-from typing import TYPE_CHECKING, Optional, Union
+from functools import singledispatchmethod
+from typing import TYPE_CHECKING, Optional
 
 import cloudpickle
 import fsspec
@@ -32,7 +33,6 @@ except ImportError:
     cudf = None
 import pandas as pd
 
-from merlin.core.dispatch import DataFrameType
 from merlin.dag import Graph
 from merlin.dag.executors import DaskExecutor, LocalExecutor
 from merlin.dag.node import iter_nodes
@@ -79,7 +79,8 @@ class Workflow:
         self.graph = Graph(output_node)
         self.executor = DaskExecutor(client)
 
-    def transform(self, dataset: Union[Dataset, DataFrameType]) -> Union[Dataset, DataFrameType]:
+    @singledispatchmethod
+    def transform(self, data):
         """Transforms the dataset by applying the graph of operators to it. Requires the ``fit``
         method to have already been called, or calculated statistics to be loaded from disk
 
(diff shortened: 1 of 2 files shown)


nvtabular/workflow/workflow.py
```diff
@@ -78,7 +78,9 @@ def __init__(self, output_node: WorkflowNode, client: Optional["distributed.Clie
         self.graph = Graph(output_node)
         self.executor = DaskExecutor(client)
 
-    def transform(self, dataset: Dataset) -> Dataset:
+    def transform(
+        self, dataset: Union[Dataset, "cudf.DataFrame", pd.DataFrame]
```
karlhigley, line 82: The dataframe types are usually represented with `DataFrameType` from Core in the NVT code base. I'm not sure what we want to do about that yet—it might be something we can clean up either wit Protocols or using our new dispatch tools—but for now I'd follow the style of the existing code, just to make it easier to find all the places that need to change later.

nvtabular/workflow/workflow.py
```diff
@@ -96,7 +98,15 @@ def transform(self, dataset: Dataset) -> Dataset:
...
-        return self._transform_impl(dataset)
+        if isinstance(dataset, Dataset):
+            return self._transform_impl(dataset)
+        elif isinstance(dataset, pd.DataFrame) or (cudf and isinstance(dataset, cudf.DataFrame)):
+            return self._transform_df(dataset)
+        else:
+            raise ValueError(
+                "Workflow.transform received an unsupported type: {type(dataset)} "
+                "Supported types are a `merlin.io.Dataset` or DataFrame (pandas or cudf)"
+            )
```
karlhigley, line 109: I also wonder whether this might be a use case for either `singledispatch` or `lazy_singledispatch` since this code is essentially doing type-based dispatch anyway.

oliverholworthy: `singledispatchmethod` seems to  work relatively well here

nvtabular/workflow/workflow.py
```diff
@@ -89,23 +90,32 @@ def transform(self, dataset: Union[Dataset, DataFrameType]) -> Union[Dataset, Da
...
+
+    @transform.register
+    def _(self, dataset: Dataset) -> Dataset:
+        return self._transform_impl(dataset)
+
+    @transform.register
+    def _(self, dataframe: pd.DataFrame) -> pd.DataFrame:
+        return self._transform_df(dataframe)
+
+    if cudf:
```
karlhigley, line 114: I'm guessing this is because `DataFrameType` doesn't work for dispatch? 👍🏻

oliverholworthy: I didn't actually try it before making this commit I have now to check this and get the following error

```
nvtabular/workflow/workflow.py:112: in Workflow
    def _(self, dataframe: DataFrameType) -> DataFrameType:
/usr/lib/python3.8/functools.py:907: in register
    return self.dispatcher.register(cls, func=method)
/usr/lib/python3.8/functools.py:860: in register
    raise TypeError(
E   TypeError: Invalid annotation for 'dataframe'. typing.Union[pandas.core.frame.DataFrame, cudf.core.dataframe.DataFrame] is not a class.
```

karlhigley: Yeah, that makes sense. I’m good with it as is, was just curious.

karlhigley:
It looks like the failing test is an unrelated Horovod issue—seems like we’ve been seeing a lot of failures from those tests lately, not sure why.

Beyond the failing test, what else would prevent us from merging this as is?

oliverholworthy:
I haven't looked into the details of why the horovod tests are failing regularly either. Seems like we have an unstable test there.

From a functionality perspective, I think the PR achieves the goal of being able to run the transform on a dataframe. 

I'm on the fence about whether the use of singledispatch here is clearer to read than the if statements I added initially to the body of the transform method, I suppose if makes it easier to add more types in future if we anticipate more we'd like to add.

karlhigley (approved):

PR #1782, opened 2023-03-15
Fix Dataloader Unittest - which broke by new DL structure

bschifferer:


karlhigley (approved):

bschifferer:
rerun tests

bschifferer:
rerun tests

bschifferer:
rerun tests

tests/unit/loader/test_tf_dataloader.py
```diff
@@ -369,37 +368,36 @@ def test_mh_support(tmpdir, batch_size):
...
-            array = array.numpy()[:, 0]
-
-            if mh_name == "Embedding":
-                assert (nnzs == 3).all()
-            else:
-                lens = [
-                    len(x) for x in data[mh_name][idx * batch_size : idx * batch_size + n_samples]
-                ]
-                assert (nnzs == np.array(lens)).all()
+            array, offsets = X[f"{mh_name}__values"], X[f"{mh_name}__offsets"]
```
oliverholworthy, line 383: Using `merlin.table.TensorTable` here would simplify this and also make this code work with the previous version of the dataloader too.

```python
mh_col = TensorTable(X)[f"{mh_name}"]
values, offsets = mh_col.values.numpy(), mh_col.offsets.numpy()

File: tests/unit/loader/test_tf_dataloader.py [hunk 2]
```diff
@@ -64,15 +62,16 @@ def test_nested_list():
 
     batch = next(iter(train_dataset))
     # [[1,2,3],[3,1],[...],[]]
-    nested_data_col = tf.RaggedTensor.from_row_lengths(
-        batch[0]["data"][0][:, 0], tf.cast(batch[0]["data"][1][:, 0], tf.int32)
+    nested_data_col = tf.RaggedTensor.from_row_splits(
+        batch[0]["data__values"], tf.cast(batch[0]["data__offsets"], tf.int32)
```
oliverholworthy, line 66: Using TensorTable here would remove the need to specify the particular naming convention we have adopted for the dictionary keys containing values/offsets. 

```python
X = TensorTable(batch[0])

tf.RaggedTensor.from_row_splits((X["data"].values, X["data"].offsets, tf.int32)
```

tests/unit/examples/test_03-Running-on-multiple-GPUs-or-on-CPU.py
```diff
@@ -56,4 +56,9 @@ def test_example_03():
 
             """
         )
+        for cell in tb.cells:
+            cell.source.replace(
+                "device_memory_limit=device_limit", "# device_memory_limit=device_limit"
```
oliverholworthy, line 61: why do we need to disable these parameters? and is this related to the change to the dataloader?

bschifferer: No this is not related to the dataloader changes.

I think we changed the CI tests. We define a CUDA Cluster with multiple GPUs. We use RMM Pool to reserve GPU memory for this CUDA Cluster. It is more efficient (faster), if we reserve a fixed size of GPU memory. The code tries to reserve X% of total GPU memory (not free). If we run multiple tests in parallel, other tests block GPU memory and we cannot reserve it.

I deactive the parameters that we use the default behavior (not using RMM)

tests/unit/loader/test_tf_dataloader.py
bschifferer: Is it fine to leave the test as is ? I was not familiar with TensorTable

tests/unit/examples/test_03-Running-on-multiple-GPUs-or-on-CPU.py
oliverholworthy: maybe worth adding something along these lines as a comment above these lines that may help our future selves recovering context of why these parameters need to be changed

oliverholworthy (approved):

PR #1784, opened 2023-03-17
update conftest for backwards compat and new api for to parquet call

jperez999:
For cudf version 22.10.00+ the to_parquet api has changed. It no longer support the chunk_size call. It was replaced with row_group_size_rows, which can take minimum of 5000 rows. We have updated conftest call to reflect this change while maintaining backwards compatibility. We need this for all future versions.


rjzamora:
>For cudf version 22.10.00+ the to_parquet api has changed. It no longer support the chunk_size call. It was replaced with row_group_size_rows

Hmm - I may be mistaken, but I don't think `chunksize` was **ever** a supported argument to `cudf.to_parquet`. I suspect that older `cudf` versions were  just silently ignoring `chunk_size=1000`.

PR #1791, opened 2023-03-29
use merlin compat for imports of gpu specific packages

jperez999:
This PR enables a user to run our GPU docker container without GPUs and successfully be able to run all tests. Previously we had try catches in a few places in the code for importing GPU specific packages. This PR remediates those try catches because in an environment where you have the package but no GPU you end up getting ccuda.pyx init failure. This is because, the package (i.e. cudf, dask_cudf, rmm) do exist but when they try to access information about GPUs it fails and throws an error something like:
```
collecting ... Traceback (most recent call last):
  File "cuda/_cuda/ccuda.pyx", line 3671, in cuda._cuda.ccuda._cuInit
  File "cuda/_cuda/ccuda.pyx", line 435, in cuda._cuda.ccuda.cuPythonInit
RuntimeError: Failed to dlopen libcuda.so
Exception ignored in: 'cuda._lib.ccudart.utils.cudaPythonGlobal.lazyInitGlobal'
  File "cuda/_cuda/ccuda.pyx", line 3671, in cuda._cuda.ccuda._cuInit
  File "cuda/_cuda/ccuda.pyx", line 435, in cuda._cuda.ccuda.cuPythonInit
RuntimeError: Failed to dlopen libcuda.so
Fatal Python error: Segmentation fault
```
This PR leverages the compat file, and makes it the single point of import for the main packages (cudf, cupy) and it adds a security around it that ensures you can only import those packages if GPUs are detected. So if you find yourself in a scenario where the package is installed but no GPUs are detected you can, now, still safely use the core package. Therefore, we can now run our containers with and without the --gpus=all flag in docker. This was a customer ask and it helps developers when trying to test cpu only environment on a resource that has GPUs.

This PR works in conjunction with https://github.com/NVIDIA-Merlin/core/pull/261

### DIFF
--- a/.github/workflows/cpu-packages.yml
+++ b/.github/workflows/cpu-packages.yml
@@ -6,11 +6,10 @@ on:
     branches: [main]
     tags:
       - v*
-  pull_request:
-    branches: [main]
 
 jobs:
   build-conda:
+    name: Build Conda
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -50,6 +49,7 @@ jobs:
           path: ${{ steps.conda_build.outputs.conda_package }}
 
   build-pypi:
+    name: Build PyPI
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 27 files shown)


karlhigley (approved):

karlhigley:
The remaining test failure here is addressed by NVIDIA-Merlin/core#276, so waiting on that PR to go through in order to merge this one

PR #1794, opened 2023-03-31
Bump `merlin-core` and `dataloader` deps to `23.02.01`

karlhigley:
We keep getting questions from people who are using newer versions of NVT with older versions of Core. We'll need to bump this again when we cut the next release, but at least this should help when people clone from `main` in the interim

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,3 +1,3 @@
-merlin-core>=0.2.0
-merlin-dataloader>=0.0.2
+merlin-core>=23.02.01
+merlin-dataloader>=23.02.01
 scipy


requirements/base.txt
```diff
@@ -1,3 +1,3 @@
-merlin-core>=0.2.0
-merlin-dataloader>=0.0.2
+merlin-core>=23.02.01
```
oliverholworthy, line 1: The version numbers in PyPI don't have leading zeros. (E.g. 23.2.1) will this work like this?

karlhigley: Yeah, seems like the install still works okay

jperez999 (approved):

PR #1795, opened 2023-04-03
Replace nvtabular.utils with merlin.core.compat

edknv:
Same change as https://github.com/NVIDIA-Merlin/Merlin/pull/892.

There was a change in Core that [moved pynvml_mem_size from merlin.core.utils to merlin.core.compat](https://github.com/NVIDIA-Merlin/core/commit/568208cf8598ed6ac982321689e4f67907c96483#diff-71c42f09a533dd2861bcb1c78b2b350190518b1154eb735c50519a090dd8061cR36). This PR replaces `nvtabular.utils` with `merlin.core.compat` in [examples/03-Running-on-multiple-GPUs-or-on-CPU.ipynb](https://github.com/NVIDIA-Merlin/NVTabular/blob/main/examples/03-Running-on-multiple-GPUs-or-on-CPU.ipynb).

### DIFF
--- a/examples/03-Running-on-multiple-GPUs-or-on-CPU.ipynb
+++ b/examples/03-Running-on-multiple-GPUs-or-on-CPU.ipynb
@@ -142,7 +142,7 @@
     "from dask_cuda import LocalCUDACluster\n",
     "from dask.distributed import Client\n",
     "import nvtabular as nvt\n",
-    "from nvtabular.utils import pynvml_mem_size, device_mem_size\n",
+    "from merlin.core.compat import pynvml_mem_size, device_mem_size\n",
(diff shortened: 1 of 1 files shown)


karlhigley (approved):
Wonder if `nvtabular.utils` should import `merlin.core.compat.pynvml_mem_size` in order to work around the breaking change in Core

edknv:
It looks like we've meant to deprecate `nvtabular.utils` and move it to Core for a while: https://github.com/NVIDIA-Merlin/NVTabular/blob/cdc5ff12de67e56101e9e495c1f3ab7e39b4e700/nvtabular/utils.py#L24-L27

PR #1804, opened 2023-04-25
Update requirements for Merlin packages to minimum version of 23.04

karlhigley:
Update requirements for Merlin Core package to minimum version of 23.04. This contains features required for the upcoming release.

Depends on merlin-core 23.4.0 package publish

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,3 +1,3 @@
-merlin-core>=0.2.0
-merlin-dataloader>=0.0.2
+merlin-core>=23.4.0
+merlin-dataloader>=23.4.0
 scipy


requirements/base.txt
```diff
@@ -1,3 +1,3 @@
-merlin-core>=0.2.0
-merlin-dataloader>=0.0.2
+merlin-core>=23.4.0
+merlin-core>=23.4.0
```
oliverholworthy, line 2: are we removing merlin-dataloader here or is that a typo?

oliverholworthy: updated as I submitted the comment

oliverholworthy:
@karlhigley This change was applied to the release branch, however, I think we still need to merge this into the main branch so that it's in the next release too

PR #1805, opened 2023-04-26
Update conda package publish for muliple python versions

oliverholworthy:
Updates conda package build so that we build and publish the NVTabular conda package for multiple python versions.

Currently we only publish a package for Python 3.8. This PR adds a package for 3.9, 3.10, in addition to 3.8.

And removes the upper bound on the cupy version in the dependencies

### DIFF
--- a/.github/workflows/cpu-packages.yml
+++ b/.github/workflows/cpu-packages.yml
@@ -122,37 +122,21 @@ jobs:
           name: pr
           path: pr/
 
-  check-conda:
-    name: Check Conda Artifacts
+  release-conda:
+    name: Release Conda
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: ["3.8", "3.9", "3.10"]
-    needs: [build-conda]
-    steps:
-      - uses: actions/setup-python@v4
-        with:
-          python-version: 3.9
-      - uses: actions/download-artifact@v2
-        with:
-          name: conda-${{ matrix.python-version }}
-          path: conda
-      - name: check artifacts
-        run: |
-          find conda -name "*.tar.gz"
-
-  release-conda:
-    name: Release Conda
-    runs-on: ubuntu-latest
     if: "startsWith(github.ref, 'refs/tags/')"
     needs: [build-conda]
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/cpu-packages.yml
```diff
@@ -10,20 +10,15 @@ on:
 jobs:
   build-conda:
     name: Build Conda
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        python-version: [3.8]
```
karlhigley, line 16: Wondering if we could add 3.9 and 3.10 here so the builds run in separate actions?

.github/workflows/cpu-packages.yml
```diff
@@ -30,22 +34,28 @@ jobs:
...
+          echo "conda pkgs dir $CONDA_PKGS_DIRS"
           conda update conda
-          conda install conda-build pybind11
-          conda build --python ${{ matrix.python-version }} . -c defaults -c conda-forge -c numba -c rapidsai -c nvidia --output-folder ./conda_packages
-          export CONDA_PACKAGE=$(conda build --python ${{ matrix.python-version }} . -c defaults -c conda-forge -c numba -c rapidsai -c nvidia --output-folder ./conda_packages --output)
-          echo "conda_package : $CONDA_PACKAGE"
-          echo "conda_package=$CONDA_PACKAGE" >> $GITHUB_OUTPUT
+          conda install -c conda-forge mamba
+          mamba install -c conda-forge conda-build boa conda-verify pybind11
+          conda mambabuild --python ${{ matrix.python-version }} . -c defaults -c conda-forge -c numba -c rapidsai -c nvidia --output-folder ./conda_packages
```
oliverholworthy, line 50: Using mambabuild instead of regular conda build cut the build time down from 3-5 hours to about 10 minutes.

oliverholworthy:
This PR has already been applied to the release branch, and sucessfully published versions of the conda package for Python 3.9 and 3.10 (in addition to 3.8). Still needs to be merged into main so that this will also apply to the next release

karlhigley (approved):

PR #1807, opened 2023-04-28
Convert `Workflow.fit()` to accept a `Dataset` instead of a ddf

karlhigley:
This enables operator `fit()` methods to pass a `Dataset` object through to the `fit()` methods of underlying subgraphs.

Depends on https://github.com/NVIDIA-Merlin/core/pull/299


nvtabular/ops/categorify.py
```diff
@@ -325,6 +325,8 @@ def fit(self, col_selector: ColumnSelector, ddf: dd.DataFrame):
                 for col in group.names:
                     self.storage_name[col] = name
 
+        ddf = dataset.to_ddf()
```
karlhigley, line 328: This is the part I'm a little bit concerned about, because I'm worried that it will read more columns than are needed for the fitting. In order to make `fit` accept a `Dataset` I had to move the `to_ddf()` call into the operators, but we don't have all the information we'd need in order to determine which columns are used and which aren't.

The `ColumnSelector` is a start, but doesn't include additional dependency columns that are used but not transformed by the operator. That part lives in the `Node` input schema, which was accessible from the Workflow fitting code, but isn't from inside the operator.

So, this works, but there may be a performance regression here.

nvtabular/ops/categorify.py
```diff
@@ -38,10 +37,11 @@
 from merlin.core import dispatch
 from merlin.core.dispatch import DataFrameType, annotate, is_cpu_object, nullable_series
 from merlin.core.utils import device_mem_size, run_on_worker
+from merlin.dag.stat_operator import StatOperator
```
karlhigley, line 40: `StatOperator` moved to Merlin Core as part of the base ops that are available to all Merlin DAGs (including NVT, Systems, etc)

nvtabular/workflow/workflow.py
```diff
@@ -478,13 +445,3 @@ def check_version(stored, current, name):
             stat.op.set_storage_path(path, copy=False)
 
         return workflow
-
-    def clear_stats(self):
```
karlhigley, line 482: Moved to Core

nvtabular/workflow/workflow.py
```diff
@@ -200,44 +201,7 @@ def fit(self, dataset: Dataset) -> "Workflow":
         Workflow
             This Workflow with statistics calculated on it
         """
-        self.clear_stats()
```
karlhigley, line 203: This chunk of code and the `clear_stats()` method moved to Core

tests/unit/workflow/test_workflow_subgraphs.py
```diff
@@ -0,0 +1,57 @@
...
+#
+import pytest
+
+import nvtabular.ops as ops
+from merlin.dag import ColumnSelector
+from merlin.dag.ops.subgraph import Subgraph
+from nvtabular import Workflow
+
+
+@pytest.mark.parametrize("engine", ["parquet"])
```
karlhigley, line 24: These tests check that the normal graph API that works on other operators also works as expected with the `Subgraph` operator. They're easier to test here than in Core because we have operators that effect actual transforms here, but I'm starting to think that the Rename and Lambda operators might be appropriate to move to Core as part of the generic DAG operators that can be used in all Merlin DAGs.

nvtabular/ops/target_encoding.py
```diff
@@ -25,11 +24,12 @@
     read_parquet_dispatch,
 )
 from merlin.dag import Node
+from merlin.dag.stat_operator import StatOperator
```
oliverholworthy, line 27: After this change,  it looks like some ops won't be subclasses of the `nvtabular.ops.Operator` any more. Is the additional `create_node` method and `infererence_initialize` on that base class no longer relevant to the StatOperator? and subsequent operators here? If we'd prefer all operators in NVTabular to inherit from `nvtabular.ops.Operator`, one option might be to change `StatOperator` to a mixin style class that can be used alongside an `Operator` to make it Stat aware (support `fit`). 

```python
class MyOperator(Operator, StatMixin):
    def transform(...):
        ...
    def fit(...):
        ...
```

oliverholworthy:
Looks like this PR also depends on https://github.com/NVIDIA-Merlin/core/pull/301

nvtabular/ops/target_encoding.py
karlhigley: Good catch!

`inference_initialize` is for the custom C++ implementations of the operators for serving; `Categorify` is both a `StatOperator` and has a custom implementation, so that probably does matter.

The `create_node` method specifies a type for Workflows that shims in additional functionality to handle implicit conversion to `LambdaOp`. I'm thinking we should probably move that into Core anyway, since there's nothing particularly NVT-specific about using lambdas to do arbitrary transforms.

PR #1818, opened 2023-05-10
Add support for int8 values with Categorify inference

oliverholworthy:
Enables CategorifyTransform operator (Categorify in inference mode) used by Merlin Systems to transform inputs of type int8.


rnyak (approved):

rnyak:
rerun tests

PR #1827, opened 2023-05-25
How to for embedding op with categorify from scratch

jperez999:
This PR adds a test that illustrates how to use pretrained embeddings and how to reuse a categorify operator to either categorify data  to what an item map, or categorify an item map to data that was already fitted in a workflow. It shows how to take that all the way to the dataloader where we can use it to load in embeddings for new data presented.

Requires the following core pr https://github.com/NVIDIA-Merlin/core/pull/337


karlhigley (commented):
This looks good, but the PR could use a title that's legible to people outside the team (since the PR titles are now part of our auto-generated release notes)

karlhigley (approved):

PR #1832, opened 2023-06-05
Use tmpdir for Categorify out_path in test_tf4rec

oliverholworthy:
Use tmpdir for Categorify out_path in test_tf4rec. This avoids conficts when re-running the test in the same environment.


nv-alaiacano (approved):

nv-alaiacano:
The `test_tf4rec.py` test is passing, so I'm going to force this PR through.

PR #1838, opened 2023-06-07
handle copying of partitioned stat files when saving workflow

nv-alaiacano:
Addresses https://github.com/NVIDIA-Merlin/NVTabular/issues/1837

For operators that produce a directory of partitioned files rather than a single parquet file, we will copy the partitions to the expected target path when saving the Workflow.

There's also a small tweak to the save/load paths to make it easier to use other path-like types (eg tmpdir)

### DIFF
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -1693,7 +1693,7 @@ def _encode(
     expr = df[selection_l.names[0]].isna()
     for _name in selection_l.names[1:]:
         expr = expr & df[_name].isna()
-    nulls = df[expr].index
+    nulls = df[expr].index.values
 
     if use_collection or not search_sorted:
         if list_col:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -97,7 +97,7 @@ _CUDA_CLUSTER = None
 
 @pytest.fixture(scope="module")
 def client():
-    distributed = Distributed(n_workers=2)
+    distributed = Distributed()
     cluster = distributed.cluster
     client = distributed.client
     yield client


tox.ini
```diff
@@ -42,10 +42,9 @@ deps =
     pytest
     pytest-cov
 commands =
-    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/core.git
-    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/dataloader.git
-    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/models.git
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/core.git@{posargs:main}
+    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/dataloader.git@{posargs:main}
```
jperez999, line 46: Why did you make this move here? we always end with core because the other packages will force a pull of core from the available packages on pypi and wont use the main version of the code. Is this not the behavior your seeing?

jperez999 (changes requested):

jperez999 (approved):

tox.ini
oliverholworthy: I think that used to be the case because of the way versioneer was working and the fact that the release tags were often not on the main branch. However, with the new dev tags on the main branch this may not be an issue anymore

oliverholworthy (approved):

tox.ini
nv-alaiacano: I moved it out of this PR for now - we can address it elsewhere

PR #1842, opened 2023-06-09
working subgraphs in workflow

jperez999:
This PR adds support for subgraphs in nvt workflows. It requires a core PR to be merged beforehand https://github.com/NVIDIA-Merlin/core/pull/343. 

Also this PR adds a test case to ensure that fit and transform work as expected and verifies that the resulting dataset has valid information.

### DIFF
--- a/nvtabular/workflow/workflow.py
+++ b/nvtabular/workflow/workflow.py
@@ -143,8 +143,8 @@ class Workflow:
         return self
 
     @property
-    def subgraphs(self):
-        return self.graph.subgraphs.keys()
+    def subworkflows(self):
+        return list(self.graph.subgraphs.keys())
 
     @property
     def input_dtypes(self):
@@ -169,7 +169,7 @@ class Workflow:
     def _input_columns(self):
         return self.graph._input_columns()
 
-    def get_subgraph(self, subgraph_name):
+    def get_subworkflow(self, subgraph_name):
         subgraph = self.graph.subgraph(subgraph_name)
         return Workflow(subgraph.output_node)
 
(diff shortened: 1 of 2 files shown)


nvtabular/workflow/workflow.py
```diff
@@ -165,6 +169,10 @@ def output_node(self):
     def _input_columns(self):
         return self.graph._input_columns()
 
+    def get_subgraph(self, subgraph_name):
+        subgraph = self.graph.subgraph(subgraph_name)
+        return Workflow(subgraph.output_node)
```
oliverholworthy, line 174: Since this method returns a `Workflow`, perhaps `get_subworkflow` would match this return type more closely?

oliverholworthy: If we did go with `get_subworkflow` or similar for this method, then the property above `subraphs` might also need to change to something like `subworkflows` or `subworkflow_names`?

tests/unit/workflow/test_workflow_subgraphs.py
```diff
@@ -0,0 +1,103 @@
...
+    if not dump:
+        assert subgraph_cat.output_node == cat_features
+        assert subgraph_cont.output_node == cont_features
+    # check failure path works as expected
+    with pytest.raises(ValueError) as exc:
+        workflow.get_subworkflow("not_exist")
+    assert "No subgraph named" in str(exc.value)
+
+    sub_cat_df = subgraph_cat.transform(dataset).to_ddf().compute()
+    assert assert_eq(sub_cat_df.reset_index(drop=True), df_pp[cat_names].reset_index(drop=True))
```
oliverholworthy, line 97: The assert is already part of `assert_eq` - this extra assert needs to be removed (the same on a few lines down)

oliverholworthy (approved):
Last thing I think is to add some docstrings for the new methods

PR #1846, opened 2023-06-21
rename temporary column to something less likely to be used

nv-alaiacano:
The internal/temporary column "labels" is a common thing for people to use. Columns with this name will not get categorified. I changed it to `__labels_tmp`.


karlhigley (approved):

rjzamora (dismissed):
Thanks @nv-alaiacano !

rjzamora:
Ah - Looks like we also make the `"labels"` assumption [in `core`](https://github.com/NVIDIA-Merlin/core/blob/af07206b3cfa5088f89a1ac2a0c97378679a614f/merlin/io/worker.py#L133) :/

PR #1869, opened 2023-10-18
fix: IndexError with columns full of NaNs 

### LINKED ISSUE #1865: Index error with Categorify on transform step for columns with 100% NaNs
I was running a `workflow.transform(sampled_dataset)` step on a sample of my inference dataset and received the following error 

```
  File "/databricks/python/lib/python3.8/site-packages/nvtabular/ops/categorify.py", line 510, in transform
    encoded = _encode(
  File "/databricks/python/lib/python3.8/site-packages/nvtabular/ops/categorify.py", line 1707, in _encode
    if isinstance(df[cl].dropna().iloc[0], (np.ndarray, list)):
  File "/databricks/python/lib/python3.8/site-packages/pandas/core/indexing.py", line 1073, in __getitem__
    return self._getitem_axis(maybe_callable, axis=axis)
  File "/databricks/python/lib/python3.8/site-packages/pandas/core/indexing.py", line 1625, in _getitem_axis
    self._validate_integer(key, axis)
  File "/databricks/python/lib/python3.8/site-packages/pandas/core/indexing.py", line 1557, in _validate_integer
    raise IndexError("single positional indexer is out-of-bounds")
IndexError: single positional indexer is out-of-bounds

The above exception was the direct cause of the following exception:

  File "/databricks/python/lib/python3.8/site-packages/merlin/dag/executors.py", line 237, in _run_node_transform
    transformed_data = node.op.transform(selection, input_data)
  File "/databricks/python/lib/python3.8/site-packages/merlin/core/dispatch.py", line 69, in inner2
    return func(*args, **kwargs)
  File "/databricks/python/lib/python3.8/site-packages/nvtabular/ops/categorify.py", line 534, in transform
    raise RuntimeError(f"Failed to categorical encode column {name}") from e
RuntimeError: Failed to categorical encode column my_categorical_column
```

I noticed this happens when the dataset to be transformed has a categorical column (`my_categorical_column`) with 100% NaNs. It looks like that happens when this line is reached 👇 where we do a `dropna()` followed by `iloc[0]`

https://github.com/NVIDIA-Merlin/NVTabular/blob/ee21af08dc7def662a661dbea35957af43e91a09/nvtabular/ops/categorify.py#L1707

It's not a huge blocker for me right now, as this mostly happens on dataset samples, but I'm wondering whether that behavior is expected. Any thoughts? 😃

lecardozo:
Closes #1865 

The error reported there was caused by an indexing on an empty series. To work around that, we are adding a pre-condition to check whether the series after `dropna()` is empty, so we don't try to check its first element.

### DIFF
--- a/nvtabular/ops/categorify.py
+++ b/nvtabular/ops/categorify.py
@@ -1704,7 +1704,10 @@ def _encode(
             codes = type(df)({"order": dispatch.arange(len(df), like_df=df)}, index=df.index)
 
         for cl, cr in zip(selection_l.names, selection_r.names):
-            if isinstance(df[cl].dropna().iloc[0], (np.ndarray, list)):
+            column_without_nans = df[cl].dropna()
+            if len(column_without_nans) and isinstance(
+                column_without_nans.iloc[0], (np.ndarray, list)
+            ):
                 ser = df[cl].copy()
                 codes[cl] = dispatch.flatten_list_column_values(ser).astype(value[cr].dtype)
             else:
--- a/tests/unit/ops/test_categorify.py
+++ b/tests/unit/ops/test_categorify.py
@@ -738,7 +738,7 @@ def test_categorify_inference():
 
 def test_categorify_transform_only_nans_column():
     train_df = make_df({"cat_column": ["a", "a", "b", "c", np.nan]})
-    cat_features = ["cat_column"] >> nvt.ops.Categorify(max_size=4)
+    cat_features = ["cat_column"] >> nvt.ops.Categorify()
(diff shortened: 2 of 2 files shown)


jperez999:
tests are running via an internal PR here https://github.com/NVIDIA-Merlin/NVTabular/pull/1870

jperez999 (approved):

lecardozo:
@jperez999 checked the GPU CI workflow and it seems like it was cancelled externally. Let me know if there's anything I should/need to do to solve that :)

PR #1874, opened 2024-04-10
Fix `Categorify` inference and testing

rjzamora:
Fixes `Categorify` inference behavior with https://github.com/NVIDIA-Merlin/systems/pull/389 in place. Also includes a minor improvement in test coverage (checks that the cpp code path is actually "correct").


nvtabular/ops/operator.py
```diff
@@ -13,6 +13,15 @@
...
+    DataFormats,
+)
 
-Operator = BaseOperator
+
+# Avoid TENSOR_TABLE by default (for now)
+class Operator(BaseOperator):
+    @property
+    def supported_formats(self):
+        return DataFormats.PANDAS_DATAFRAME | DataFormats.CUDF_DATAFRAME
```
rjzamora, line 27: With https://github.com/NVIDIA-Merlin/systems/pull/389 applied to `systems`, I was seeing inference errors in `Groupby`. This is because the `Groupby` operator calls `sort_values`, and sorting is not supported for merlins `TensorTable` abstraction.

Since I am not entirely sure which NVTabular operations are supported with `TensorTable`, I figured the safest fix was to assume **none** of the operations are supported for now. (Any thoughts on this @jperez999 ?)

In a follow-up to this PR, it probably makes sense to add `GeneralOperator = BaseOperator` above this. That way, any operator with known support for `TensorTable` could inherit from `GeneralOperator` instead.

tests/unit/ops/test_categorify.py
```diff
@@ -734,3 +734,8 @@ def test_categorify_inference():
     output_tensors = inference_op.transform(cats.input_columns, input_tensors)
     for key in input_tensors:
         assert output_tensors[key].dtype == np.dtype("int64")
+
+    # Check results are consistent with python code path
+    expect = workflow.transform(df)
+    got = pd.DataFrame(output_tensors)
+    assert_eq(expect, got)
```
rjzamora, line 741: This test was only.checking the data type of the result. Now it also checks if the result is correct.

oliverholworthy (approved):

nvtabular/ops/operator.py
rnyak: @oliverholworthy and @jperez999 do you have a chance to look at Rick's comments above? thanks.

jperez999: This should be alright. All of the operators in nvtabular were created with data frames in mind. If we ever decide to add in tensor table, we can make the change then. If this speeds up the runs I say we should do it and solves breaking issues we should execute.

